PR #102, opened 2019-01-24
Updated spark start command

klunacognite:
Updated spark start command based on Joels suggestions and changed mvn to sbt.
Additional jar required


hakontro:
Outdated.

PR #159, opened 2019-02-28
Allow base URL to be configured

hakontro:
Refactor code to let users specify other base URLs than
api.cognitedata.com, such as greenfield.cognitedata.com.

### DIFF
--- a/src/main/scala/com/cognite/spark/datasource/Constants.scala
+++ b/src/main/scala/com/cognite/spark/datasource/Constants.scala
@@ -9,4 +9,5 @@ object Constants {
   val DefaultDataPointsAggregationBatchSize = 10000
   val DefaultInitialRetryDelay = 150.millis
   val DefaultMaxBackoffDelay = 120.seconds
+  val DefaultBaseUrl = "https://api.cognitedata.com"
 }
--- a/src/main/scala/com/cognite/spark/datasource/RawTableRelation.scala
+++ b/src/main/scala/com/cognite/spark/datasource/RawTableRelation.scala
@@ -136,7 +136,7 @@ class RawTableRelation(
   }
 
   def baseRawTableURL(project: String, database: String, table: String): Uri =
-    uri"${baseUrl(project, "0.5")}/raw/$database/$table"
+    uri"${baseUrl(project, "0.5", config.baseUrl)}/raw/$database/$table"
 }
 
 object RawTableRelation {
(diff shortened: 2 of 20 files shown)


src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -97,5 +97,5 @@ class TimeSeriesRelation(config: RelationConfig)(val sqlContext: SQLContext)
   }
 
   def baseTimeSeriesUrl(project: String): Uri =
-    uri"${baseUrl(project)}/timeseries"
+    uri"${baseUrl(project, "0.5", Constants.DefaultUrl)}/timeseries"
```
wjoel: Should be `config.baseUrl` I think?

src/main/scala/com/cognite/spark/datasource/Constants.scala
```diff
@@ -9,4 +9,5 @@ object Constants {
   val DefaultDataPointsAggregationBatchSize = 10000
   val DefaultInitialRetryDelay = 150.millis
   val DefaultMaxBackoffDelay = 120.seconds
+  val DefaultUrl = "https://api.cognitedata.com"
```
wjoel: `DefaultBaseUrl` to match the others.

src/main/scala/com/cognite/spark/datasource/RawTableRelation.scala
```diff
@@ -136,7 +136,7 @@ class RawTableRelation(
   }
 
   def baseRawTableURL(project: String, database: String, table: String): Uri =
-    uri"${baseUrl(project, "0.5")}/raw/$database/$table"
+    uri"${baseUrl(project, "0.5", Constants.DefaultUrl)}/raw/$database/$table"
```
wjoel: Should be `config.baseUrl`

src/test/scala/com/cognite/spark/datasource/BaseUrlTest.scala
```diff
@@ -0,0 +1,18 @@
...
+class BaseUrlTest extends FlatSpec with SparkTest {
+  private val greenfieldApiKey = System.getenv("TEST_API_KEY_GREENFIELD")
+
+  it should "list the correct number of 3D models on greenfield.cognitedata.com" taggedAs GreenfieldTest in {
+    val df = spark.read
+      .format("com.cognite.spark.datasource")
+      .option("apiKey", greenfieldApiKey)
+      .option("type", "3dmodels")
+      .option("limit", 4)
+      .option("baseUrl", "https://greenfield.cognitedata.com")
```
wjoel: This test wouldn't catch if this parameter was ignored. Can we find something where we can load from each project and compare the results, making sure they're different? Maybe files metadata is a good option, just counting the number of files in each and ensuring they're not the same?

PR #168, opened 2019-03-06
Report datapoints.read metrics

serenityue:
Please ignore if this was left out intentionally.


wjoel:
If there is a nicer way we can find it later.

PR #170, opened 2019-03-07
Add developers metadata required by OSSRH Sonatype repository

wjoel:
Let me know if you'd like to be added to the list of developers, it's a required field so it needs to be set to something, but I don't want it to look like I'm trying to take all the credit. ;)


Tapped:
```    
Developer(
      id = "tapped",
      name = "Emil Sandstø",
      email = "emil.sandsto@cognite.com",
      url = url("https://github.com/tapped")
    )
```

PR #174, opened 2019-03-08
Enable upsert of time series

hakontro:
Tests are still not 100% stable wrt. eventual consistency.

### DIFF
--- a/src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
+++ b/src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
@@ -4,14 +4,25 @@ import cats.effect.{ContextShift, IO}
 import cats.implicits._
 import com.softwaremill.sttp._
 import io.circe.generic.auto._
+import io.circe.parser.decode
 import org.apache.spark.sql.sources._
-import org.apache.spark.sql.types._
+import org.apache.spark.sql.types.{DataTypes, StructType}
 import org.apache.spark.sql.{Row, SQLContext}
 import SparkSchemaHelper._
 
 import scala.concurrent.ExecutionContext
 
 case class PostTimeSeriesDataItems[A](items: Seq[A])
+case class TimeSeriesConflict(notFound: Seq[Long])
+case class TimeSeriesNotFound(notFound: Seq[String])
+case class Setter[A](set: A, setNull: Boolean)
+object Setter {
+  def apply[A](set: Option[A]): Option[Setter[A]] =
+    set match {
+      case None => None
+      case _ => Some(new Setter(set.get, false))
+    }
+}
 
 case class TimeSeriesItem(
     name: String,
@@ -36,7 +47,34 @@ case class PostTimeSeriesItem(
     assetId: Option[Long],
     isStep: Boolean,
     description: Option[String],
-    securityCategories: Option[Vector[Long]])
+    securityCategories: Option[Vector[Long]],
+    id: Long)
+
+case class UpdateTimeSeriesItem(
+    id: Long,
+    name: Option[Setter[String]],
+    metadata: Option[Setter[Map[String, String]]],
+    unit: Option[Setter[String]],
+    assetId: Option[Setter[Long]],
+    description: Option[Setter[String]],
+    securityCategories: Option[Map[String, Option[Vector[Long]]]],
+    isString: Option[Setter[Boolean]],
+    isStep: Option[Setter[Boolean]]
+)
+object UpdateTimeSeriesItem {
+  def apply(postTimeSeriesItem: PostTimeSeriesItem): UpdateTimeSeriesItem =
+    new UpdateTimeSeriesItem(
+      postTimeSeriesItem.id,
+      Setter[String](Some(postTimeSeriesItem.name)),
+      Setter[Map[String, String]](postTimeSeriesItem.metadata),
+      Setter[String](postTimeSeriesItem.unit),
+      Setter[Long](postTimeSeriesItem.assetId),
+      Setter[String](postTimeSeriesItem.description),
+      securityCategories = postTimeSeriesItem.securityCategories.map(a => Map("set" -> Option(a))),
+      Setter[Boolean](Some(postTimeSeriesItem.isString)),
+      Setter[Boolean](Some(postTimeSeriesItem.isStep))
+    )
+}
 
 class TimeSeriesRelation(config: RelationConfig)(val sqlContext: SQLContext)
     extends CdpRelation[TimeSeriesItem](config, "3dmodelrevisionnodes")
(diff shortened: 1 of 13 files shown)


src/test/scala/com/cognite/spark/datasource/SparkTest.scala
```diff
@@ -29,4 +32,17 @@ trait SparkTest extends CdpConnector {
 
     (modelId, revisionId)
   }
+
+  def retryUntil[A](action: => A, shouldRetry: A => Boolean): A =
```
wjoel: I think the name was my idea... but it should be `retryWhile` because that's what it actually does. ;)

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -87,13 +91,42 @@ class TimeSeriesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
         }
-        item
-      })
+      }
+    }
+  }
+
+  def resolveConflict(timeSeriesItems: Seq[PostTimeSeriesItem], timeSeriesConflict: TimeSeriesNamesNotFound): IO[Unit] = {
+    implicit val contextShift: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
+    val notExistingNames = timeSeriesConflict.notFound
```
wjoel: Call it `notFound` instead, it's names today but it will be "external id" in the future it seems.

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -87,13 +91,42 @@ class TimeSeriesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+
+    val alreadyExistingPostTimeSeriesItems = timeSeriesItems.filter(p => !notExistingNames.contains(p.name))
```
wjoel: And I guess we can call this `foundTimeSeriesItems` then.

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -87,13 +91,42 @@ class TimeSeriesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+
+    val putItems =  if(alreadyExistingPostTimeSeriesItems.isEmpty){
```
wjoel: `if (...) {`, I'm surprised IntelliJ and/or scalafmt doesn't catch this?

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -87,13 +91,42 @@ class TimeSeriesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+      IO.unit
+    } else {
+      put(config.apiKey, baseTimeSeriesUrl(config.project), alreadyExistingPostTimeSeriesItems, maxRetries)
+    }
+
+    val nonExistingPostTimeSeriesItems = timeSeriesItems.filter(p => notExistingNames.contains(p.name))
+    val postItems = if( nonExistingPostTimeSeriesItems.isEmpty ) {
```
wjoel: `if (...) {`

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -87,13 +91,42 @@ class TimeSeriesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
-    post(config.apiKey, baseTimeSeriesUrl(config.project), timeSeriesItems, maxRetries)
-      .map(item => {
+    putOr(config.apiKey, baseTimeSeriesUrl(config.project), timeSeriesItems, maxRetries) {
+      case Response(Right(body), StatusCodes.NotFound, _, _, _) =>
+        decode[Error[TimeSeriesNamesNotFound]](body) match {
+          case Right(conflict) => resolveConflict(timeSeriesItems, conflict.error)
+          case Left(error) => IO.raiseError(error)
+        }
+    }.flatTap { _ =>
+      IO {
```
wjoel: We should only increase this metric if time series were created, and not do so if they were just updated. Maybe that's still what happens, but I'd like us to be sure about this.

src/test/scala/com/cognite/spark/datasource/TimeSeriesUpsertTest.scala
```diff
@@ -0,0 +1,114 @@
...
+    val namesList = namesDf.rdd.map(r => r(0)).collect().toList
+    cleanupTimeSeries(namesList)
+
+    val testTimeSeriesAfterCleanup = retryUntil[DataFrame](
+      spark.sql(s"""select * from destinationTimeSeries where unit = 'test'"""),
+      df => df.count > 0)
+    assert(testTimeSeriesAfterCleanup.count == 0)
+  }
+
+  def cleanupTimeSeries(names: List[Any]): Unit = {
```
wjoel: Should probably be something else than `Any`?

src/test/scala/com/cognite/spark/datasource/TimeSeriesUpsertTest.scala
```diff
@@ -0,0 +1,114 @@
...
+import org.scalatest.FlatSpec
+import org.apache.spark.sql.functions.col
+import com.softwaremill.sttp._
+
+class TimeSeriesUpsertTest extends FlatSpec with SparkTest {
+  val readApiKey = System.getenv("TEST_API_KEY_READ")
+  val writeApiKey = System.getenv("TEST_API_KEY_WRITE")
+
+  it should "successfully both update and insert time series" taggedAs WriteTest in {
+    val initialDescription = """'upsert testing'"""
```
wjoel: Do we need those to be triple quotes? Is the single quote within the string important? Seems like a normal string could do, but maybe we want to test the support for single quotes or something?

src/test/scala/com/cognite/spark/datasource/TimeSeriesUpsertTest.scala
```diff
@@ -0,0 +1,114 @@
...
+    val project = getProject(writeApiKey, Constants.DefaultMaxRetries, Constants.DefaultBaseUrl)
+
+    for (name <- names) {
+      delete(
+        writeApiKey,
+        uri"${Constants.DefaultBaseUrl}/api/0.5/projects/$project/timeseries/$name",
+        maxRetries = 10
+      ).unsafeRunSync()
+    }
+
```
wjoel: Remove empty line

src/test/scala/com/cognite/spark/datasource/TimeSeriesUpsertTest.scala
```diff
@@ -0,0 +1,114 @@
...
+  }
+
```
wjoel: Remove empty line

src/test/scala/com/cognite/spark/datasource/TimeSeriesUpsertTest.scala
```diff
@@ -0,0 +1,114 @@
...
+         |id,
+         |createdTime,
+         |lastUpdatedTime
+         |from sourceTimeSeries
+         |limit 5
+     """.stripMargin)
+      .select(destinationDf.columns.map(col): _*)
+      .write
+      .insertInto("destinationTimeSeries")
+
```
wjoel: Remove empty line

src/test/scala/com/cognite/spark/datasource/TimeSeriesUpsertTest.scala
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  def cleanupTimeSeries(names: List[String]): Unit = {
+    val project = getProject(writeApiKey, Constants.DefaultMaxRetries, Constants.DefaultBaseUrl)
+
+    for (name <- names) {
+      delete(
+        writeApiKey,
+        uri"${Constants.DefaultBaseUrl}/api/0.5/projects/$project/timeseries/$name",
+        maxRetries = 10
```
wjoel: Can use a lower value here, I think. Otherwise it takes a long time for the tests to fail when we do some thing wrong, and if we've failed to delete a time series five times it probably won't help to try another five times.

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -87,13 +90,48 @@ class TimeSeriesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+    } else {
+      post(
+        config.apiKey,
+        baseTimeSeriesUrl(config.project),
+        nonExistingPostTimeSeriesItems,
+        maxRetries)
+    }
+
+    (putItems, postItems).parMapN((_, _) => ())
+
```
wjoel: Remove empty line

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -87,13 +90,48 @@ class TimeSeriesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+      }
+    }
+  }
+
+  def resolveConflict(
+      timeSeriesItems: Seq[PostTimeSeriesItem],
+      timeSeriesConflict: TimeSeriesNotFound): IO[Unit] = {
+    implicit val contextShift: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
+    val notFound = timeSeriesConflict.notFound
+
```
wjoel: Remove empty line

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -38,7 +49,34 @@ case class PostTimeSeriesItem(
...
+    metadata: Option[Setter[Map[String, String]]],
+    unit: Option[Setter[String]],
+    assetId: Option[Setter[Long]],
+    description: Option[Setter[String]],
+    securityCategories: Option[Map[String, Option[Vector[Long]]]],
+    isString: Option[Setter[Boolean]],
+    isStep: Option[Setter[Boolean]]
+)
+object UpdateTimeSeriesItem {
+  def fromPostTimeSeriesItem(postTimeSeriesItem: PostTimeSeriesItem): UpdateTimeSeriesItem =
```
wjoel: Might be reasonable to define this as an `apply()`?

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -38,7 +49,34 @@ case class PostTimeSeriesItem(
...
+    new UpdateTimeSeriesItem(
+      postTimeSeriesItem.id,
+      Setter[String](Some(postTimeSeriesItem.name)),
+      Setter[Map[String, String]](postTimeSeriesItem.metadata),
+      Setter[String](postTimeSeriesItem.unit),
+      Setter[Long](postTimeSeriesItem.assetId),
+      description = Setter[String](postTimeSeriesItem.description),
+      None,
```
wjoel: Why is this (securityCategories) always set to `None`? Can't be changed? Worth a comment at least.

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -82,18 +120,62 @@ class TimeSeriesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+    } else {
+      post(config.apiKey, timeSeriesUrl, timeSeriesToCreate, maxRetries)
+        .flatTap { _ =>
+          IO {
+            if (config.collectMetrics) {
+              timeSeriesCreated.inc(timeSeriesToCreate.length)
+            }
+          }
+        }
+
```
wjoel: Remove empty line ;)

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -82,18 +120,62 @@ class TimeSeriesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+      batches.parTraverse(updateRowsZeroSix).unsafeRunSync()
     })
 
-  private def postRows(rows: Seq[Row]): IO[Unit] = {
-    val timeSeriesItems = rows.map(r => fromRow[PostTimeSeriesItem](r))
-    post(config.apiKey, baseTimeSeriesUrl(config.project), timeSeriesItems, maxRetries)
-      .map(item => {
-        if (config.collectMetrics) {
-          timeSeriesCreated.inc(rows.length)
+  private def updateRowsZeroSix(rows: Seq[Row]): IO[Unit] = {
```
wjoel: I don't think we need to call it out that it's "zero six" unless we also have something that is not "zero six" as well.

Rename it to "updateTimeSeries" or something and do the `rows.map(r => fromRow[PostTimeSeriesItem](r))` at the caller site: that's the only thing rows is being used for in this method so it's better specify that it's about PostTimeSeriesItem in the method signature.

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -82,18 +120,62 @@ class TimeSeriesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+    val notFound = duplicateIds.notFound
+    val timeSeriesToUpdate = updateTimeSeriesItems.filter(p => !notFound.contains(p.id))
+    val putItems = if (timeSeriesToUpdate.isEmpty) {
+      IO.unit
+    } else {
+      put(config.apiKey, updateTimeSeriesUrl, timeSeriesToUpdate, maxRetries)
+    }
+
+    val timeSeriesToCreate = timeSeriesItems.filter(p => notFound.contains(p.id))
+    val postedItems = if (timeSeriesToCreate.isEmpty) {
```
wjoel: Just like `putItems`, this is an IO action, so it should be `postItems` (that's what it does when the action is evaluated)

Issue #175, opened 2019-03-11
Data points relation attempts to post with empty body


wjoel:
@BergsethCognite is trying to post some data points and is getting an error response from CDP: `com.cognite.spark.datasource.CdpApiException: Request to https://api.cognitedata.com/api/0.5/projects/akso/timeseries/data failed with status 400: Missing body for POST request`

This shouldn't happen since the source DataFrame consists of data points (seemingly valid?) for a single time series.

Figure out how this can happen, write a test that fails on this, fix the problem and make the test pass.

wjoel:
Fixed by https://github.com/cognitedata/cdp-spark-datasource/pull/177

PR #177, opened 2019-03-11
Fix posting of data points

wjoel:
TODO: Add a test, since this really should have been caught by a test before being shipped.


wjoel:
Fixes https://github.com/cognitedata/cdp-spark-datasource/issues/175

wjoel:
Made it so that a test that catches this (should get a 404 when posting data points to non-existing time series, not a 400).

PR #185, opened 2019-03-14
Include all fields from API

hakontro:
Add some fields from data sources that have been left out. Events used
0.5 for updates, that had to be upgraded to 0.6 to allow for more fields,
which also gave need for the UpdateEventItem.

### DIFF
--- a/src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
+++ b/src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
@@ -17,13 +17,31 @@ import scala.concurrent.ExecutionContext
 case class PostAssetsDataItems[A](items: Seq[A])
 
 case class AssetsItem(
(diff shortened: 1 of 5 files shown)


src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
```diff
@@ -17,13 +17,31 @@ import scala.concurrent.ExecutionContext
...
-    id: Long)
+    source: Option[String],
+    sourceId: Option[String],
+    createdTime: Long,
+    lastUpdatedTime: Long)
 
-case class PostAssetsItem(name: String, description: String, metadata: Map[String, String])
+case class PostAssetsItem(
+    name: String,
+    redIf: Option[String],
```
wjoel: Should this be `refId`?

Issue #184, opened 2019-03-14
Add filter push down support for Raw lastChanged


serenityue:

Tapped:
I'll steal this issue Håkon

hakontro:
No problem!

PR #186, opened 2019-03-14
Add push down filter for raw lastChanged

serenityue:
#184

### DIFF
--- a/src/main/scala/com/cognite/spark/datasource/DataPointsRelation.scala
+++ b/src/main/scala/com/cognite/spark/datasource/DataPointsRelation.scala
@@ -81,7 +81,8 @@ abstract class DataPointsRelation(
       case GreaterThanOrEqual("timestamp", value) => Seq(Min(value.toString.toLong))
       case And(f1, f2) => getTimestampLimit(f1) ++ getTimestampLimit(f2)
       // case Or(f1, f2) => we might possibly want to do something clever with joining an "or" clause
-      //                    with timestamp limits on each side; just ignore them for now
+      //                    with timestamp limits on each side (including replacing "max of Min's" with the less strict
+      //                    "min of Min's" when aggregating filters on the same side); just ignore them for now
       case _ => Seq.empty
     }
 
@@ -161,8 +162,10 @@ abstract class DataPointsRelation(
     }
 
     Tuple2(
-      Try(timestampLimits.filter(_.isInstanceOf[Min]).min).toOption.map(_.value),
-      Try(timestampLimits.filter(_.isInstanceOf[Max]).max).toOption.map(_.value))
+      // Note that this way of aggregating filters will not work with "Or" predicates.
+      Try(timestampLimits.filter(_.isInstanceOf[Min]).max).toOption.map(_.value),
+      Try(timestampLimits.filter(_.isInstanceOf[Max]).min).toOption.map(_.value)
+    )
   }
 
   def postTimeSeries(data: MultiNamedTimeseriesData): IO[Unit] = {
--- a/src/main/scala/com/cognite/spark/datasource/RawTableRelation.scala
+++ b/src/main/scala/com/cognite/spark/datasource/RawTableRelation.scala
@@ -10,13 +10,15 @@ import io.circe.generic.auto._
 import io.circe.syntax._
 import org.apache.spark.datasource._
 import org.apache.spark.rdd.RDD
-import org.apache.spark.sql.sources.{BaseRelation, InsertableRelation, TableScan}
+import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types.{DataTypes, StructField, StructType}
 import org.apache.spark.sql.{DataFrame, Row, SQLContext}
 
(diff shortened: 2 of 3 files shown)


serenityue:
I haven't proof-read my code and haven't fixed the tests.

src/main/scala/com/cognite/spark/datasource/DataPointsRelation.scala
```diff
@@ -258,8 +258,8 @@ class DataPointsRelation(
     }
 
     Tuple2(
-      Try(timestampLimits.filter(_.isInstanceOf[Min]).min).toOption.map(_.value),
-      Try(timestampLimits.filter(_.isInstanceOf[Max]).max).toOption.map(_.value))
+      Try(timestampLimits.filter(_.isInstanceOf[Min]).max).toOption.map(_.value),
```
wjoel: Shouldn't we have the minimum of the minimums, and maximum of the maximums?

serenityue: If users want "select * from _cdp.datapoints where name=<something> and timestamp > 1 and timestamp > 2 ", we should give them timestamp > 2. So that's the max of min.

wjoel: Consider `select * from _cdp.datapoints where name=some_name and timestamp > 1 or timestamp > 2`. It's better to return too much data, than to return too little data.

serenityue: Hmm... Currently in getTimestampLimit(filters), OR is ignored, only AND is parsed. And adding support for OR will need more work than using min of min and max of max, since we can have "timestamp > num1 or timestamp < num2"

wjoel: That's a good point. Still, I don't see any harm in using the safer choice of the minimum min, even if it ends up returning a bit of extra data in your example. Spark will always do its own filtering afterwards.

If you really want to leave this in I'm ok with it as well, but then it should be commented and a note should be added to the commented out OR parsing, warning that this needs to be changed when/if we add support for OR.

src/main/scala/com/cognite/spark/datasource/RawTableRelation.scala
```diff
@@ -143,18 +190,22 @@ class RawTableRelation(
 }
 
 object RawTableRelation {
+  val lastChangedColName = "lastChanged"
```
wjoel: Can/should be private?

src/main/scala/com/cognite/spark/datasource/RawTableRelation.scala
```diff
@@ -163,54 +214,70 @@ object RawTableRelation {
...
+    import org.apache.spark.sql.functions.col
+    val dfWithSchema =
+      df.select(
+        $"key",
+        col(lastChangedColName),
+        from_json($"columns", jsonFieldsSchema).alias("columns"))
 
-    if (keyColumns(jsonFieldsSchema).isEmpty) {
-      dfWithSchema.select("key", "columns.*")
+    if (keyColumns(jsonFieldsSchema).isEmpty && lastChangedColumns((jsonFieldsSchema)).isEmpty) {
```
wjoel: Remove extra parenthesis in `((jsonFieldsSchema))`

src/main/scala/com/cognite/spark/datasource/RawTableRelation.scala
```diff
@@ -163,54 +214,70 @@ object RawTableRelation {
...
+  private def unRenameColumns(df: DataFrame): DataFrame = {
+    val columnsToRename = keyColumns(df.schema) ++ lastChangedColumns(df.schema)
+    // when renaming them back we instead start with the shortest column name, for similar reasons
+    columnsToRename.sortWith(_ > _).foldLeft(df) { (df, column) =>
+      df.withColumnRenamed(column, column.substring(1))
     }
   }
 
   private val temporaryKeyName = s"TrE85tFQPCb2fEUZ"
+  private val temporaryLastChangedName = s"TrE85tFQPCb2fEUZnand1061"
```
wjoel: Make this something completely unique. Here you go (from `pwgen -s 16`): J2p972xzM9bf32oD

src/test/scala/com/cognite/spark/datasource/RawTableRelationTest.scala
```diff
@@ -145,4 +228,39 @@ class RawTableRelationTest extends FlatSpec with Matchers with SparkTest {
...
+      .option("apiKey", writeApiKey)
+      .option("type", "raw")
+      .option("database", "testdb")
+      .option("table", "raw-write-test")
+      .option("inferSchema", false)
+      .load()
+    destinationDf.createTempView("destinationTable")
+
+    spark
+      .sql(s"""
```
wjoel: Formatting looks off here.

Tapped:
Thanks for the review, your suggestions are now fixed.

PR #187, opened 2019-03-18
Remove period from the metric prefix

Tapped:
We now end up with double period in the metrics, since https://github.com/cognitedata/cdp-spark-datasource/blob/master/src/main/scala/com/cognite/spark/datasource/MetricsSource.scala#L52 also adds a period after the `metricnamespace`.


Tapped:
Do you know why the timeseries test are breaking?

PR #190, opened 2019-03-18
MetricsSource: Remove period

Tapped:
Spark is automatically adding period before the metric names. So now we get `executorId..someMetric`.


Tapped:
Bump

PR #197, opened 2019-03-19
Add filter pushdown support for events

hakontro:


src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -136,13 +139,111 @@ class EventsRelation(config: RelationConfig)(@transient val sqlContext: SQLConte
     (postUpdate, postNewItems).parMapN((_, _) => ())
   }
 
+  sealed case class TypeFilter(`type`: String)
+
+  // scalastyle:off cyclomatic.complexity
+  def getTypeFilters(filter: Filter): Seq[TypeFilter] =
```
wjoel: Is it possible to combine the two by making it a function of `(filter: Filter, columnName: String, mkFilter: (String) => ColumnFilter)` (not tested) or similar? And make `TypeFilter` and `SubtypeFilter` subclasses of `ColumnFilter`.

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -136,13 +139,111 @@ class EventsRelation(config: RelationConfig)(@transient val sqlContext: SQLConte
...
+  override def buildScan(): RDD[Row] = buildScan(Array.empty, Array.empty)
+
+  override def buildScan(requiredColumns: Array[String], filters: Array[Filter]): RDD[Row] = {
+    val types = filters.flatMap(getTypeFilters).map(_.`type`)
+    val subTypes = filters.flatMap(getSubTypeFilters).map(_.subType)
+
+    val params = Map("type" -> types.toString, "subtype" -> subTypes.toString)
+
+    val rdds: Array[CdpRdd[EventItem]] = for {
+      eventType <- if (types.isEmpty) { Array(None) } else { types }
```
wjoel: Do we need the "if" here? Is it to get the right type? I'd have thought that `eventType <- types` would give a None if `types.isEmpty`. I haven't checked though. ;)

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -136,13 +139,111 @@ class EventsRelation(config: RelationConfig)(@transient val sqlContext: SQLConte
...
+      eventSubType <- if (subTypes.isEmpty) { Array(None) } else { subTypes }
+    } yield {
+      val urlWithType =
```
wjoel: `eventType.map(t => listUrl(config).param("type")).getOrElse(listUrl(config))` I think.

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -136,13 +139,111 @@ class EventsRelation(config: RelationConfig)(@transient val sqlContext: SQLConte
...
+          }
+          toRow(e, requiredColumns)
+        },
+        urlWithTypeAndSubType.param("onlyCursors", "true"),
+        urlWithTypeAndSubType,
+        config
+      )
+    }
+
+    rdds.foldLeft(sqlContext.sparkContext.emptyRDD[Row])((a, b) => a.union(b))
```
wjoel: We should try to get rid of those `.unions`, they create potentially large DAGs that I suspect may have a performance impact. Can probably be fixed by having smarter custom `RDD` implementations - we shoudln't fix it here and now but I just wanted to point this out.

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
```diff
@@ -22,6 +22,24 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
...
+      .where(s"type = 'future'")
+    assert(df.count == 1000)
+  }
+
+  it should "apply multiple pushdown filters" taggedAs WriteTest in {
+    val df = spark.read.format("com.cognite.spark.datasource")
+      .option("apiKey", writeApiKey)
+      .option("type", "events")
+      .load()
+      .where(s"type = '***' and subtype = 'Valhall'")
```
wjoel: Add a test for "or" as well, and a couple of tests doing multiple types and subtypes.

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
wjoel: Or even better, `eventType.fold(listUrl(config))(listUrl(config).param("type", _))` similar to https://github.com/cognitedata/cdp-spark-datasource/blob/0582083db1b0411898357cecc8f53f74a9e375d4/src/main/scala/com/cognite/spark/datasource/CdpRdd.scala#L41

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
hakontro: For will yield nothing if one or more of the Arrays are empty, so this works around that by giving an Array(none).

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
hakontro: Very nice! Thanks

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -164,13 +167,80 @@ class EventsRelation(config: RelationConfig)(@transient val sqlContext: SQLConte
...
   override def schema: StructType = structType[EventItem]
 
   override def toRow(t: EventItem): Row = asRow(t)
 
+  def toRow(t: EventItem, requiredColumns: Array[String]): Row = {
+    val values = t.productIterator
+    val eventMap = t.getClass.getDeclaredFields.map(_.getName -> values.next).toMap
+
+    Row.fromSeq(requiredColumns.map(eventMap(_)).toSeq)
+
```
wjoel: Move `}` from the next line to this line.

PR #201, opened 2019-03-20
Fix error in assetIds implementation for Events

hakontro:


src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -48,7 +48,7 @@ object UpdateEventItem {
       Setter[String](eventItem.`type`),
       Setter[String](eventItem.subtype),
       Setter[Map[String, String]](eventItem.metadata),
-      Setter[Seq[Long]](eventItem.assetIds),
+      eventItem.assetIds.map(a => Map("set" -> a)),
```
wjoel: Interesting, I see we already have this for `securityCategories` in time series metadata as well.

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
```diff
@@ -96,7 +96,7 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
                  |endTime,
                  |type,
                  |subtype,
-                 |null as assetIds,
+                 |array(2091657868296883) as assetIds,
```
wjoel: Maybe set it to "null" here, then change it in the "Update events" step, and then check the results to ensure they have the new value in the following tests?

hakontro: Good idea, I'll implement that.

PR #206, opened 2019-03-22
Add support for string data points

wjoel:


src/main/scala/com/cognite/spark/datasource/NumericDataPointsRdd.scala
```diff
@@ -0,0 +1,123 @@
...
+          .unsafeRunSync()
+    }
+    if (dataPoints.lastOption.fold(true)(_.timestamp < start)) {
+      (Seq.empty, None)
+    } else {
+      (dataPoints.map(toRow), dataPoints.lastOption.map(_.timestamp + granularityMilliseconds))
+    }
+  }
+
+  def parseResult(response: Response[Array[Byte]]): Response[Seq[NumericDatapoint]] = {
```
hakontro: The implementation of parseResult is quite similar in NumericDataPointsRdd and StringDataPointsRdd, we could consider implementing it in the superclass.

src/main/scala/com/cognite/spark/datasource/DataPointsRelation.scala
```diff
@@ -371,7 +155,11 @@ class DataPointsRelation(
       .map(r => {
         if (config.collectMetrics) {
           val numPoints = data.namedTimeseriesData
-            .map(_.getNumericData.points.length)
+            .map(
+              ts =>
+                ts.data.numericData
+                  .map(_.points.length)
```
hakontro: Can the flatMap at the end of this method be removed?

wjoel: It can! Well spotted, thanks.

src/main/scala/com/cognite/spark/datasource/NumericDataPointsRdd.scala
wjoel: Yes, it's a bit annoying, but I couldn't find a great way to do it: the obvious way would require a type parameter for DataPointsRdd which isn't nice either. I'm going to leave it like this for now.

PR #215, opened 2019-03-28
Fix bug of same metric registered multiple times

serenityue:


src/main/scala/com/cognite/spark/datasource/MetricsSource.scala
```diff
@@ -1,34 +1,20 @@
...
 
 import scala.collection.JavaConverters._
-import java.util.concurrent.ConcurrentHashMap
 
 import com.codahale.metrics._
 import org.apache.spark._
 
-class MetricsSource(val metricNamespace: String) {
+class MetricsSource(val metricNamespace: String) extends Serializable {
+  def getOrElseUpdate(metricName: String, metric: Metric): Metric = {
```
wjoel: I would be very careful about removing the `ConcurrentHashMap` and the lazy wrapper. They were added to fix concurrency bugs that were caught by tests: https://github.com/cognitedata/spark-metrics/commit/4ca5723c6b8213e086b411c70a521a3d329ac622
and https://github.com/groupon/spark-metrics/pull/24 with a new attempt in https://github.com/groupon/spark-metrics/pull/28

serenityue: I've created a new PR (https://github.com/cognitedata/cdp-spark-datasource/pull/217) for a thread-safe fix.

Issue #117, opened 2019-02-06
Have different schema for reading and writing


Tapped:
Writing resources to CDP got different schema than reading them. We should make that clearer by also have different schema for reading and writing in the spark datasource. 

Plan is to add a option called `mode` which can either be read or write, and by default is read. This will not be a backcompatible change.

hakontro:
#216

PR #216, opened 2019-03-29
Add support for writing in save mode

hakontro:
Work in progress, based on Emil's work in #142 and #117.

Implemented three different save modes: write, upsert and update, provided as .option("savemode", "upsert"). Could help users avoid making errors, what do you think?

Declaring EventsInsertion inside EventsRelation is temporary, we should probably extract common methods in a superclass or similar.

### DIFF
--- a/src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
+++ b/src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
@@ -1,6 +1,6 @@
 package com.cognite.spark.datasource
+import cats.effect.IO
 import org.apache.spark.sql.sources.{BaseRelation, TableScan}
-import io.circe.generic.auto._
 import io.circe.generic.decoding.DerivedDecoder
 import com.softwaremill.sttp.Uri
 import org.apache.spark.datasource.MetricsSource
@@ -49,4 +49,21 @@ abstract class CdpRelation[T: DerivedDecoder: TypeTag: ClassTag](
   def listUrl(): Uri
 
   def cursors(): Iterator[(Option[String], Option[Int])] = NextCursorIterator(listUrl(), config)
+
+  def insert(rows: Seq[Row]): IO[Unit] =
+    throw new IllegalArgumentException(
+      s"""${shortName} does not support the "onconflict" option "abort".""")
+
+  def upsert(rows: Seq[Row]): IO[Unit] =
+    throw new IllegalArgumentException(
+      s"""${shortName} does not support the "onconflict" option "upsert".""")
+
+  def update(rows: Seq[Row]): IO[Unit] =
+    throw new IllegalArgumentException(
+      s"""${shortName} does not support the "onconflict" option "update".""")
+
+  def delete(rows: Seq[Row]): IO[Unit] =
+    throw new IllegalArgumentException(
+      s"""${shortName} does not support the "onconflict" option "delete".""")
+
 }
--- a/src/main/scala/com/cognite/spark/datasource/DefaultSource.scala
+++ b/src/main/scala/com/cognite/spark/datasource/DefaultSource.scala
@@ -1,13 +1,18 @@
 package com.cognite.spark.datasource
 
-import org.apache.spark.sql.SQLContext
+import org.apache.spark.sql.{DataFrame, SQLContext, SaveMode}
 import org.apache.spark.sql.sources.{
   BaseRelation,
+  CreatableRelationProvider,
   DataSourceRegister,
   RelationProvider,
   SchemaRelationProvider
 }
 import org.apache.spark.sql.types.StructType
+import cats.effect.{ContextShift, IO}
+import cats.implicits._
+
+import scala.concurrent.ExecutionContext
 
 case class RelationConfig(
     apiKey: String,
@@ -18,10 +23,17 @@ case class RelationConfig(
     maxRetries: Int,
     collectMetrics: Boolean,
     metricsPrefix: String,
-    baseUrl: String)
+    baseUrl: String,
+    onConflict: OnConflict.Value)
+
+object OnConflict extends Enumeration {
+  type Mode = Value
+  val ABORT, UPDATE, UPSERT, DELETE = Value
+}
 
 class DefaultSource
     extends RelationProvider
+    with CreatableRelationProvider
     with SchemaRelationProvider
     with DataSourceRegister
     with CdpConnector {
@@ -70,6 +82,8 @@ class DefaultSource
       case None => ""
     }
     val collectMetrics = toBoolean(parameters, "collectMetrics")
+    val saveMode =
+      OnConflict.withName(parameters.getOrElse("onconflict", "ABORT").toUpperCase())
     RelationConfig(
       apiKey,
       project,
@@ -79,7 +93,8 @@ class DefaultSource
       maxRetries,
       collectMetrics,
       metricsPrefix,
-      baseUrl)
+      baseUrl,
+      saveMode)
   }
 
   // scalastyle:off cyclomatic.complexity method.length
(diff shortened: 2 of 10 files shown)


Tapped:
It's great, love it ❤️

src/main/scala/com/cognite/spark/datasource/DefaultSource.scala
```diff
@@ -158,5 +167,38 @@ class DefaultSource
       case _ => sys.error("Unknown resource type: " + resourceType)
     }
   }
+
+  override def createRelation(
+      sqlContext: SQLContext,
+      mode: SaveMode,
```
wjoel: Couldn't we use `SaveMode` instead of `.option("savemode", "our-thing")`?

Seems `SaveMode.Overwrite` maps well enough to upsert. `SaveMode.Append` may not make sense for our data. `SaveMode.Ignore` would be "don't do upsert on conflict, just leave it how it is" I guess. The default `SaveMode.ErrorIfExists` would be "don't do upsert and just abort on conflict".

hakontro: SaveMode works as well, but we can't define new `SaveMode`s (at least I couldn't find a way to do it). With `option()` we can get the exact names, which I think is clearer for the user.

hakontro:
Savemodes are now defined in CdpRelation, to be overridden in resource relations. If this look ok I'll start implementing this for the rest of the resource types, so any feedback on the overall structure very welcome.

src/main/scala/com/cognite/spark/datasource/DefaultSource.scala
wjoel: Maybe using `SaveMode` if "savemode" isn't defined? This should be documented, of course. :)

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
```diff
@@ -49,4 +50,17 @@ abstract class CdpRelation[T: DerivedDecoder: TypeTag: ClassTag](
   def listUrl(): Uri
 
   def cursors(): Iterator[(Option[String], Option[Int])] = NextCursorIterator(listUrl(), config)
+
+  def insert(rows: Seq[Row]): IO[Unit] =
+    throw new IllegalArgumentException(s"${shortName} does not support inserts in save mode.")
```
wjoel: > does not support the "savemode" option of type "insert"

or something like that instead.

src/main/scala/com/cognite/spark/datasource/DefaultSource.scala
```diff
@@ -18,10 +23,12 @@ case class RelationConfig(
     maxRetries: Int,
     collectMetrics: Boolean,
     metricsPrefix: String,
-    baseUrl: String)
+    baseUrl: String,
+    saveMode: String)
```
wjoel: This should be an enum/sealed trait and case class instead of a string.

src/main/scala/com/cognite/spark/datasource/DefaultSource.scala
```diff
@@ -158,5 +167,44 @@ class DefaultSource
...
+        new TimeSeriesRelation(config)(sqlContext)
+      case _ => sys.error(s"Resource type '$resourceType does not support savemode")
+    }
+
+    data.foreachPartition(rows => {
+      implicit val contextShift: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
+      val batches = rows.grouped(Constants.DefaultBatchSize).toVector
+
+      config.saveMode match {
+        case "insert" => {
```
wjoel: No need for braces in those case statements.

src/main/scala/com/cognite/spark/datasource/DefaultSource.scala
```diff
@@ -158,5 +167,44 @@ class DefaultSource
...
+        case "upsert" => {
+          batches.parTraverse(relation.upsert).unsafeRunSync()
+        }
+        case "update" => {
+          batches.parTraverse(relation.update).unsafeRunSync()
+        }
+        case "delete" => {
+          batches.parTraverse(relation.delete).unsafeRunSync()
+        }
+        case _ => sys.error("Unknown savemode: " + config.saveMode)
```
wjoel: We should use enums instead and give an error when the user attempts to create a relation with an unsupported savemode, rather than waiting until they try to write to the relation.

src/main/scala/com/cognite/spark/datasource/DefaultSource.scala
```diff
@@ -70,6 +77,7 @@ class DefaultSource
       case None => ""
     }
     val collectMetrics = toBoolean(parameters, "collectMetrics")
+    val saveMode = parameters.getOrElse("savemode", "write")
```
wjoel: I don't see "write" in the `config.saveMode match` below, shouldn't this be set to one of the values in that `match`?

src/test/scala/com/cognite/spark/datasource/DataPointsRelationTest.scala
```diff
@@ -241,6 +241,7 @@ class DataPointsRelationTest extends FlatSpec with Matchers with SparkTest {
       .load()
     destinationDf.createTempView("destinationDatapoints")
 
+    spark.sparkContext.setLogLevel("OFF")
```
wjoel: I understand why right now but we should add a comment and explain the why, because future developers (possibly including me!) will not understand this. :)

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
```diff
@@ -196,6 +191,175 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
     assert(dfWithCorrectAssetIds.count == 1000)
   }
 
+  it should "allow safe inserts in savemode" taggedAs WriteTest in {
```
wjoel: What's a "safe insert"?

src/test/scala/com/cognite/spark/datasource/TimeSeriesUpsertTest.scala
```diff
@@ -91,6 +92,191 @@ class TimeSeriesUpsertTest extends FlatSpec with SparkTest {
...
+      spark.sql(s"""
+              |select '$insertDescription' as description,
+              |isString,
+              |concat('TEST_', name) as name,
+              |metadata,
+              |'$saveModeUnit' as unit,
+              |NULL as assetId,
+              |isStep,
+              |cast(array() as array<long>) as securityCategories,
+              |id+2 as id,
```
wjoel: Isn't id ignored anyway on inserts? The user can't set this. Clearer to use `null` here as well, I think?

src/test/scala/com/cognite/spark/datasource/TimeSeriesUpsertTest.scala
```diff
@@ -91,6 +92,191 @@ class TimeSeriesUpsertTest extends FlatSpec with SparkTest {
     assert(initialDescriptionsAfterUpsert.count == 0)
   }
 
+  it should "support inserts, updates and upserts in savemode" taggedAs WriteTest in {
```
wjoel: I think we should split those into separate tests, it would make it easier to pinpoint what went wrong if one case fails.

src/test/scala/com/cognite/spark/datasource/TimeSeriesUpsertTest.scala
hakontro: Good catch, turns out id could be deleted from the PostTimeSeriesItem class as well.

hakontro:
> I wonder if we should call it "mode" instead of "savemode"? It's not quite the same as `SaveMode` now, and if/when we support delete it wouldn't be saving. On the other hand "mode" is quite generic and I don't love it either. Any other suggestions?
I agree, it could confuse some people. Another thing about it being an option, rather than an argument, is that people might assume the option applies to the InsertInto pattern as well. For that reason I prefer "savemode" over "mode".

src/main/scala/com/cognite/spark/datasource/DefaultSource.scala
hakontro: Write was just a working name I forgot to change :)

src/test/scala/com/cognite/spark/datasource/DataPointsRelationTest.scala
hakontro: Good point

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
hakontro: Another working name I forgot to remove

wjoel:
Some more suggestions (to reduce possible confusion with `SaveMode`): "insertmode", "onconflict" (in which case "insert" should be "abort" instead maybe?), "writemode".

hakontro:
`"onconflict"` is nice, I'll use that.

PR #217, opened 2019-03-29
Fix bug of same metric registered multiple times

serenityue:
Previously when Jetfire was using the new metrics in cdp-spark-datasource, we get errors about registering metrics that already exist, possibly due to Spark creating multiple instances of <DataType>Relation classes (and thus multiple instances of MetricSource)for read and for write.

This fix makes MetricSources a global singleton, so all metrics of all jobs are put into one hashmap. They are later removed when job is finished. This should be thread-safe.

### DIFF
--- a/src/main/scala/com/cognite/spark/datasource/MetricsSource.scala
+++ b/src/main/scala/com/cognite/spark/datasource/MetricsSource.scala
@@ -3,37 +3,28 @@ package org.apache.spark.datasource
 import scala.collection.JavaConverters._
 import java.util.concurrent.ConcurrentHashMap
 
+import cats.Eval
 import com.codahale.metrics._
 import org.apache.spark._
 
-class MetricsSource(val metricNamespace: String) {
-
-  class LazyWrapper[T](wrapped: => T) {
-    lazy val value = wrapped
-  }
-
-  private def wrap[T](value: => T): LazyWrapper[T] =
-    new LazyWrapper[T](value)
-
-  private def unwrap[T](lazyWrapper: LazyWrapper[T]): T =
-    lazyWrapper.value
-
+class MetricsSource {
+  // Add metricNamespace to differentiate with spark system metrics.
   // Keeps track of all the Metric instances that are being published
-  val metricsMap = new ConcurrentHashMap[String, LazyWrapper[Metric]]().asScala
+  val metricsMap = new ConcurrentHashMap[String, Eval[Metric]]().asScala
 
-  def getOrElseUpdate(metricName: String, metric: => Metric): Metric = {
-    val wrapped = wrap(metric)
-    metricsMap.putIfAbsent(metricName, wrapped) match {
+  def getOrElseUpdate(metricNamespace: String, metricName: String, metric: => Metric): Metric = {
+    val wrapped = Eval.later(metric)
+    metricsMap.putIfAbsent(s"${metricNamespace}.${metricName}", wrapped) match {
       case Some(wrappedMetric) => wrappedMetric.value
       case None => {
-        registerMetricSource(metricName, wrapped.value)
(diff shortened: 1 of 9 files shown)


wjoel:
Overall, looks good to me, but the removal of metrics is a Jetfire thing and should stay there. A normal of user of `cdp-spark-datasource` wouldn't want metrics to be cleared immediately after running `df.write.insertInto("something")` finishes. I don't see why it would be necessary in the MetricsSource itself?

I also think the job group id always becoming part of the metric name(s) is a bit awkward - if it works without the job group id (or it being null), which I guess it will be for non-Jetfire use cases, wouldn't it also work if the job group id remains part of the metrics prefix when Jetfire creates those relations as it is today?

serenityue:
For the first comment, if I understand correctly, Jetfire backend can only find the metrics registered in driver, not the workers. I don't actually fully understand why or why not we want this unregister process... For Jetfire, would all metrics, including the workers', persist forever if we don't unregister? And for non-Jetfire, they might register the sink and report metrics to db after spark job has completed? is that why we shouldn't unregister?

For the second comment on job group id, it was done this way to keep consistent with the "remove the entries related to this job from the global hash map (metric name -> Counter) and unregister source", which finds the metric and source by job group id. I didn't find a better way to identify the metrics using only info from SparkListenerJobEnd, and with the original "metricsPrefix" in config, it might be confusing to just magically find metrics by job group id. I agree that using "job group id" instead of "metric prefix" is awkward, but didn't find a better way to do removal of metrics... Suggestions?

serenityue:
An alternative would be to not remove metrics from hash map or unregister. Then we can get rid of everything in spark listener and use "metric prefix" in config. The hash map would grow, but it would take ~50 million jobs to have a memory problem according to Emil's calculation, which I guess shouldn't be a problem?

serenityue:
Updated.

wjoel:
It seems whatever we did in the previous commit to unregister the metrics from the workers should be possible to do in code living in the Jetfire backend instead.

For non-Jetfire uses the job group id would typically not set, so jobs running one after another would increase the same metrics and there shouldn't be any explosion of metrics.

If, in an interactive session, I do this:
```scala
spark.read...option("table", "table1").load().count()
spark.read...option("table", "table2").load().count()
```
then afterwards I should be able to use all the `raw.read` metrics from all the nodes in my cluster to get a count of the number of raw rows that have been read. The metrics shouldn't be removed in this use case (arguably not in any non-Jetfire use case).

If the metrics are cleared after each job is finished they would be gone. On the other hand we have no usage of metrics outside of Jetfire today, which is why another option may be to (within Jetfire) extend the data sources we have in this repo and add instrumentation/metrics within Jetfire.

serenityue:
I see... I've updated the PR (force pushed to one commit).

src/main/scala/com/cognite/spark/datasource/MetricsSource.scala
```diff
@@ -3,37 +3,28 @@ package org.apache.spark.datasource
...
   // Keeps track of all the Metric instances that are being published
-  val metricsMap = new ConcurrentHashMap[String, LazyWrapper[Metric]]().asScala
+  val metricsMap = new ConcurrentHashMap[String, Eval[Metric]]().asScala
 
-  def getOrElseUpdate(metricName: String, metric: => Metric): Metric = {
-    val wrapped = wrap(metric)
-    metricsMap.putIfAbsent(metricName, wrapped) match {
+  def getOrElseUpdate(metricNamespace: String, metricName: String, metric: => Metric): Metric = {
+    val wrapped = Eval.later(metric)
+    metricsMap.putIfAbsent(metricNamespace + "." + metricName, wrapped) match {
```
wjoel: Use string interpolation instead of `+`

PR #221, opened 2019-04-08
Enable cross-building for Scala 2.11 and 2.12

wjoel:


src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
```diff
@@ -47,10 +47,11 @@ class AssetsRelation(config: RelationConfig, assetPath: Option[String])(val sqlC
     MetricsSource.getOrCreateCounter(config.metricsPrefix, s"assets.created")
 
   override def insert(df: org.apache.spark.sql.DataFrame, overwrite: scala.Boolean): scala.Unit =
-    df.foreachPartition(rows => {
+    df.foreachPartition((rows: Iterator[Row]) => {
```
wjoel: Scala 2.12 needs some help to get the right overloaded method of `foreachPartition`.

src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
```diff
@@ -47,10 +47,11 @@ class AssetsRelation(config: RelationConfig, assetPath: Option[String])(val sqlC
...
       implicit val contextShift: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
       val batches = rows.grouped(config.batchSize.getOrElse(Constants.DefaultBatchSize)).toVector
       batches.parTraverse(postRows).unsafeRunSync()
+      ()
```
hakontro: This is unusual, perhaps we should add a comment to explain why it's needed?

Tapped: Hmm, that's the only way I know for how to create an unit type. Though `unsafeRunRync()` should return an unit in this case, I believe.

wjoel: It's part of the "Scala 2.12 needs some help to get the right overloaded method of `foreachPartition`", without this the return type is `Vector[Unit]` and the compiler doesn't know what to do.

`()` explicitly makes it return the required `Unit` type. I don't want to add comments on this everywhere it's done.

PR #223, opened 2019-04-11
Let DataPointsRdd handle multiple time series, avoid DataFrame union

### LINKED ISSUE #222: Avoid using .union for DataPoints
Using "union" in for example data points can easily create very large execution graphs, dramatically slowing down Spark (sometimes causing it to fail) and making the DAG visualization either crash or become very slow.

Rewrite the usages of "union" RDDs to instead use RDDs that can handle multiples of whatever we need the union for. For example, DatapointsRdd could handle multiple time series names.

wjoel:
Fixes #222


wjoel:
Thanks! Yes, I've seen it and it's an interesting alternative, but I'm not sure how much it would help. Seems like it would result in fewer but potentially slower API calls (with less predictable read patterns I think, but not sure how much/if that matters), so we'd have to benchmark it to see if it's actually helpful. It would only be used for aggregates as well, since it doesn't support protobuf.

We should check with the time lords if they think it's a good idea to use it.

PR #227, opened 2019-04-16
Avoid unions in Event pushdown filters

hakontro:


src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -217,30 +218,18 @@ class EventsRelation(config: RelationConfig)(@transient val sqlContext: SQLConte
...
-      )
-    }
-
-    rdds.foldLeft(sqlContext.sparkContext.emptyRDD[Row])((a, b) => a.union(b))
+    } yield { new EventFilter(eventType, eventSubType) }
+
+    EventsRdd(
+      sqlContext.sparkContext,
+      listUrl(),
+      (e: EventItem) => toRow(e, requiredColumns),
```
Tapped: Why was collectMetrics removed?

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -82,6 +82,7 @@ object UpdateEventItem {
 }
 case class SourceWithResourceId(id: Long, source: String, sourceId: String)
 case class EventConflict(duplicates: Seq[SourceWithResourceId])
+case class EventFilter(_type: Option[String], subtype: Option[String])
```
Tapped: Personally I like 
```
`type`
```
better than `_type` but that's subjective.

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
hakontro: Good catch, that wasn't on purpose.

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
hakontro: Looks like we've used `type` elsewhere as well, so I'll stick to that.

hakontro:
As stated in #222 a large number of unions gives pretty messy execution graphs in Spark. While filter pushdown for Events using union isn't quite as bad as DataPoints, since the number of unions will be way less, it can still mess up the UI quite bad.

src/main/scala/com/cognite/spark/datasource/EventsRdd.scala
```diff
@@ -0,0 +1,59 @@
...
+    extends RDD[Row](sparkContext, Nil)
+    with CdpConnector {
+
+  override def getPartitions: Array[Partition] =
+    (for {
+      cursor <- cursors
+      eventFilter <- eventFilters
+    } yield (cursor, eventFilter)).zipWithIndex.map {
+      case (((cursor, size), eventFilter), index) =>
+        val partitionSize = (config.limit, size) match {
```
wjoel: This is taken care of in `Batch.withCursor`, or maybe somewhere else, I think...? In this particular case, however, "size" will always be set to None since the cursors are from the cursors endpoint.

src/main/scala/com/cognite/spark/datasource/EventsRdd.scala
```diff
@@ -0,0 +1,59 @@
...
+    toRow: EventItem => Row,
+    eventFilters: Array[EventFilter],
+    config: RelationConfig,
+    cursors: Iterator[(Option[String], Option[Int])])
+    extends RDD[Row](sparkContext, Nil)
+    with CdpConnector {
+
+  override def getPartitions: Array[Partition] =
+    (for {
+      cursor <- cursors
```
wjoel: We can do destructuring here:

```scala
  override def getPartitions: Array[Partition] =
    (for {
      (cursor, size) <- cursors
      eventFilter <- eventFilters
    } yield EventsRddPartition(cursor, size, 1, eventFilter)).zipWithIndex.map {
      case (p, idx) => p.copy(index = idx)
    }.toArray
```

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -217,30 +218,17 @@ class EventsRelation(config: RelationConfig)(@transient val sqlContext: SQLConte
...
-            eventsRead.inc()
-          }
-          toRow(e, requiredColumns)
-        },
-        urlWithTypeAndSubType,
-        config,
-        cursors()
-      )
-    }
+    } yield { new EventFilter(eventType, eventSubType) }
```
wjoel: Remove `new` (don't need it for case classes) and the `{` in `yield { ... }`

wjoel:
There's not too much copy-paste tbh, but we could probably restructure a few things and get rid of some duplication (for example having a URL as part of some more generic CdpPartition class), but we can do that later.

PR #229, opened 2019-04-24
Add support for partial updates

hakontro:
@Tapped I couldn't make `val column = q"${typeOf[Try[_]]}(Option($r.getAs[$rowType]($name))).toOption.flatten"` work, so importing Try everywhere.

```
/cdp-spark-datasource/src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala:
121:53: scala.util.Try[_$5] does not take parameters
val timeSeriesItem = fromRow[TimeSeriesItem](r)
                                     ^
```

### DIFF
--- a/src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
+++ b/src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
@@ -18,10 +18,10 @@ import scala.concurrent.ExecutionContext
 case class PostAssetsDataItems[A](items: Seq[A])
 
 case class AssetsItem(
-    id: Option[Long],
+    id: Long,
     path: Option[Seq[Long]],
     depth: Option[Long],
-    name: Option[String],
+    name: String,
     parentId: Option[Long],
     description: Option[String],
     metadata: Option[Map[String, String]],
@@ -39,12 +39,53 @@ case class PostAssetsItem(
     metadata: Option[Map[String, String]]
 )
 
-class AssetsRelation(config: RelationConfig, assetPath: Option[String])(val sqlContext: SQLContext)
+// This class is needed to enable partial updates
+// since AssetsItem has values that are not optional
+case class UpdateAssetsItemBase(
+    id: Option[Long],
+    name: Option[String],
+    description: Option[String],
+    metadata: Option[Map[String, String]],
+    source: Option[String],
+    sourceId: Option[String]
+)
+
+case class UpdateAssetsItem(
+    id: Option[Long],
+    name: Option[NonNullableSetter[String]],
+    description: Option[Setter[String]],
+    metadata: Option[Setter[Map[String, String]]],
+    source: Option[Setter[String]],
+    sourceId: Option[Setter[String]]
+)
+object UpdateAssetsItem {
+  def apply(assetItem: UpdateAssetsItemBase): UpdateAssetsItem =
+    new UpdateAssetsItem(
+      assetItem.id,
+      assetItem.name.map(NonNullableSetter(_)),
+      Setter[String](assetItem.description),
+      Setter[Map[String, String]](assetItem.metadata),
+      Setter[String](assetItem.source),
+      Setter[String](assetItem.sourceId)
+    )
+}
+
+class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
     extends CdpRelation[AssetsItem](config, "assets")
     with InsertableRelation
     with CdpConnector {
   @transient lazy private val assetsCreated =
     MetricsSource.getOrCreateCounter(config.metricsPrefix, s"assets.created")
+
+  override def update(rows: Seq[Row]): IO[Unit] = {
+    val updateAssetsItem = rows.map(r => UpdateAssetsItem(fromRow[UpdateAssetsItemBase](r)))
+
+    post(
+      config.auth,
+      uri"${baseAssetsURL(config.project)}/update",
+      updateAssetsItem,
+      config.maxRetries)
+  }
 
   override def insert(df: org.apache.spark.sql.DataFrame, overwrite: scala.Boolean): scala.Unit =
     df.foreachPartition((rows: Iterator[Row]) => {
--- a/src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
+++ b/src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
@@ -18,6 +18,7 @@ object Setter {
       case _ => Some(new Setter(set.get, false))
     }
 }
+case class NonNullableSetter[A](set: A)
 
 abstract class CdpRelation[T: DerivedDecoder: TypeTag: ClassTag](
     config: RelationConfig,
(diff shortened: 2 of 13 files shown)


Tapped:
`val column = q"new ${typeOf[Try[_]]}(Option($r.getAs[$rowType]($name))).toOption.flatten`
Ah yes i forgot that you'll need the new keyword. It will call the apply method if appropriate.

Tapped:
I don't remember the trick now, but what will work is `val column = q"scala.util.Try(Option($r.getAs[$rowType]($name))).toOption.flatten"`

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
```diff
@@ -18,6 +18,14 @@ object Setter {
       case _ => Some(new Setter(set.get, false))
     }
 }
+case class NonNullableSetter[A](set: A)
+object NonNullableSetter {
+  def apply[A](set: Option[A]): Option[NonNullableSetter[A]] =
+    set match {
```
wjoel: I think this is `set.map(NonNullableSetter(_))`? In which case we may not need the object and the apply method, seems like a small benefit compared to just doing `set.map(NonNullableSetter(_))` instead of `NonNullableSetter(map)`.

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -18,12 +18,12 @@ case class TimeSeriesConflict(notFound: Seq[Long])
 case class TimeSeriesNotFound(notFound: Seq[String])
 
 case class TimeSeriesItem(
-    name: String,
-    isString: Boolean,
+    name: Option[String],
+    isString: Option[Boolean],
```
wjoel: Why are we changing this? Makes sense for PostTimeSeriesItem but I don't see how it makes sense here...?

src/main/scala/com/cognite/spark/datasource/DefaultSource.scala
```diff
@@ -191,6 +191,9 @@ class DefaultSource
         new EventsRelation(config)(sqlContext) //.EventsInsertion(config)
       case "timeseries" =>
         new TimeSeriesRelation(config)(sqlContext)
+      case "assets" =>
+        val assetsPath = parameters.get("assetsPath")
```
wjoel: Is this parameter documented?

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
hakontro: SparkSchemaHelper interprets all non-Option fields as required when getting the schema in fromRow. I suppose we could find another way to get UpdateTimeSeriesItem from a row, any suggestions?

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
hakontro: Good catch, was caught up in the thinking from Setter here.

src/main/scala/com/cognite/spark/datasource/DefaultSource.scala
hakontro: Looks like an older parameter that's no longer in use, let's remove this from AssetsRelation.

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
wjoel: This is still relevant it seems. :)

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
wjoel: Could we make it a separate case class and use fromRow with whatever schema we get from that?

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
hakontro: Good catch again!

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
hakontro: Definitely

src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
```diff
@@ -39,13 +39,53 @@ case class PostAssetsItem(
     metadata: Option[Map[String, String]]
 )
 
-class AssetsRelation(config: RelationConfig, assetPath: Option[String])(val sqlContext: SQLContext)
+case class UpdateAssetsItemBase(
```
wjoel: Add a comment explaining why we need this. We will forget it otherwise. ;)

src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
```diff
@@ -39,13 +39,53 @@ case class PostAssetsItem(
...
+
+class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
     extends CdpRelation[AssetsItem](config, "assets")
     with InsertableRelation
     with CdpConnector {
   @transient lazy private val assetsCreated =
     MetricsSource.getOrCreateCounter(config.metricsPrefix, s"assets.created")
 
+  override def update(rows: Seq[Row]): IO[Unit] = {
+    val assetsItems = rows.map(r => fromRow[UpdateAssetsItemBase](r))
```
wjoel: I think it's readable enough in one line, `val updateAssetsItems = rows.map(r => UpdateAssetsItem(fromRow[UpdateAssetsItemBase](r)))`.

Note "items" in the val name, since we have many. I assume there's a limit on how many we can update in one go, so this should probably be grouped and updated in batches with a `parTraverse`?

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
```diff
@@ -292,41 +307,51 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
...
          |where source = '$source'
       """.stripMargin)
-      .select(destinationDf.columns.map(col): _*)
       .write
+      .format("com.cognite.spark.datasource")
+      .option("apiKey", writeApiKey.apiKey)
+      .option("type", "events")
       .option("onconflict", "update")
-      .insertInto("destinationEvent")
+      .save
```
wjoel: `save()` on those imo

src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
hakontro: Agreed, I'll change this elsewhere as well.
This is actually already being grouped in DefaultSource, perhaps we could find a way to do this for  insertInto writing as well.

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
hakontro: Done

PR #230, opened 2019-04-25
Fix tests failing because of eventual consistency

hakontro:
This fixes the failing test from FilesRelation, and uses collect in AssetsRelationTest for consistency. We should also find a way to fix TimeSeriesRelationTest, which has two main problems:

- should successfully both update and insert time series
This test times out sometimes
- should support abort in savemode
This sometimes fails because it doesn't always get a conflict on inserting existing rows due to EC.

Any other tests we know of that fail? The `does not throw exception` should be fixed everywhere we use the pattern, and since we'll add more of it once we implement savemode for other resource types it will be an even bigger problem soon.


wjoel:
This is currently my favorite PR. :D

PR #233, opened 2019-04-30
Implement read and write for the asset types field

hakontro:


src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
```diff
@@ -11,11 +11,55 @@ import org.apache.spark.sql.sources.InsertableRelation
...
+      for {
+        id <- c.downField("id").as[Long]
+        name <- c.downField("name").as[String]
+        valueType <- c.downField("valueType").as[String]
+        value <- valueType match {
+          case "String" => c.downField("value").as[String]
+          case "Long" => c.downField("value").as[Long].map(_.toString)
+          case "Double" => c.downField("value").as[Double].map(_.toString)
+          case "Boolean" => c.downField("value").as[Boolean].map(_.toString)
+        }
```
wjoel: I wonder what the error message looks like if it doesn't match one of the cases. Maybe we should add a default case to get an informative error?

It's possible that more value types might be added in the future. The perfect scenario would be if there was some generic "toString" method in circe.

src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
```diff
@@ -11,11 +11,55 @@ import org.apache.spark.sql.sources.InsertableRelation
...
+      } yield {
```
wjoel: I'm personally not a fan of curly brackets here, I prefer `yield Field(...)` and I think that's what we usually do, but we may not be consistent about this at the moment. I don't have any strong feelings about it.

src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
```diff
@@ -11,11 +11,55 @@ import org.apache.spark.sql.sources.InsertableRelation
...
+
+  implicit val fieldEncoder: Encoder[Field] = new Encoder[Field] {
+    override def apply(field: Field): Json = {
+      val value: Json = field.valueType match {
+        case "String" => Json.fromString(field.value)
+        case "Long" => Json.fromLong(field.value.toLong)
+        case "Double" =>
+          Json.fromDouble(field.value.toDouble).getOrElse(throw DoubleIsTooLargeForJSON(field.name))
+        case "Boolean" => Json.fromBoolean(field.value.toBoolean)
+      }
```
wjoel: Add a default case with a good error message.

src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
hakontro: Agreed, fits better with the no-redundant-parenthesis rule from scalafmt.

PR #236, opened 2019-05-13
Update Readme with asset types and partial updates

hakontro:
Also provide better error message when referencing non-existing asset
type field ids.


README.md
```diff
@@ -114,13 +115,39 @@ an API-key and the resource type. To read from a table you should also specify t
 
 ### Writing
 
-To write to a resource you'll need to register a DataFrame that was read from
+There are two ways you can use cdp-spark-datasource to write to CDF: using `insertInto` or using the `save` function.
```
wjoel: Should probably still use "CDP" here and rename it to "CDF" everywhere in a separate commit.

README.md
```diff
@@ -114,13 +115,39 @@ an API-key and the resource type. To read from a table you should also specify t
...
+#### Writing with `.insertInto()`
+
+To write to a resource using the insert into pattern you'll need to register a DataFrame that was read from
 that resource as a temporary view. You'll need a project where you have write access
 and replace `myApiKey` in the examples below.
 
 Your schema will have to match that of the target exactly. A convenient way to ensure
 this is to copy the schema from the DataFrame you read into with `sourceDf.select(destinationDf.columns.map(col):_*)`, see time series example.
 
+`.insertInto()` will do upsert, as in updating existing rows and inserting new rows, for events, assets and time series.
```
wjoel: Typically on "id" for upsert, I guess?

README.md
```diff
@@ -114,13 +115,39 @@ an API-key and the resource type. To read from a table you should also specify t
...
+
+#### Writing with `.save()`
+
+Writing with `.save()` is currently supported for assets, events and time series.
+
+You'll need to provide an API-key and the event type you'd like to write to. In addition you can specify
+the desired behaviour when rows in your Dataframe are present in CDF with the `.option("onconflict", value)`.
+
+The valid options for onconflict are
+- `abort` - will try to insert all rows in the Dataframe. An error will be thrown if the resource item already exists and no rows will be written.
```
wjoel: "no more items will be written" since it's possible to write valid items before a write fails due to conflict.

README.md
```diff
@@ -114,13 +115,39 @@ an API-key and the resource type. To read from a table you should also specify t
...
+- `update` - will look for all rows in the Dataframe in CDF and try to update them. If one or more rows are missing, none will be updated.
```
wjoel: I suspect some might be updated as well...? Does it throw an error if something isn't found? Due to batching some previous batch might have succeeded already, I'd guess?

README.md
```diff
@@ -136,11 +163,30 @@ val df = spark.sqlContext.read.format("com.cognite.spark.datasource")
...
-val someAsset = Seq(("99-BB-99999",99L,"This is another asset",Map("sourceSystem"->"MySparkJob"),99L))
-val someAssetDf = someAsset.toDF("name", "parentID", "description","metadata","id")
-someAssetDf
- .write
- .insertInto("assets")
+// Note that parentId, asset type IDs and asset type field IDs have to exist
+val someAsset = Seq(
+(
+99L, Seq(0L), 99L, "99-BB-99999", 2231996316030451L, "This is another asset",
+Seq((100L, // asset type ID
```
wjoel: While it's very useful to have an example with an asset type, I think we should keep the simple case. The example this replaces is very easy to understand, while this new one is quite intimidating and most people don't need to know about asset types.

README.md
```diff
@@ -221,17 +267,33 @@ val df = spark.read.format("com.cognite.spark.datasource")
...
   .load()
 myProjectDf.createTempView("events")
 
-// Copy the Valhall events to your project
-df.filter($"subtype" === "Valhall")
-  .write
-  .insertInto("events")
+// Update the description of all events from Open Industrial Data
+spark.sql("""
+ |select 'Manually copied data from publicdata' as description,
```
wjoel: GitHub highlights this strangely, might be a bug on their side. Just make sure it renders ok before merging.

README.md
hakontro: Done!

README.md
hakontro: Yep, and source+sourceId for events. Coming to think of it assets actually don't have upserts on insertInto.

README.md
hakontro: Ah, right. My understanding was a bit off here, but that's obviously how it'll behave. Would have liked it to be atomic though.. We _could_ implement this by listing the resource and comparing locally first, but that would be way slower. What do you think?

README.md
hakontro: Yep, fixed.

README.md
hakontro: Yeah, noticed this as well. It's their syntax highlighting that's off, it seems. I guess ```scala doesn't like triple quoted strings or something.

README.md
wjoel: Yeah, way slower. I don't think it's worth the performance impact, and it's pretty standard for those things to fail after some unknown number of items have been written when using other data sources in Spark, so it's not unexpected for someone who is used to Spark.

README.md
wjoel: Well, now the one with an asset type is gone, I think we can have both! ;)
First the simple case, then a complex one with an asset type.

PR #239, opened 2019-05-16
Improve robustness of TimeSeriesUpsertTest

wjoel:


src/test/scala/com/cognite/spark/datasource/TimeSeriesUpsertTest.scala
```diff
@@ -269,16 +266,29 @@ class TimeSeriesUpsertTest extends FlatSpec with Matchers with SparkTest {
     )
     assert(dfWithDescriptionUpsertTest.length == 10)
 
+    cleanupTestData(saveModeUnit)
   }
+
+  def cleanupTestData(testName: String): Unit = {
+    val namesDf =
+      spark.sql(s"""select name from sourceTimeSeries where unit = '$testName'""")
+    val namesList = namesDf.rdd.map(r => r.getAs[String](0)).collect().toList
```
hakontro: Looks like toList could be avoided if we make cleanupTimeSeries take an Array[String].

Issue #189, opened 2019-03-18
Set X-CDP-Client header


wjoel:
Investigate if we can/should set this header in preparation for API 1.0

wjoel:
This is now supported. See https://github.com/cognitedata/api-auth/pull/98

PR #242, opened 2019-05-27
Set x-cdp-sdk header for all API-calls

hakontro:


wjoel:
See https://github.com/cognitedata/api-auth/pull/98 for some more headers that we should set (linked from https://github.com/cognitedata/cdp-spark-datasource/issues/189). Arguably `app` would be the more appropriate one for identifying Spark, but I'm not sure.

`clienttag` should be set to the Spark job/group id, if available.

build.sbt
```diff
@@ -60,8 +60,10 @@ lazy val commonSettings = Seq(
   fork in Test := true
 )
 
+managedSourceDirectories in Compile += target.value / "protobuf-generated"
```
wjoel: Why?

src/main/scala/com/cognite/spark/datasource/CdpConnector.scala
```diff
@@ -32,16 +32,22 @@ trait CdpConnector {
...
-    val jsonResult = getJson[Data[Login]](auth, loginStatusUrl, maxRetries)
+    val jsonResult = getJson[Data[Login]](auth, loginStatusUrl, maxRetries, None)
     jsonResult.unsafeRunSync().data.project
   }
 
   def baseUrl(project: String, version: String, baseUrl: String): Uri =
     uri"$baseUrl/api/$version/projects/$project"
 
-  def getJson[A: Decoder](auth: Auth, url: Uri, maxRetries: Int): IO[A] = {
+  def getJson[A: Decoder](
```
wjoel: I think it's time to refactor this to take a `config` argument, since most arguments are coming from there already, and then we wouldn't need to add `applicationId` as yet another argument.

src/main/scala/com/cognite/spark/datasource/CdpRdd.scala
```diff
@@ -17,6 +17,9 @@ case class CdpRdd[A: Decoder](
     extends RDD[Row](sparkContext, Nil)
     with CdpConnector {
 
+  // This needs to be broadcast here since compute will be called from the executors
+  // where the SparkContext is not available
+  val broadcastedApplicationId = sparkContext.broadcast(Some(sparkContext.applicationId))
```
wjoel: Nice!

src/main/scala/com/cognite/spark/datasource/Constants.scala
```diff
@@ -15,4 +17,5 @@ object Constants {
   val DefaultBaseUrl = "https://api.cognitedata.com"
   val MetadataValuePostMaxLength = 512
   val MaxConcurrentRequests = 1000
+  val SparkDatasourceVersion = BuildInfo.organization + "-" + BuildInfo.version
```
wjoel: Use string interpolation instead.

src/main/scala/com/cognite/spark/datasource/CdpConnector.scala
wjoel: Then we could also make it a non-optional field and set it only once (when `config` is initialized, possibly using a default value for `applicationId`) instead of having to do `applicationId.getOrElse` in several places.

Unless I'm missing something which is quite possible. :)

src/main/scala/com/cognite/spark/datasource/CdpConnector.scala
```diff
@@ -32,16 +32,22 @@ trait CdpConnector {
...
+      auth: Auth,
+      url: Uri,
+      maxRetries: Int,
+      applicationId: Option[String]): IO[A] = {
     val result = sttp
       .header("Accept", "application/json")
+      .header("x-cdp-sdk", Constants.SparkDatasourceVersion)
+      .header("x-cdp-app", applicationId.getOrElse("unknown application"))
```
wjoel: In other places we use `"Scala"` but we should define a default value in `Constants` and use that. `"Spark"` is probably a better choice than `"Scala"` since our users can be calling this from PySpark.

src/main/scala/com/cognite/spark/datasource/test.sc
```diff
@@ -1,2 +0,0 @@
-val a = 10
-a + 10
```
wjoel: Weird. Where did this come from?

src/test/scala/com/cognite/spark/datasource/AssetsRelationTest.scala
```diff
@@ -264,8 +274,7 @@ class AssetsRelationTest extends FlatSpec with Matchers with SparkTest {
       spark.createDataset(assetTypesJsonString :: Nil)
     val assetTypesDataFrame = spark.read.json(assetTypesDataSet)
 
-    assetTypesDataFrame
-      .write
+    assetTypesDataFrame.write
```
wjoel: Based on the other formatting changes it seems like this *should* be
```
assetTypesDataFrame
  .write
```

src/test/scala/com/cognite/spark/datasource/CdpRddTest.scala
```diff
@@ -61,85 +73,109 @@ class CdpRddTest extends FlatSpec with Matchers with SparkTest {
...
         url.paramsMap.get("limit") should be(Some("2"))
         url.paramsMap.get("cursor") should be(expectedCursors.next)
         IO.pure(decode(numberedItems.next).right.get)
       }
     }
 
-    rdd.partitions should have length 5
-    rdd.partitions.map(p => p.asInstanceOf[CdpRddPartition].cursor) should
-      contain only (Seq(None) ++ (0 to 3).map(x => Some(x.toString)): _*)
+    (rdd.partitions should have).length(5)
```
wjoel: While I'm not the biggest fan, ScalaTest does have a DSL and we (as well as most others) use the DSLish way of writing our tests. Don't change it in this PR. If we want to change it we should do so in a separate PR and we should change all tests at once, but I don't think we should tbh.

build.sbt
hakontro: https://github.com/sbt/sbt/issues/1583#issuecomment-262166588
This fixed a race condition between ProtoBuf and BuildInfo, I'll add a comment so we remember.

src/main/scala/com/cognite/spark/datasource/test.sc
hakontro: Looks like it snuck in here https://github.com/cognitedata/cdp-spark-datasource/commit/740e3672ac9f473048f00ef638520aa40f87b304#diff-fe8c54b29b54518d08cd9ea7e35a63b6

hakontro:
Now ignoring two tests in Events that make use of the events/update endpoint which is currently broken for 0.6

PR #252, opened 2019-06-14
Check null on update asset

ebrattli:


src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
```diff
@@ -96,10 +96,18 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
 
   override def update(rows: Seq[Row]): IO[Unit] = {
-    val assetsItems = rows.map(r => fromRow[UpdateAssetsItemBase](r))
-    val updateAssetsItem = assetsItems.map(a => UpdateAssetsItem(a))
-
-    post(config, uri"${baseAssetsURL(config.project)}/update", updateAssetsItem)
+    val updateAssetsItems = rows
+      .map(r => fromRow[UpdateAssetsItemBase](r))
+      .map(a => UpdateAssetsItem(a))
+    updateAssetsItems
```
wjoel: Use [exists](https://www.scala-lang.org/api/2.13.x/scala/collection/Seq.html#exists(p:A=%3EBoolean):Boolean) and say `if (updateAssetsItems.exists(_.id.isEmpty))` to decide when to throw the exception.

src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
```diff
@@ -96,10 +96,18 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+      .filter(updateAssetItem =>
+        updateAssetItem.id match {
+          case Some(id) => false
+          case None => true
+      })
+      .foreach(a => throw new IllegalArgumentException("Got asset with null id."))
```
wjoel: "Assets must have an id when using update" or similar so the user knows more about what is happening.

src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
```diff
@@ -96,10 +98,16 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
   override def update(rows: Seq[Row]): IO[Unit] = {
-    val assetsItems = rows.map(r => fromRow[UpdateAssetsItemBase](r))
-    val updateAssetsItem = assetsItems.map(a => UpdateAssetsItem(a))
+    val updateAssetsItems = rows
+      .map(r => fromRow[UpdateAssetsItemBase](r))
+      .map(a => UpdateAssetsItem(a))
+
+    // Assets must have an id when using update
+    if (updateAssetsItems.exists(_.id.isEmpty)) {
+      throw new IllegalArgumentException("Got asset with null id.")
```
wjoel: "Assets must have an id when using update" is a better message. ;)

wjoel:
Seems we've rebased to the wrong commit message though, ""Metals requires that to work properly". Feel free to merge once that's changed! :)

Issue #253, opened 2019-06-16
Data points doesn't use config.baseUrl


wjoel:
`.option("baseUrl", ...)` doesn't work when using the `datapoints` type.

To reproduce, try creating a destination data frame for data points with a custom baseUrl, and try writing to it.

```python
src_dp = spark \
  .read \
  .format("com.cognite.spark.datasource") \
  .option("apikey", dbutils.secrets.get(secret_scope, src)) \
  .option("type", "datapoints") \
  .load()

dst_dp = spark \
  .read \
  .format("com.cognite.spark.datasource") \
  .option("apikey", dbutils.secrets.get(secret_scope, dst)) \
  .option("baseUrl", "https://greenfield.cognitedata.com") \
  .option("type", "datapoints") \
  .load()
dst_dp.createOrReplaceTempView("dst_dp")

src_dp \
    .where(src_dp.name.isin(tsmd_names)) \
    .where(src_dp.timestamp > 1546300800000) \
    .write \
    .insertInto("dst_dp")
```

The driver contains errors like
`https://api.cognitedata.com/api/0.5/projects/wjoel/timeseries/data failed with status 401: Unauthorized`

It should be using greenfield.cognitedata.com instead (`wjoel` is the name of the destination project).

PR #254, opened 2019-06-17
Use the baseUrl option in DataPointsRelation

wjoel:
Fixes https://github.com/cognitedata/cdp-spark-datasource/issues/253


wjoel:
We should try to come up with a good way to test this. Maybe a corresponding test project in Greenfield with different data?

PR #256, opened 2019-06-18
Implement generalized push-down filters

hakontro:


src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
```diff
@@ -76,6 +76,20 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
...
+    val df = spark.read
+      .format("com.cognite.spark.datasource")
+      .option("apiKey", writeApiKey.apiKey)
+      .option("type", "events")
+      .load()
+      .where(s"((type = 'maintenance' or type = 'upgrade') " +
+        s"and subtype in('manual', 'automatic')) " +
+        s"or (type = 'maintenance' and subtype = 'manual') " +
+        s"or (type = 'upgrade') and source = 'something'")
+    assert(df.count == 4)
```
Tapped: How would you know how many items that were downloaded to fulfill the query? In other words are we now actually checking if it filters the source system (CDF) itself.

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
```diff
@@ -22,31 +33,66 @@ case class DeleteItem(
...
     )
+  }
 
   def toRow(t: T): Row
 
+  def toRow[A <: Product](item: A, requiredColumns: Array[String]): Row = {
+    val values = item.productIterator
+    val itemMap = item.getClass.getDeclaredFields.map(_.getName -> values.next).toMap
+
+    Row.fromSeq(requiredColumns.map(itemMap(_)).toSeq)
```
Tapped: This function doesn't work for higher kinded types (Option, Scala Seq etc) I believe?

Tapped: Which means that requiredColumns has to point to primitives in a case class / class.

src/main/scala/com/cognite/spark/datasource/CdpRdd.scala
```diff
@@ -28,18 +34,28 @@ case class CdpRdd[A: Decoder](
...
+    val rowIteratorsSeq = urlsWithFilters
+      .map { url =>
+        get[A](
+          config.copy(limit = split.size, applicationId = broadcastApplicationId.value),
+          url,
+          split.cursor
+        )
+      }
+
+    val rowIteratorsSet = rowIteratorsSeq.foldLeft(Set[A]())(_ ++ _)
```
Tapped: You can use `foldK` from Cats here instead (`import cats.implicits._`)

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
```diff
@@ -22,31 +33,66 @@ case class DeleteItem(
...
+      .distinct
+
     CdpRdd[T](
       sqlContext.sparkContext,
       (e: T) => {
         if (config.collectMetrics) {
           itemsRead.inc()
         }
-        toRow(e)
+        if (requiredColumns.isEmpty) toRow(e) else toRow(e, requiredColumns)
```
wjoel: I think we can put this logic into `toRow(e, requiredColumns)` and handle requiredColumns being empty in toRow.

src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
```diff
@@ -102,6 +102,8 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
     post(config, uri"${baseAssetsURL(config.project)}/update", updateAssetsItem)
   }
 
+  override def fieldsWithPushdownFilter: Seq[String] = Seq("name", "source", "depth", "description")
```
wjoel: Do we need this to be a def? Seems a constant `val` would suffice.

src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
```diff
@@ -144,12 +146,8 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
 
   override def toRow(t: AssetsItem): Row = asRow(t)
 
-  // TODO: As of 2019-03-22 the normal /assets endpoint doesn't work on some projects when
-  // using cursors retrieved from /assets/cursors, but the cursors do work with /assets/list.
-  // This is ok as long as we don't support predicate pushdown for assets, as we don't need
-  // the search capabilities of /assets, but this should be removed once /assets works as expected.
   override def listUrl(): Uri =
-    uri"${config.baseUrl}/api/0.6/projects/${config.project}/assets/list"
+    uri"${config.baseUrl}/api/0.6/projects/${config.project}/assets"
```
wjoel: 👍

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
hakontro: You're right, thanks.

src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
hakontro: Good catch.

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
hakontro: Seems to work fine, requiredColumns is usually the name of all columns and all tests are still passing.

Tapped: Hmm, last time I checked it didn't create a row as I expected. However if that's the case, we can remove the macro code for creating a row, though the macro code will probably be faster since it unrolls a dynamic loop at compile time.

Tapped: And given that toRow will potentially be called for billions of rows, that type of performance improvement actually matters. However, performance measuring is something we should add to our todo list.

hakontro: I agree. `SparkSchemaHelperImpl.asRow` is still called when scanning without pushdown filters, so let's keep it.

src/main/scala/com/cognite/spark/datasource/CdpRdd.scala
wjoel: `Set.empty[A]` is better I think.

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
```diff
@@ -22,31 +33,67 @@ case class DeleteItem(
...
       config,
+      pushdownFilters,
       cursors()
     )
+  }
 
   def toRow(t: T): Row
 
+  def toRow(item: T, requiredColumns: Array[String]): Row =
+    if (requiredColumns.isEmpty) { toRow(item) } else {
```
wjoel:
```
if (...) {
  ...
} else {
  ...
}```

File: src/main/scala/com/cognite/spark/datasource/EventsRelation.scala [hunk 8]
```diff
@@ -180,52 +178,7 @@ class EventsRelation(config: RelationConfig)(@transient val sqlContext: SQLConte
...
-    } yield EventFilter(eventType, eventSubType)
-
-    EventsRdd(sqlContext.sparkContext, listUrl(), (e: EventItem) => {
-      if (config.collectMetrics) {
-        eventsRead.inc()
-      }
-      toRow(e, requiredColumns)
-    }, eventFilters, config, cursors())
-  }
+  override val fieldsWithPushdownFilter: Seq[String] = Seq("source", "type", "assetIds")
```
wjoel: `assetIds` is new, we should have some tests for that.

Could we also add push-down filtering for the timestamp columns while we're at this? Updated, created, start time, end time. Would be very useful.

Also id, if that makes sense (may be more difficult?)

hakontro: Push-down filtering for timestamp should be fine, but will require some custom work for Events so let's put that in another PR.

I'm not sure about ID, for 0.6 we would have to use the /events/{eventId} endpoint. We should probably just let ID take precedence over other pushdown filters since there'll be just a few results anyways.

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
wjoel: Could perhaps check the metrics for this?

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
hakontro: Looks like `assetIds` will need special treatment as well, I'll remove it from here and add an issue for it.

wjoel: `subType` however should be easy, right?

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
hakontro: Good tip! Works perfectly.

src/main/scala/com/cognite/spark/datasource/CdpRdd.scala
hakontro: Agreed, done.

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
hakontro: `subType` will throw an error unless the API call has a filter on `type` as well, which doesn't fit too easily with the current implementation (one API call per pushdown filter).

PR #257, opened 2019-06-21
Improve error msg on invalid onConflict option

ebrattli:


src/test/scala/com/cognite/spark/datasource/AssetsRelationTest.scala
```diff
@@ -148,6 +149,31 @@ class AssetsRelationTest extends FlatSpec with Matchers with SparkTest {
       .save
   }
 
+  it should "Should give good error on invalid onconflict options" taggedAs WriteTest in {
```
Tapped: remove `Should` from the string literal

src/main/scala/com/cognite/spark/datasource/DefaultSource.scala
```diff
@@ -24,6 +24,8 @@ case class RelationConfig(
 object OnConflict extends Enumeration {
   type Mode = Value
   val ABORT, UPDATE, UPSERT, DELETE = Value
+
+  def withNameOpt(s: String): Option[Value] = values.find(_.toString.toLowerCase == s.toLowerCase())
```
Tapped: Can't we use
https://www.scala-lang.org/api/current/scala/Enumeration.html#withName(s:String):Enumeration.this.Value?

Or is that method not case insensitive?

wjoel:
Small detail, but it adds up over time: please capitalize the first word in the commit message. We generally try to follow the guidelines in https://chris.beams.io/posts/git-commit/

src/main/scala/com/cognite/spark/datasource/DefaultSource.scala
ebrattli: The problem with that method is that it throws the exception referenced in the related issue. The new function returns an Option and allows us to throw a more specific error message.

Tapped: Oh, I see then it makes sense. And it's also nicer than relying on exception for code flow, as when doing `Try(OnConflict.withName(parameters.getOrElse("onconflict", "ABORT").toUpperCase())).toOption`.

PR #258, opened 2019-06-21
Check for null id on update

ebrattli:
Make sure we don't send values with null ids to the API on update.
Instead check and throw an IllegalArgumentException if we receive
any null ids on update. Previously we only checked this on assets, but now we also check on timeSeries and events.

### DIFF
--- a/src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
+++ b/src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
@@ -81,6 +81,7 @@ case class UpdateTimeSeriesItem(
     isString: Option[Setter[Boolean]],
     isStep: Option[Setter[Boolean]]
 )
+
 object UpdateTimeSeriesItem {
   def apply(timeSeriesItem: TimeSeriesItem): UpdateTimeSeriesItem =
     new UpdateTimeSeriesItem(
@@ -129,8 +130,15 @@ class TimeSeriesRelation(config: RelationConfig)(val sqlContext: SQLContext)
   }
 
   override def update(rows: Seq[Row]): IO[Unit] = {
-    val updateTimeSeriesItems =
-      rows.map(r => UpdateTimeSeriesItem(fromRow[UpdateTimeSeriesBase](r)))
+    val updateTimeSeriesBaseItems =
+      rows.map(r => fromRow[UpdateTimeSeriesBase](r))
+
+    // Time series must have an id when using update
+    if (updateTimeSeriesBaseItems.exists(_.id.isEmpty)) {
+      throw new IllegalArgumentException("Time series must have an id when using update")
+    }
+
+    val updateTimeSeriesItems = updateTimeSeriesBaseItems.map(UpdateTimeSeriesItem(_))
 
     post(
       config,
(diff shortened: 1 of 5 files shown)


src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
```diff
@@ -134,7 +134,7 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
     assert(storedMetadata.get("foo").contains("bar"))
   }
 
-  it should "support upserts" taggedAs WriteTest ignore {
+  (it should "support upserts" taggedAs WriteTest).ignore {
```
wjoel: Don't add parenthesis here

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
```diff
@@ -275,7 +275,7 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
     spark.sparkContext.setLogLevel("WARN")
   }
 
-  it should "allow partial updates in savemode" taggedAs WriteTest ignore {
+  (it should "allow partial updates in savemode" taggedAs WriteTest).ignore {
```
wjoel: Don't add parenthesis here

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -129,8 +130,15 @@ class TimeSeriesRelation(config: RelationConfig)(val sqlContext: SQLContext)
   }
 
   override def update(rows: Seq[Row]): IO[Unit] = {
-    val updateTimeSeriesItems =
-      rows.map(r => UpdateTimeSeriesItem(fromRow[UpdateTimeSeriesBase](r)))
+    val updateTimeSeriesItemsBase =
```
wjoel: Guess it should be `updateTimeSeriesBaseItems` to be consistent with the others.

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -129,8 +130,15 @@ class TimeSeriesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+      rows.map(r => fromRow[UpdateTimeSeriesBase](r))
+
+    // TimeSeries must have an id when using update
+    if (updateTimeSeriesItemsBase.exists(_.id.isEmpty)) {
+      throw new IllegalArgumentException("TimeSeries must have an id when using update")
```
wjoel: "Time series" in text.

src/test/scala/com/cognite/spark/datasource/TimeSeriesRelationTest.scala
```diff
@@ -268,6 +284,55 @@ class TimeSeriesRelationTest extends FlatSpec with Matchers with SparkTest {
...
 
+  it should "check for null ids on time series update" taggedAs WriteTest in {
+    val df = spark.read
+      .format("com.cognite.spark.datasource")
+      .option("apiKey", writeApiKey.apiKey)
+      .option("type", "timeseries")
+      .load()
+    df.createOrReplaceTempView("sourceTimeSeries")
+
+    val initialDescription = "'post testing'"
```
hakontro: Looks like we haven't been consistent in this file wrt. whether or not to add the single quotes in the value definition, or add it where the value is referenced. We've been consistent in the other test files though, so I suggest removing the single quotes from the value definition here and adding them to the string where they're referenced.

PR #259, opened 2019-06-21
Use inferschema limit

Tapped:
Inferschema limit was before ignored, now it's enabled again.

### DIFF
--- a/src/main/scala/com/cognite/spark/datasource/RawTableRelation.scala
+++ b/src/main/scala/com/cognite/spark/datasource/RawTableRelation.scala
@@ -57,7 +57,10 @@ class RawTableRelation(
 
   override val schema: StructType = userSchema.getOrElse {
     if (inferSchema) {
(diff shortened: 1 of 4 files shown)


src/main/scala/com/cognite/spark/datasource/RawTableRelation.scala
```diff
@@ -57,7 +57,10 @@ class RawTableRelation(
 
   override val schema: StructType = userSchema.getOrElse {
     if (inferSchema) {
-      val rdd = readRows(inferSchemaLimit, collectSchemaInferenceMetrics)
+      val rdd =
+        readRows(
+          inferSchemaLimit.orElse(Some(Constants.DefaultRawBatchSize)),
```
wjoel: Better make a new constant for this (but this value, 10k, seems fine).

Tapped: Maybe, I thought batch size always would be a good metric for infer schema limit, maybe even use the user provided batch size for this. What do you think about that?

PR #265, opened 2019-06-26
Expose struct type schema

Tapped:
This feature can be used by Jetfire to retrieve the right schema based on save mode, when Jetfire gets support for it.

### DIFF
--- a/src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
+++ b/src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
@@ -18,6 +18,7 @@ import org.apache.spark.sql.sources.{
 import org.apache.spark.datasource.MetricsSource
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.{Row, SQLContext}
+import org.apache.spark.sql.types.{DataTypes, StructField, StructType}
 
 case class Setter[A](set: A, setNull: Boolean)
 object Setter {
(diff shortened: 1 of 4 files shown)


src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
```diff
@@ -124,5 +125,20 @@ abstract class CdpRelation[T <: Product: Decoder](config: RelationConfig, shortN
   def delete(rows: Seq[Row]): IO[Unit] =
     throw new IllegalArgumentException(
       s"""$shortName does not support the "onconflict" option "delete".""")
+}
+
+trait InsertSchema {
```
wjoel: "Has" or "With" maybe?

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
```diff
@@ -124,5 +125,20 @@ abstract class CdpRelation[T <: Product: Decoder](config: RelationConfig, shortN
...
+
+trait UpsertSchema {
+  val upsertSchema: StructType
+}
+
+trait UpdateSchema {
+  val updateSchema: StructType
+}
 
+abstract class DeleteSchema {
```
wjoel: While tempting to use an abstract class to get a default value it's a bit annoying with the single inheritance.
I'd perhaps rather just define the value once in an object or something and do `val deleteSchema = DeleteSchema.withOnlyId` in the classes implementing it.

This is also ok though, we can always change it later if the inheritance thing becomes an issue.

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -66,9 +66,7 @@ case class UpdateTimeSeriesBase(
     isStep: Option[Boolean],
     description: Option[String],
     securityCategories: Option[Seq[Long]],
-    id: Option[Long],
-    createdTime: Option[Long],
-    lastUpdatedTime: Option[Long])
+    id: Option[Long])
```
wjoel: Oops. Good catch!

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
Tapped: Yeah, was thinking the same thing, and if we going to have a prefix I'm in favor of `Has`. It's a trait to have InsertSchema

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
Tapped: Good point, I think I'll keep it as this for now, since it can as you said always be changed later.

PR #266, opened 2019-06-26
Implement pushdown filters on Events

hakontro:
Implement pushdown filters for type, subtype, assetIds, startTime and
source. ID will have to be handled in another way, since ID-filters are
passed in the body rather than as query parameters.

### DIFF
--- a/src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
+++ b/src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
@@ -49,12 +49,7 @@ abstract class CdpRelation[T <: Product: Decoder](config: RelationConfig, shortN
   val sqlContext: SQLContext
   override def buildScan(): RDD[Row] = buildScan(Array.empty, Array.empty)
 
-  override def buildScan(requiredColumns: Array[String], filters: Array[Filter]): RDD[Row] = {
-    val pushdownFilters: Seq[PushdownFilter] = fieldsWithPushdownFilter
-      .map(f => (f, filters.flatMap(getFilter(_, f)))) // map from fieldname to its filtering values
-      .flatMap(pdf => pdf._2.map(v => PushdownFilter(pdf._1, v)))
-      .distinct
-
+  override def buildScan(requiredColumns: Array[String], filters: Array[Filter]): RDD[Row] =
     CdpRdd[T](
       sqlContext.sparkContext,
       (e: T) => {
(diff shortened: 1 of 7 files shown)


src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
```diff
@@ -65,9 +60,18 @@ abstract class CdpRelation[T <: Product: Decoder](config: RelationConfig, shortN
       },
       listUrl(),
       config,
-      pushdownFilters,
+      urlsWithFilters(filters, listUrl),
       cursors()
     )
+
+  def urlsWithFilters(filters: Array[Filter], uri: Uri): Seq[Uri] = {
+    val urlsWithFilters = fieldsWithPushdownFilter
```
wjoel: I think this might be better if written as
```
if (fieldsWithPushdownFilter.isEmpty) {
  Seq(uri)
} else {
  fieldsWithPushdownFilter.map(...).flatMap(...).distinct.map(...)
}

File: src/main/scala/com/cognite/spark/datasource/RawTableRelation.scala [hunk 2]
```diff
@@ -104,7 +104,7 @@ class RawTableRelation(
       },
       baseUrlWithParams,
       configWithLimit,
-      Seq[PushdownFilter](),
+      Seq[Uri](baseUrlWithParams),
```
wjoel: Do we need "Uri" here?

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -183,7 +184,73 @@ class EventsRelation(config: RelationConfig)(@transient val sqlContext: SQLConte
...
+    val filterMaps: Map[String, Array[String]] =
+      fieldsWithPushdownFilter.map(col => (col, filters.flatMap(getFilter(_, col)))).toMap
+    val assetIdsFilter = filterMaps.getOrElse("assetId", Array())
+    val typeFilter = filterMaps.getOrElse("type", Array())
+    val subtypeFilter = filterMaps.getOrElse("subtype", Array())
+    val sourceFilter = filterMaps.getOrElse("source", Array())
+
+    val startTimeFilters = getStartTimeFilters(filters)
+
+    val assetIds = if (assetIdsFilter.nonEmpty) {
```
wjoel: There must be a better way to do those `if (x.nonEmpty) { Seq(...) } else { Seq() }` things, I'm a bit suspicious of so many ifs and "toSeq".

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -183,7 +184,73 @@ class EventsRelation(config: RelationConfig)(@transient val sqlContext: SQLConte
...
+      typeAndSubtypes.toSeq.map { p =>
+        uri.param("type", p._1).param("subtype", p._2)
+      }
+    } else { Seq() }
+
+    val types = if (typeFilter.nonEmpty) {
+      typeFilter.map(p => uri.param("type", p)).toSeq
+    } else { Seq() }
+
+    val startTimes = if (startTimeFilters._1.isDefined || startTimeFilters._2.isDefined) {
```
wjoel: Add a test covering this logic.

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
hakontro: Definitely, thanks.

src/main/scala/com/cognite/spark/datasource/RawTableRelation.scala
hakontro: We do not! :)

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
hakontro: Agree, that was a bit rushed. Changed to Options + pattern matching.

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
hakontro: Done

PR #276, opened 2019-07-09
Add pushdown on event id

ebrattli:
Pushdown on event ids now uses the "events/byids" endpoint and makes a post request with the ids in the body. Currently writing more tests.

### DIFF
--- a/src/main/scala/com/cognite/spark/datasource/CdpConnector.scala
+++ b/src/main/scala/com/cognite/spark/datasource/CdpConnector.scala
@@ -66,6 +66,26 @@ trait CdpConnector {
     retryWithBackoff(result, Constants.DefaultInitialRetryDelay, config.maxRetries)
   }
 
+  def postJsonWithBody[A: Decoder, B: Encoder](
+      config: RelationConfig,
+      url: Uri,
+      items: Seq[B]): IO[A] = {
+    val result = sttp
+      .header("Accept", "application/json")
+      .header("x-cdp-sdk", Constants.SparkDatasourceVersion)
+      .header("x-cdp-app", config.applicationId)
+      .header("Content-Type", "application/json")
+      .body(Items(items))
+      .auth(config.auth)
+      .post(url)
+      .response(asJson[A])
+      .parseResponseIf(_ => true)
+      .send()
+      .flatMap(r => onError(url)(r))
+
+    retryWithBackoff(result, Constants.DefaultInitialRetryDelay, config.maxRetries)
+  }
+
   def getProtobuf[A](
       config: RelationConfig,
       url: Uri,
@@ -95,6 +115,16 @@ trait CdpConnector {
       initialCursor: Option[String] = None): Iterator[A] =
     getWithCursor(config, url, initialCursor)
       .flatMap(_.chunk)
+
+  def postWithBody[A: Decoder, B: Encoder](
+      config: RelationConfig,
+      url: Uri,
+      items: Seq[B]): Iterator[A] =
+    postJsonWithBody[DataItems[A], B](config, url, items)
+      .unsafeRunSync()
+      .data
+      .items
+      .toIterator
 
   def onError[A, B](url: Uri): PartialFunction[Response[Either[B, A]], IO[A]] = {
     case Response(Right(Right(data)), _, _, _, _) => IO.pure(data)
@@ -253,5 +283,6 @@ object CdpConnector {
     AsyncHttpClientCatsBackend[IO]()
 
   type DataItemsWithCursor[A] = Data[ItemsWithCursor[A]]
+  type DataItems[A] = Data[Items[A]]
   type CdpApiError = Error[CdpApiErrorPayload]
 }
--- a/src/main/scala/com/cognite/spark/datasource/CdpEventsRdd.scala
+++ b/src/main/scala/com/cognite/spark/datasource/CdpEventsRdd.scala
@@ -1,0 +1,42 @@
+package com.cognite.spark.datasource
+
+import io.circe.syntax._
+import io.circe.generic.auto._
+import org.apache.spark._
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.Row
+import com.softwaremill.sttp._
+
+case class CdpEventByIdsRddPartition(index: Int, uri: Uri, eventIds: Seq[EventId]) extends Partition
+
+case class CdpEventsByIdsRdd(
+    @transient override val sparkContext: SparkContext,
+    toRow: EventItem => Row,
+    config: RelationConfig,
+    uri: Uri,
+    eventIds: Seq[EventId])
+    extends RDD[Row](sparkContext, Nil)
+    with CdpConnector {
+  // This needs to be broadcast here since compute will be called from the executors
+  // where the SparkContext is not available
+  private val broadcastApplicationId = sparkContext.broadcast(sparkContext.applicationId)
+  override def getPartitions: Array[Partition] = {
+    val partitionedBodies = partitionUrlBodies(eventIds)
+    partitionedBodies.zipWithIndex.map {
+      case (ids, i) =>
+        CdpEventByIdsRddPartition(i, uri, ids)
+    }.toArray
+  }
+
+  override def compute(_split: Partition, context: TaskContext): Iterator[Row] = {
+    val split = _split.asInstanceOf[CdpEventByIdsRddPartition]
+    val rowIteratorsSeq =
+      postWithBody[EventItem, EventId](config, split.uri, split.eventIds)
+    val rowIterators = rowIteratorsSeq.toList.map(toRow).toIterator
+
+    new InterruptibleIterator(context, rowIterators)
+  }
+
+  private def partitionUrlBodies(eventUriBody: Seq[EventId]): Seq[Seq[EventId]] =
+    eventUriBody.grouped(Constants.DefaultBatchSize).toSeq
+}
--- a/src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
+++ b/src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
@@ -1,5 +1,6 @@
 package com.cognite.spark.datasource
 import cats.effect.IO
+import com.codahale.metrics.Counter
 import com.cognite.spark.datasource.SparkSchemaHelper._
 import com.softwaremill.sttp._
 import io.circe.Decoder
@@ -42,7 +43,7 @@ abstract class CdpRelation[T <: Product: Decoder](config: RelationConfig, shortN
     with Serializable
     with PrunedFilteredScan
     with CdpConnector {
-  @transient lazy private val itemsRead =
+  @transient lazy protected val itemsRead: Counter =
     MetricsSource.getOrCreateCounter(config.metricsPrefix, s"$shortName.read")
 
   val fieldsWithPushdownFilter: Seq[String] = Seq[String]()
@@ -50,7 +51,14 @@ abstract class CdpRelation[T <: Product: Decoder](config: RelationConfig, shortN
   val sqlContext: SQLContext
   override def buildScan(): RDD[Row] = buildScan(Array.empty, Array.empty)
 
-  override def buildScan(requiredColumns: Array[String], filters: Array[Filter]): RDD[Row] =
+  override def buildScan(requiredColumns: Array[String], filters: Array[Filter]): RDD[Row] = {
+    val urlsFilters = urlsWithFilters(filters, listUrl())
+    val urls = if (urlsFilters.isEmpty) {
+      Seq(listUrl())
+    } else {
+      urlsFilters
+    }
+
     CdpRdd[T](
       sqlContext.sparkContext,
       (e: T) => {
@@ -61,18 +69,17 @@ abstract class CdpRelation[T <: Product: Decoder](config: RelationConfig, shortN
       },
       listUrl(),
       config,
-      urlsWithFilters(filters, listUrl),
+      urls,
       cursors()
     )
+  }
 
-  def urlsWithFilters(filters: Array[Filter], uri: Uri): Seq[Uri] = {
-    val urlsWithFilters = fieldsWithPushdownFilter
+  def urlsWithFilters(filters: Array[Filter], uri: Uri): Seq[Uri] =
+    fieldsWithPushdownFilter
       .map(f => (f, filters.flatMap(getFilter(_, f)))) // map from fieldname to its filtering values
       .flatMap(pdf => pdf._2.map(v => PushdownFilter(pdf._1, v)))
       .distinct
       .map(f => uri.param(f.fieldName, f.value))
-    if (urlsWithFilters.isEmpty) { Seq(uri) } else urlsWithFilters
-  }
 
   def toRow(t: T): Row
 
--- a/src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
+++ b/src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
@@ -7,6 +7,7 @@ import com.softwaremill.sttp._
 import io.circe.generic.auto._
 import io.circe.parser.decode
 import org.apache.spark.datasource.MetricsSource
+import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types.{DataTypes, StructType}
 import org.apache.spark.sql.{DataFrame, Row, SQLContext}
@@ -81,6 +82,7 @@ object UpdateEventItem {
     )
 }
 
+case class EventId(id: Long)
 case class IdSourceAndResourceId(id: Long, source: String, sourceId: String)
 case class EventConflict(duplicates: Seq[IdSourceAndResourceId])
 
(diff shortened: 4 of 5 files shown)


src/main/scala/com/cognite/spark/datasource/CdpConnector.scala
```diff
@@ -96,6 +114,13 @@ trait CdpConnector {
     getWithCursor(config, url, initialCursor)
       .flatMap(_.chunk)
 
+  def postWithBody[A: Decoder](config: RelationConfig, url: Uri, body: String): Iterator[A] =
```
ebrattli: Should body be a string or should it be a generic type with an encoder?

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -184,7 +204,53 @@ class EventsRelation(config: RelationConfig)(@transient val sqlContext: SQLConte
...
 
-  override val fieldsWithPushdownFilter = Seq("type", "subtype", "assetIds", "startTime", "source")
+  override val fieldsWithPushdownFilter: Seq[String] =
+    Seq("type", "subtype", "assetIds", "startTime", "source")
+
+  override def buildScan(requiredColumns: Array[String], filters: Array[Filter]): RDD[Row] = {
+    val b = urlWithBody(filters, listUrl)
+    val u = urlsWithFilters(filters, listUrl)
+
+    if (b.body.nonEmpty && u.nonEmpty && u.head != listUrl) {
```
ebrattli: should probably allow for additional filters to be performed locally.

.scalafmt.conf
```diff
@@ -7,7 +7,6 @@ maxColumn = 100
 align = none
 
 rewrite.rules = [
-  AvoidInfix
```
wjoel: Why are we removing this?

src/main/scala/com/cognite/spark/datasource/CdpConnector.scala
wjoel: Isn't this `post` except with `DataItems[A]` instead of `Items[A]`? Which has `[A: Encoder]`.

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
```diff
@@ -42,7 +43,7 @@ abstract class CdpRelation[T <: Product: Decoder](config: RelationConfig, shortN
     with Serializable
     with PrunedFilteredScan
     with CdpConnector {
-  @transient lazy private val itemsRead =
+  @transient lazy val itemsRead: Counter =
```
wjoel: We don't need to expose this, see `getNumberOfRowsRead` in SparkTest for how to retrieve those metrics through the metric system.

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -81,6 +84,23 @@ object UpdateEventItem {
...
 }
 
+sealed trait RetrieveEvents
+case class EventId(id: Long) extends RetrieveEvents
+case class EventExternalId(externalId: String) extends RetrieveEvents
+
+object RetrieveEventCoders {
+  implicit val encodeRetrieveEvents: Encoder[RetrieveEvents] = new Encoder[RetrieveEvents] {
+    override def apply(a: RetrieveEvents): Json =
+      Json.obj(a match {
```
wjoel: I'm not sure that it's valid to mix the two types of retrieval in one request, but maybe...? Have we tested that?

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -81,6 +84,23 @@ object UpdateEventItem {
...
+        case EventId(id) => ("id", Json.fromLong(id))
+        case EventExternalId(id) => ("externalId", Json.fromString(id))
+      })
+  }
+
+  implicit val decodeRetrieveEvents: Decoder[RetrieveEvents] =
+    Decoder[EventId].widen or Decoder[EventExternalId].widen
```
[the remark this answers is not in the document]
wjoel: I wonder if it's really necessary, though. Where do we need to decode them?

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -81,6 +84,23 @@ object UpdateEventItem {
...
+sealed trait RetrieveEvents
+case class EventId(id: Long) extends RetrieveEvents
+case class EventExternalId(externalId: String) extends RetrieveEvents
+
+object RetrieveEventCoders {
+  implicit val encodeRetrieveEvents: Encoder[RetrieveEvents] = new Encoder[RetrieveEvents] {
+    override def apply(a: RetrieveEvents): Json =
+      Json.obj(a match {
+        case EventId(id) => ("id", Json.fromLong(id))
+        case EventExternalId(id) => ("externalId", Json.fromString(id))
```
wjoel: This seems to be doing the same encoding as the default/inferred encoder would do...?

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -184,7 +204,50 @@ class EventsRelation(config: RelationConfig)(@transient val sqlContext: SQLConte
     (postUpdate, postNewItems).parMapN((_, _) => ())
   }
 
-  override val fieldsWithPushdownFilter = Seq("type", "subtype", "assetIds", "startTime", "source")
+  override val fieldsWithPushdownFilter: Seq[String] =
+    Seq("type", "subtype", "assetIds", "startTime", "source")
+
+  override def buildScan(requiredColumns: Array[String], filters: Array[Filter]): RDD[Row] = {
+    val b = urlWithBody(filters, byIdUrl)
+    val u = urlsWithFilters(filters, listUrl)
```
wjoel: This doesn't seem to be used, which seems a bit suspicious?

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -184,7 +204,50 @@ class EventsRelation(config: RelationConfig)(@transient val sqlContext: SQLConte
...
+          if (config.collectMetrics) {
+            itemsRead.inc()
+          }
+          toRow(e, requiredColumns)
+        },
+        config,
+        partitionUrlBodies(b)
+      )
+    } else {
+      super.buildScan(requiredColumns, filters)
```
wjoel: We don't call `super.buildScan` anywhere else, why do we need it here? What does it do?

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
```diff
@@ -161,6 +162,54 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
...
+      .format("com.cognite.spark.datasource")
+      .option("apiKey", writeApiKey.apiKey)
+      .option("type", "events")
+      .option("collectMetrics", "true")
+      .option("metricsPrefix", metricsPrefix)
+      .load()
+      .where("""id In(
+        607444033860, 3965637099169,
+        10477877031034, 17515837146970,
+        19928788984614, 21850891340773) and description = "eventspushdowntest" """)
```
wjoel: Add a test with `or` as well.

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
```diff
@@ -161,6 +162,54 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
...
+  it should "handle pusdown filters on many eventIds" taggedAs WriteTest in {
+    val metricsPrefix = "pushdown.filters.ids"
+    val df = spark.read
+      .format("com.cognite.spark.datasource")
+      .option("apiKey", writeApiKey.apiKey)
+      .option("type", "events")
+      .option("collectMetrics", "true")
+      .option("metricsPrefix", metricsPrefix)
+      .load()
+      .where("id In(607444033860, 3965637099169, 10477877031034, 17515837146970, 19928788984614, 21850891340773)")
```
wjoel: Add a test using both `id in` and `externalId in`. I think that for now, it's ok to not support externalId at all.

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
wjoel: Looks like it's allowed in 0.6, but not in v1: https://doc.cognitedata.com/api/v1/#operation/getEventsByIds
I think we can just not bother with predicate pushdown for externalId for now.

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
```diff
@@ -161,6 +162,54 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
...
+    val metricsPrefix = "pushdown.filters.idsAndDescription"
+    val df = spark.read
+      .format("com.cognite.spark.datasource")
+      .option("apiKey", writeApiKey.apiKey)
+      .option("type", "events")
+      .option("collectMetrics", "true")
+      .option("metricsPrefix", metricsPrefix)
+      .load()
+      .where("""id In(
+        607444033860, 3965637099169,
```
wjoel: Add a test with something lik e`id in ("asdf")` (invalid value) and make sure the error message looks ok.

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -184,7 +204,50 @@ class EventsRelation(config: RelationConfig)(@transient val sqlContext: SQLConte
...
+      .toSeq
+
+  def urlWithBody(filters: Array[Filter], uri: Uri): UriWithBody[Seq[RetrieveEvents]] = {
+    val elements = for {
+      field <- Seq("id", "externalId")
+      filter <- filters
+      id <- getFilter(filter, field)
+    } yield (field, id)
+    val e = elements.map {
+      case ("id", id) => EventId(id.toLong)
```
wjoel: This might need something else to get a better error message.

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
```diff
@@ -184,7 +204,50 @@ class EventsRelation(config: RelationConfig)(@transient val sqlContext: SQLConte
...
+      .grouped(1000)
+      .map(UriWithBody(eventUriWithBody.uri, _))
+      .toSeq
+
+  def urlWithBody(filters: Array[Filter], uri: Uri): UriWithBody[Seq[RetrieveEvents]] = {
+    val elements = for {
+      field <- Seq("id", "externalId")
+      filter <- filters
+      id <- getFilter(filter, field)
+    } yield (field, id)
```
wjoel: Seems we can just yield `EventId(id.toLong)` here instead, if we skip the support for EventExternalId.

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
wjoel: I see you're already using that, so this should still be ok as private, I think.

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
```diff
@@ -590,6 +639,35 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
     assert(idsAfterDelete.length == 0)
   }
 
+  it should "encode and decode RetrieveEvents" in {
```
wjoel: We (currently and possibly never) don't have such tests, since they will/should be caught by the integration tests.

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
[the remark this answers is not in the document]
ebrattli: It was mostly just two have the whole pair, but I can remove the decoder

.scalafmt.conf
ebrattli: it changes all expressions like:
`something.map(_ or _)`
into
`something.map(_.or(_))`

ebrattli: same thing for our unit tests

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
ebrattli: I saw that the python-sdk allows for retrieval by id and external id. https://cognite-sdk-python.readthedocs-hosted.com/en/latest/cognite.html#retrieve-multiple-events-by-id

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
ebrattli: The autoencoder would not properly encode the sealed trait class.
https://circe.github.io/circe/codecs/adt.html

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
ebrattli: The urlWithBody gets the filters on `"id", "externalId"`. If these are not present, then we would like to call super.buildscan which does pushdown on the other filters `"type", "subtype", "assetIds", "startTime", "source"`.

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
ebrattli: This error message will never happen since we have specified the fields further up as constants. The error is just to be able to match all cases.

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
ebrattli: It does seem that spark ignores illegal values. When I create this test, spark does not even call buildscan.

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
ebrattli: removed

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
ebrattli: How can I find an externalId?

.scalafmt.conf
wjoel: Yes, `.map(_.or(_))` is better, and should probably be written as `.map((x, y) => x.or(y))` anyway. Formatting in unit tests is less strict (for now) and should probably be visited to be rewritten in a style different than FlatSpec some day.

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
wjoel: I meant for `id.toLong`, is it possible to get to cause an exception?

src/main/scala/com/cognite/spark/datasource/CdpConnector.scala
wjoel: Do we actually need and use that return value, though?

ebrattli: How would we otherwise return the data that we receive from the request? The return value is used similarly to the way it is used by `get`.

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
ebrattli: It is hard to make it fail, because as noted above if the user types in something that is not a number, then `buildscan` is not even called.

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
ebrattli: removed

.scalafmt.conf
ebrattli: I put it back

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
ebrattli: done

src/main/scala/com/cognite/spark/datasource/EventsRelation.scala
wjoel: I see, this would be easier to follow if we checked for `id` filters separately and said `if (idFilters.nonEmpty) { CdpEventsByIdsRdd(...)` and so on.

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
```diff
@@ -161,6 +162,71 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
...
+    val metricsPrefix = "pushdown.filters.orids"
+    val df = spark.read
+      .format("com.cognite.spark.datasource")
+      .option("apiKey", writeApiKey.apiKey)
+      .option("type", "events")
+      .option("collectMetrics", "true")
+      .option("metricsPrefix", metricsPrefix)
+      .load()
+      .where("""
+          id = 607444033860 or id = 3965637099169 or id = 10477877031034 or
```
wjoel: This is also useful, but I meant `id = 607444033860 or subtype="something"` :)

src/main/scala/com/cognite/spark/datasource/CdpConnector.scala
wjoel: Oh, right. Of course.

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
```diff
@@ -590,6 +656,21 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
     assert(idsAfterDelete.length == 0)
   }
 
+  it should "encode RetrieveEvents to the API schema" in {
```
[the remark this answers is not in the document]
ebrattli: removed

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
ebrattli: Ahh, this was a good test as it revealed a bug in the implementation.

src/main/scala/com/cognite/spark/datasource/CdpEventsRdd.scala
```diff
@@ -0,0 +1,46 @@
...
+    extends RDD[Row](sparkContext, Nil)
+    with CdpConnector {
+
+  // This needs to be broadcast here since compute will be called from the executors
+  // where the SparkContext is not available
+  private val broadcastApplicationId = sparkContext.broadcast(sparkContext.applicationId)
+  override def getPartitions: Array[Partition] =
+    urlsWithBody.zipWithIndex.map {
+      case (u, i) =>
+        CdpEventByIdsRddPartition(Some(1000), i, u)
```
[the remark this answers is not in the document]
ebrattli: I realized that the urls was aleready partitioned

src/main/scala/com/cognite/spark/datasource/CdpEventsRdd.scala
```diff
@@ -0,0 +1,46 @@
...
+    size: Option[Int],
+    index: Int,
+    urlWithBody: UriWithBody[Seq[RetrieveEvents]])
+    extends Partition
+
+case class CdpEventsByIdsRdd(
+    @transient override val sparkContext: SparkContext,
+    toRow: EventItem => Row,
+    config: RelationConfig,
+    urlsWithBody: Seq[UriWithBody[Seq[RetrieveEvents]]])
```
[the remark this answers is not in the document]
ebrattli: done

src/main/scala/com/cognite/spark/datasource/CdpEventsRdd.scala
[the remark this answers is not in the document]
ebrattli: fixed

PR #283, opened 2019-07-31
Document filter pushdown capabilities

hakontro:


hakontro:
@ebrattli Thanks! True that. We've kept out under the hood-details from the Readme this far, but we should definitely add it if we decide to change that in the future.

PR #284, opened 2019-07-31
Remove metadata keys with null values

hakontro:
This was already done for insertInto, but is now done for all ways of
writing to CDF.

### DIFF
--- a/src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
+++ b/src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
@@ -120,7 +120,10 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
 
   override def update(rows: Seq[Row]): IO[Unit] = {
     val updateAssetsItems = rows
-      .map(r => fromRow[UpdateAssetsItemBase](r))
+      .map { r =>
+        val assetItem = fromRow[UpdateAssetsItemBase](r)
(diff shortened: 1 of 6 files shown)


hakontro:
```
rows.map { r =>
  val items = fromRow[ResourceItem](r)
  items.copy(metadata = filterMetadata(items.metadata))
}
```
Couldn't find a good way to abstract over this pattern. Writing a function like `def getFromRowAndFilterMetadata[A <: HasMetadata](rows: Seq[Row])` is tricky, since the copy-method is auto-generated for case classes. Anyone see a good way?

src/main/scala/com/cognite/spark/datasource/AssetsRelation.scala
```diff
@@ -120,7 +120,10 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
 
   override def update(rows: Seq[Row]): IO[Unit] = {
     val updateAssetsItems = rows
-      .map(r => fromRow[UpdateAssetsItemBase](r))
+      .map { r =>
+        val assetItems = fromRow[UpdateAssetsItemBase](r)
```
wjoel: `assetItem`

hakontro: True! Found a few more of those.

hakontro:
@wjoel Good catch, new tests caught some bugs.

PR #288, opened 2019-08-13
Initial rewrite to API v1 using Scala SDK

hakontro:


macro/src/main/scala/SparkSchemaHelper.scala
```diff
@@ -11,7 +11,18 @@ class SparkSchemaHelperImpl(val c: Context) {
     import c.universe._
 
     val constructor = weakTypeOf[T].decl(termNames.CONSTRUCTOR).asMethod
-    val params = constructor.paramLists.flatten.map(param => q"$x.${param.name.toTermName}")
+    val params = constructor.paramLists.flatten
+      .map(param => {
+        val baseExpr = q"$x.${param.name.toTermName}"
+        def toSqlTimestamp(expr: c.Tree) = q"new java.sql.Timestamp($expr.toEpochMilli())"
```
Tapped: Hmm, I found a better way of converting between Timestamp and Instant.
See https://www.baeldung.com/java-time-instant-to-java-sql-timestamp

I think maybe the current solution will loose zone information, while the approach in the link above will keep the zone info,

src/main/scala/com/cognite/spark/datasource/SdkRddV1.scala
```diff
@@ -0,0 +1,41 @@
...
+        }
+        SdkPartition(cursor, partitionSize, index)
+    }.toArray
+  override def compute(_split: Partition, context: TaskContext): Iterator[Row] = {
+    val split = _split.asInstanceOf[SdkPartition]
+    implicit val cs =
+      IO.contextShift(ExecutionContext.fromExecutor(Executors.newCachedThreadPool()))
+    implicit val sttpBackend = AsyncHttpClientCatsBackend[cats.effect.IO]()
+    implicit val auth = config.auth
+    val client = new GenericClient[IO, Nothing]("cdp-spark-datasource")
```
Tapped: Does creating a new client cause a HTTP client pool to be made? We should share the client if that's the case.

src/main/scala/com/cognite/spark/datasource/ThreeDModelsRelation.scala
```diff
@@ -1,19 +1,51 @@
...
 
-  override def listUrl(): Uri =
-    uri"${config.baseUrl}/api/0.6/projects/${config.project}/3d/models"
+  def compute(c: GenericClient[IO, Nothing], p: SdkPartition): Iterator[Row] = {
+    if (config.collectMetrics) {
+      itemsRead.inc()
+    }
+    val items: IO[common.ItemsWithCursor[ThreeDModel]] =
+      c.threeDModels.readWithCursor(p.cursor, config.limit.map(_.toLong))
+    val it = items.unsafeRunSync()
```
Tapped: I would rather like the `unsafeRunSync` part to be at the end of `readWithCursor`

src/main/scala/com/cognite/spark/datasource/SdkRddV1.scala
erlendvollset: Shouldn't sharing connection pools be handled in the SDK?

Tapped: Yes, so I assume each Client got a connection pool, and then you share the client across the code. Though I think it's the sttpBackend which stores the connection pool, and at the moment we create a new sttpBackend per partition, which is ineffective, since Spark will probably run multiple partitions per worker (multiple CPU cores). Instead we should share the sttpBackend and maybe the scala sdk client (depends if it actually allocates resources, or if it's just proxies stuff down to the backend).

src/main/scala/com/cognite/spark/datasource/SdkV1Rdd.scala
```diff
@@ -0,0 +1,41 @@
...
+          case (None, s) => s
+          case (Some(l), Some(s)) => Some(scala.math.min(l, s))
+          case (l, None) => l
+        }
+        SdkPartition(cursor, partitionSize, index)
+    }.toArray
+  override def compute(_split: Partition, context: TaskContext): Iterator[Row] = {
+    val split = _split.asInstanceOf[SdkPartition]
+    implicit val auth = config.auth
+    val client = new GenericClient[IO, Nothing]("cdp-spark-datasource")
```
wjoel: We should include the Spark data source version in the application name to be able to track usage for different releases.

src/main/scala/com/cognite/spark/datasource/ThreeDModelsRelation.scala
hakontro: Definitely, good catch.

src/main/scala/com/cognite/spark/datasource/SdkRddV1.scala
hakontro: Good points! Made this work with a single sttpBackend and client.

src/main/scala/com/cognite/spark/datasource/SdkV1Rdd.scala
hakontro: Good point, I'll use `Constants.SparkDatasourceVersion` to match with `x-cdp-sdk`.

src/main/scala/com/cognite/spark/datasource/FilesRelation.scala
```diff
@@ -25,64 +26,34 @@ case class FileItem(
...
-  def postFiles(rows: Seq[Row]): IO[Unit] = {
-    val fileItems = rows.map { r =>
-      val fileItem = fromRow[FileItem](r)
-      fileItem.copy(metadata = filterMetadata(fileItem.metadata))
+  override def getFromRowAndCreate(rows: Seq[Row]): IO[Seq[File]] = {
+    val fileUpdates = rows.map { r =>
+      val file = fromRow[File](r)
+      fileUpdateFromFile(file.copy(metadata = filterMetadata(file.metadata)))
     }
+    client.files.update(fileUpdates)
```
wjoel: Couldn't we use `updateFromRead`, after filtering the metadata?

src/test/scala/com/cognite/spark/datasource/AssetsRelationTest.scala
```diff
@@ -58,7 +55,7 @@ class AssetsRelationTest extends FlatSpec with Matchers with SparkTest {
     assert(df.count() == 10)
   }
 
-  it should "support pushdown filters on name and depth" taggedAs ReadTest in {
+  it should "support pushdown filters on name and depth" taggedAs ReadTest ignore {
```
wjoel: Rather than ignoring it we should change it to not use depth, I guess?

src/test/scala/com/cognite/spark/datasource/AssetsRelationTest.scala
```diff
@@ -91,16 +88,13 @@ class AssetsRelationTest extends FlatSpec with Matchers with SparkTest {
...
-        |null as depth,
+        |select 1 as externalId,
         |'asset name' as name,
         |null as parentId,
         |'asset description' as description,
-        |null as types,
         |null as metadata,
         |'$assetsTestSource' as source,
-        |null as sourceId,
+        |10 as id,
```
wjoel: sourceId should become externalId, while the id above should remain "null as id", I think?

src/test/scala/com/cognite/spark/datasource/AssetsRelationTest.scala
```diff
@@ -187,7 +178,7 @@ class AssetsRelationTest extends FlatSpec with Matchers with SparkTest {
       rows => rows.length < 1)
   }
 
-  it should "support upserts when using insertInto()" taggedAs WriteTest in {
+  it should "support upserts when using insertInto()" taggedAs WriteTest ignore {
```
wjoel: Shouldn't be ignored.

src/test/scala/com/cognite/spark/datasource/AssetsRelationTest.scala
```diff
@@ -197,64 +188,56 @@ class AssetsRelationTest extends FlatSpec with Matchers with SparkTest {
       rows => rows.length > 0)
 
     // Post new assets
-    spark
+    val somedf = spark
```
wjoel: `someDf` but I see this is only for debugging ;)

src/test/scala/com/cognite/spark/datasource/BaseUrlTest.scala
```diff
@@ -6,19 +6,19 @@ class BaseUrlTest extends FlatSpec with SparkTest {
...
   private val readApiKey = System.getenv("TEST_API_KEY_READ")
 
-  it should "read different files metadata from greenfield and api" taggedAs GreenfieldTest in {
+  it should "read different files metadata from greenfield and api" taggedAs GreenfieldTest ignore {
 
     val dfGreenfield = spark.read
       .format("com.cognite.spark.datasource")
       .option("apiKey", greenfieldApiKey)
-      .option("type", "files")
+      .option("type", "assets")
```
wjoel: I guess it doesn't matter too much but we picked files because it has fewer items than assets. Using assets will probably be a bit slower.

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
```diff
@@ -111,9 +112,9 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
       .option("metricsPrefix", metricsPrefix)
       .load()
       .where(s"startTime > 1584601199999")
-    assert(df.count == 3)
+    assert(df.count == 2)
     val eventsRead = getNumberOfRowsRead(metricsPrefix, "events")
-    assert(eventsRead == 3)
+    assert(eventsRead == 2)
```
wjoel: Why does this change?

src/main/scala/com/cognite/spark/datasource/FilesRelation.scala
hakontro: Makes a lot of sense, thanks!

src/test/scala/com/cognite/spark/datasource/AssetsRelationTest.scala
hakontro: Just ignoring the tests for things I haven't implemented yet, such as pushdown filters and upsert.

src/test/scala/com/cognite/spark/datasource/BaseUrlTest.scala
hakontro: Agreed, changed this since the sdk couldn't read from Greenfield, so it broke the test. Can change it back once we have that in place.

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
hakontro: Excellent question, I did no changes to Events in this PR so some test data must have changed.

src/test/scala/com/cognite/spark/datasource/AssetsRelationTest.scala
hakontro: Yep!

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
hakontro: Looks like a few of these tests were actually reading test data created by other tests, and so were based on them having completed successfully on the previous run. I'll change them to not do that.

src/main/scala/com/cognite/spark/datasource/DataPointsRelation.scala
```diff
@@ -1,5 +1,7 @@
 package com.cognite.spark.datasource
 
+import java.util.concurrent.Executors
```
wjoel: Is this still being used?

src/test/scala/com/cognite/spark/datasource/AssetsRelationTest.scala
wjoel: Alright then, let's fix that next. :)

src/test/scala/com/cognite/spark/datasource/AssetsRelationTest.scala
```diff
@@ -220,41 +208,33 @@ class AssetsRelationTest extends FlatSpec with Matchers with SparkTest {
...
-                 |types,
-                 |'bar' as description,
-                 |map("foo", null, "bar", "test") as metadata,
-                 |source,
-                 |sourceId,
-                 |createdTime,
-                 |lastUpdatedTime
-                 |from destinationAssets
-                 |where source = '$source'""".stripMargin)
+              |select externalId,
```
wjoel: Are we intentionally changing the order here because we have to? Maybe having `id` first is the better order? We can change this in the SDK to make it "the most natural". What do you think, @hakontro and @Tapped?

src/main/scala/com/cognite/spark/datasource/DataPointsRelation.scala
hakontro: Nope! Couple of dead imports there, I'll remove them.

src/test/scala/com/cognite/spark/datasource/AssetsRelationTest.scala
Tapped: Jetfire have stopped using the `insertable` functionality of relations for all resource types except Datapoints. I think we should stop having insertable relations, and instead only support save mode.

hakontro: No conscious decision on the order for these tests, the correct order is ensured by `.select(destinationDf.columns.map(col): _*)` anyways.

hakontro: @Tapped we need `insertable` for writing with SparkSQL, I believe?

Tapped: Yeah, you'll need that but it isn't a pleasant experience, since read and write schema is in practice different + you've to make the order of the columns correct.

wjoel: But in order to use `insert into` with Spark SQL, this is what's being used. We have no choice but to support it.

So the question is what order makes sense for the user. Probably best to match the order in the documentation? I think I tried to do that with the Scala SDK but we should double check it.

PR #293, opened 2019-09-04
Rewrite events to use V1

hakontro:
Well aware this is wrecking the test coverage, but that'll change when we remove the old files (`CdpRelation`, `CdpRdd`, `EventsRelation`, `AssetsRelation`). At least for Assets I think we should make a way to use the old version, since Asset types won't be in `v1` for a while. EventsRelation could be deleted, but pushdown filters aren't ready in EventsRelationV1 yet, so we could make `0.6` avilable here too.

### DIFF
--- a/src/main/scala/com/cognite/spark/datasource/EventsRelationV1.scala
+++ b/src/main/scala/com/cognite/spark/datasource/EventsRelationV1.scala
@@ -1,0 +1,93 @@
+package com.cognite.spark.datasource
+
+import cats.effect.IO
+import com.cognite.sdk.scala.v1.{Event, EventUpdate, GenericClient}
+import com.cognite.sdk.scala.v1.resources.Events
+import com.cognite.spark.datasource.SparkSchemaHelper.{asRow, fromRow, structType}
+import com.softwaremill.sttp.Uri
+import com.softwaremill.sttp._
+import org.apache.spark.sql.{Row, SQLContext}
+import org.apache.spark.sql.sources.InsertableRelation
+import io.circe.generic.auto._
+import org.apache.spark.sql.types.{DataTypes, StructType}
+import cats.implicits._
+import com.cognite.sdk.scala.common.CdpApiException
+
+class EventsRelationV1(config: RelationConfig)(val sqlContext: SQLContext)
+    extends SdkV1Relation[Event, Events[IO], EventItem](config, "events")
+    with InsertableRelation {
+
+  override def insert(rows: Seq[Row]): IO[Unit] = {
+    val events = fromRowWithFilteredMetadata(rows)
+    client.events.createFromRead(events) *> IO.unit
+  }
+
+  override def update(rows: Seq[Row]): IO[Unit] = {
+    val events = rows.map(r => fromRow[EventUpdate](r))
+    client.events.update(events) *> IO.unit
+  }
+
+  override def delete(rows: Seq[Row]): IO[Unit] = {
+    val ids = rows.map(r => fromRow[DeleteItem](r).id)
+    client.events.deleteByIds(ids)
+  }
+
+  override def upsert(rows: Seq[Row]): IO[Unit] = getFromRowAndCreate(rows)
+
+  def fromRowWithFilteredMetadata(rows: Seq[Row]): Seq[Event] =
+    rows.map { r =>
+      val event = fromRow[Event](r)
+      event.copy(metadata = filterMetadata(event.metadata))
+    }
+  override def getFromRowAndCreate(rows: Seq[Row]): IO[Unit] = {
+    val events = fromRowWithFilteredMetadata(rows)
+
+    client.events
+      .createFromRead(events)
+      .handleErrorWith {
+        case e: CdpApiException =>
+          if (e.code == 409) {
+            val existingExternalIds =
+              e.duplicated.get.map(j => j("externalId").get.asString.get)
+            resolveConflict(existingExternalIds, events)
+          } else { IO.raiseError(e) }
+      } *> IO.unit
+  }
+
+  def resolveConflict(existingExternalIds: Seq[String], events: Seq[Event]): IO[Unit] = {
+    import CdpConnector.cs
+    val (eventsToUpdate, eventsToCreate) = events.partition(
+      p => existingExternalIds.contains(p.externalId.get)
+    )
+
+    val idMap = client.events
+      .retrieveByExternalIds(existingExternalIds)
+      .unsafeRunSync()
+      .map(e => e.externalId -> e.id)
+      .toMap
+
+    val create =
+      if (eventsToCreate.isEmpty) IO.unit else client.events.createFromRead(eventsToCreate)
+    val update =
+      if (eventsToUpdate.isEmpty) { IO.unit } else {
+        client.events.updateFromRead(eventsToUpdate.map(e => e.copy(id = idMap(e.externalId))))
+      }
+
+    (create, update).parMapN((_, _) => ())
+  }
+
+  override def schema: StructType = structType[Event]
+
+  override def toRow(a: Event): Row = asRow(a)
+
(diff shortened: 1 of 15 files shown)


src/test/scala/com/cognite/spark/datasource/AssetsRelationTest.scala
```diff
@@ -94,10 +94,11 @@ class AssetsRelationTest extends FlatSpec with Matchers with SparkTest {
         |'asset description' as description,
         |null as metadata,
         |'$assetsTestSource' as source,
-        |null as id,
-        |null as createdTime,
-        |null as lastUpdatedTime
+        |10 as id,
+        |0 as createdTime,
```
wjoel: Is "null" valid here, or do we need the explicit 0 now?

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -33,301 +30,98 @@ case class TimeSeriesItem(
...
+    baseTimeSeriesUrl(config.project, version)
+}
 object TimeSeriesRelation
     extends DeleteSchema
     with UpsertSchema
     with InsertSchema
-    with UpdateSchema {
-  val insertSchema = structType[PostTimeSeriesItem]
-  val upsertSchema = StructType(structType[TimeSeriesItem].filterNot(field =>
+//    with UpdateSchema
```
wjoel: No more update for time series? Coming back in the next PR?

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -33,301 +30,98 @@ case class TimeSeriesItem(
...
-        timeSeriesUrl,
-        postTimeSeriesToCreate
-      ).flatTap { _ =>
-        IO {
-          if (config.collectMetrics) {
-            timeSeriesCreated.inc(postTimeSeriesToCreate.length)
-          }
-        }
+    val create =
+      if (timeSeriesToCreate.isEmpty) { IO.unit } else {
```
wjoel: I think this could be something like:

```
timeSeriesToCreate.map(client.timeSeries.createFromRead).getOrElse(IO.unit)
```

and similarly for other `if (x.isEmpty) { IO.unit } else { f(x.map(...)) }`, but I haven't tested it.

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
```diff
@@ -33,301 +30,98 @@ case class TimeSeriesItem(
...
+      }
+    val update =
+      if (timeSeriesToUpdate.isEmpty) { IO.unit } else {
+        client.timeSeries.updateFromRead(timeSeriesToUpdate.map(ts =>
+          ts.copy(id = idMap(ts.externalId))))
       }
-    }
 
-    (putItems, postItems).parMapN((_, _) => ())
+    (create, update).parMapN((_, _) => ())
```
wjoel: Do we still log something useful if something fails?

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
```diff
@@ -563,8 +566,8 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
         .option("onconflict", "update")
         .save()
     }
-    e.getCause shouldBe a[CdpApiException]
-    val cdpApiException = e.getCause.asInstanceOf[CdpApiException]
+    e.getCause shouldBe a[com.cognite.sdk.scala.common.CdpApiException]
```
wjoel: Makes more sense to have it imported and not need to namespace it like this, imo.

src/main/scala/com/cognite/spark/datasource/EventsRelationV1.scala
```diff
@@ -0,0 +1,92 @@
...
+    client.events
+      .createFromRead(events)
+      .handleErrorWith {
+        case e: com.cognite.sdk.scala.common.CdpApiException =>
+          if (e.code == 409) {
+            val existingExternalIds =
+              e.duplicated.get.map(j => j("externalId").get.asString.get)
+            resolveConflict(existingExternalIds, events)
+          } else { IO.raiseError(e) }
+      } *> IO.unit
```
wjoel: I might be wrong here but I think this ignores the possible error from the left hand side, and this should `.map(IO.unit)` instead... @Tapped what do you think?

Tapped: No, it doesn't, an IO product takes the product of the right side if there are no error, however when there are an error it will fail-fast.

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
hakontro: Still update for time series! This is just for exposing the schema, which wasn't being used by anyone right now so I skipped writing a special macro for schemas with Setter/NonNullableSetters.

src/test/scala/com/cognite/spark/datasource/AssetsRelationTest.scala
hakontro: Needs to be explicit, `name`, `id`, `createTime` and `lastUpdatedTime` are no longer `Option`s since I'm casting `Row`s to `sdk.scala.Asset`s.

wjoel:
Overall it looks good to me, just some small questions.

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
hakontro: Would love something cleaner, but couldn't find a better way. `.map` will map over each `TimeSeries`. `Some(timeSeriesToCreate).map().getOrElse` doesn't work since empty list, curiously, is considered `Some` rather than `None`.

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
hakontro: This will log any exception returned from the SDK upon failure.

src/test/scala/com/cognite/spark/datasource/EventsRelationTest.scala
hakontro: Agreed, I'll rename the `CdpApiException` in `CdpConnector` to something else (we can replace it with the SdkException in another PR).

hakontro: Good point. This was to get around the name collision with `CdpConnector.CdpApiException`, but that could be replaced with the one from the sdk without too much hassle.

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
wjoel: Cool :)

src/test/scala/com/cognite/spark/datasource/AssetsRelationTest.scala
wjoel: Hmm, I wonder if we should manage that in the Spark layer. Maybe we should get user feedback (as well as test it out ourselves and see how we feel about it) first though.

src/main/scala/com/cognite/spark/datasource/TimeSeriesRelation.scala
wjoel: Oh, I see! So it would be much nicer if `client.timeSeries.createFromRead` allowed an empty list and did the right thing by just returning an en empty `Items[Asset]` or something similar, right?

I think we can make that happen. It's ok to leave this for now, we can clean it up later.

hakontro: Yep!

PR #295, opened 2019-09-05
Fix a bug in time series upsert

hakontro:
Time series upsert was only working correctly if rows to be updated had
an ID, while rows to be inserted had no id.


hakontro:
This was only run to make Jenkins publish the artifact to OSS Sonatype.

PR #296, opened 2019-09-05
Release version 0.4.13

hakontro:
This was released in a hot fix, where test data was changed, so had to change tests here as well to make them pass.


hakontro:
Yes, released it manually, to avoid releasing the `v1` stuff on the master branch.

PR #298, opened 2019-09-09
Fix a bug in time series upsert

hakontro:
Time series upsert was only working correctly if rows to be updated had
an ID, while rows to be inserted had no id.


build.sbt
```diff
@@ -14,7 +14,7 @@ lazy val commonSettings = Seq(
   organization := "com.cognite.spark.datasource",
   organizationName := "Cognite",
   organizationHomepage := Some(url("https://cognite.com")),
-  version := "0.4.12",
+  version := "0.4.14",
```
wjoel: With this one we're on CDF v1, right? I think we should bump it to 0.5.x at least.

hakontro: Only for time series, and soft `v1` I guess since it doesn't actually expose `externalId` for time series. I suggest postponing the bump until next release which will include scala SDK and several resources to `v1`.

PR #299, opened 2019-09-10
Support pushdown filters on Assets in v1

hakontro:


src/main/scala/com/cognite/spark/datasource/AssetsRelationV1.scala
```diff
@@ -1,23 +1,46 @@
...
+    val getAll = shouldGetAll(pushdownFilterExpression, fieldNames)
+    val params = pushdownToParameters(pushdownFilterExpression)
+
+    val pushdownFilters = if (params.isEmpty || getAll) {
+      Seq(AssetsFilter())
+    } else {
+      params.map(assetsFilterFromMap)
+    }
+
+    pushdownFilters.map(f => client.assets.filterWithCursor(f, cursor, limit))
```
wjoel: I added a commit with a suggestion showing how to use `parTraverse` instead, that would fetch those in parallel.

src/main/scala/com/cognite/spark/datasource/AssetsRelationV1.scala
```diff
@@ -1,23 +1,46 @@
...
+  }
+
+  def assetsFilterFromMap(m: Map[String, String]): AssetsFilter =
+    AssetsFilter(name = m.get("name"), source = m.get("source"))
```
wjoel: Clever! :) Maybe make it `private`.

src/main/scala/com/cognite/spark/datasource/SdkV1Relation.scala
```diff
@@ -75,6 +85,16 @@ abstract class SdkV1Relation[A, T <: Readable[A, IO], C: Decoder](
...
     })
 
+  def toRow(item: A, requiredColumns: Array[String]): Row =
+    if (requiredColumns.isEmpty) {
+      toRow(item)
+    } else {
+      val fieldNamesInOrder = item.getClass.getDeclaredFields.map(_.getName)
+      val indices = requiredColumns.map(f => fieldNamesInOrder.indexOf[String](f))
+      val fullRow = toRow(item)
+      Row.fromSeq(indices.map(fullRow.get))
```
wjoel: Maybe `idx => fullRow.get(idx)` would be easier to follow (at least for me - I had to double check the type of `fullRow`)

src/main/scala/com/cognite/spark/datasource/AssetsRelationV1.scala
hakontro: Nice, thanks!

src/main/scala/com/cognite/spark/datasource/SdkV1Relation.scala
hakontro: Agreed, tried to make it a bit clearer now.

src/main/scala/com/cognite/spark/datasource/AssetsRelationV1.scala
hakontro: Done!

src/main/scala/com/cognite/spark/datasource/AssetsRelationV1.scala
hakontro: Looks like we're running into problems serializing `ContextShift`. See any way around it?
https://github.com/typelevel/cats-effect/issues/587

wjoel: We use `ContextShift` elsewhere, so it should be possible. Maybe it needs to be marked as transient and/or lazy?

wjoel: I guess that's why we define it inline with the method in other places, but it would be better if we could have it as an instance variable.

PR #305, opened 2019-09-25
Events pushdown and parallel retrieval

hakontro:
- Events filter pushdown
- Retrieval by partitions
- No more cursors when using the Scala SDK, and streams instead of `IO`
- Deleted a bunch of old ResourceRelations/Rdds/case classes
- Some, mostly unrelated, rework of TimeSeriesRelationTest to accomodate for the new limit on one update/5 seconds


src/main/scala/com/cognite/spark/datasource/FilesRelation.scala
```diff
@@ -37,8 +23,13 @@ class FilesRelation(config: RelationConfig)(val sqlContext: SQLContext)
     client.files.updateFromRead(files) *> IO.unit
   }
 
-  override def clientToResource(client: GenericClient[IO, Nothing]): Files[IO] =
-    client.files
+  override def getStreams(filters: Array[Filter])(
+      client: GenericClient[IO, Nothing],
+      limit: Option[Long],
+      numPartitions: StatusCode): Seq[fs2.Stream[IO, File]] =
```
sebbert: Why is `numPartitions` a `StatusCode`?

hakontro: Good question! I have no idea how that got there, so will use `Int` here.

sebbert: Autocomplete is a double-edged sword 😄

src/main/scala/com/cognite/spark/datasource/EventsRelationV1.scala
```diff
@@ -1,21 +1,88 @@
...
+  def assetIdsFromWrappedArray(wrappedArray: String): Seq[Long] =
+    wrappedArray.split("\\D+").filter(_.nonEmpty).map(_.toLong)
+
+  def timeRangeFromMinAndMax(minTime: Option[String], maxTime: Option[String]): Option[TimeRange] =
+    (minTime, maxTime) match {
+      case (None, None) => None
+      case _ => {
+        val minimumTimeAsInstant =
+          minTime
+            .map(java.sql.Timestamp.valueOf(_).toInstant.plusMillis(1))
```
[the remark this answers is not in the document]
hakontro: This is to adjust for the API being inclusive, while the filter is exclusive.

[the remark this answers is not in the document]
sebbert: Gotcha, good to know 👍

sebbert:
Note: I'm not approving this, as I would like someone who has experience with the spark datasource to take a look as well.

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
```diff
@@ -90,23 +90,20 @@ object PushdownUtilities {
       case EqualTo(colName, value) => PushdownFilter(colName, value.toString)
       case EqualNullSafe(colName, value) => PushdownFilter(colName, value.toString)
       case GreaterThan(colName, value) =>
-        PushdownFilter("min" + colName.capitalize, toMinTimeFormat(value))
+        PushdownFilter("min" + colName.capitalize, value.toString)
```
wjoel: Hmm, we're sure we're ok to remove the `toMin/MaxTimeFormat`? That stuff is tricky. Maybe it's less tricky nowadays with V1?

wjoel: Ok, looks like we're doing that still but in other places.

src/main/scala/com/cognite/spark/datasource/FilesRelation.scala
```diff
@@ -37,13 +20,15 @@ class FilesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
   }
 
-  override def clientToResource(client: GenericClient[IO, Nothing]): Files[IO] =
-    client.files
+  override def getStreams(filters: Array[Filter])(
+      client: GenericClient[IO, Nothing],
+      limit: Option[Long],
+      numPartitions: Int): Seq[fs2.Stream[IO, File]] =
+    Seq(
+      config.limit.map(client.files.listWithLimit(_)).getOrElse(client.files.list)
```
wjoel: You can leave out `(_)` I think, the newline after `Seq(` seems excessive but alright ;)

src/main/scala/com/cognite/spark/datasource/SdkV1Rdd.scala
```diff
@@ -1,48 +1,43 @@
...
 import com.cognite.sdk.scala.v1.GenericClient
 import com.cognite.sdk.scala.common.Auth
 import org.apache.spark.{Partition, SparkContext, TaskContext}
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.Row
+import fs2.Stream
 
 case class SdkPartition(cursor: Option[String], size: Option[Int], index: Int) extends Partition
 
+case class SdkStream[A](stream: Stream[IO, A], index: Int) extends Partition
```
wjoel: Doesn't seem to be used anymore.

src/test/scala/com/cognite/spark/datasource/TimeSeriesRelationTest.scala
```diff
@@ -243,8 +275,46 @@ class TimeSeriesRelationTest extends FlatSpec with Matchers with SparkTest {
   }
 
   it should "support upsert in savemode" taggedAs WriteTest in {
+    val insertDescription = "spark-insert-test"
     val upsertDescription = "spark-upsert-test"
-    val saveModeUnit = "spark-savemode-test"
+
+    // Clean up any old test data
+    cleanUpTimeSeriesTestDataByUnit(testDataUnit)
```
wjoel: Just curious, why did this test change so much? We had just forgotten a cleanup in here or something?

src/main/scala/com/cognite/spark/datasource/FilesRelation.scala
hakontro: Couldn't get rid of `(_)` it seems, apparently couldn't decide between `scala.Option.map` and `cats.Functor.Ops.map`. Agree one-liner is better, though.

src/main/scala/com/cognite/spark/datasource/SdkV1Rdd.scala
hakontro: Good catch!

src/test/scala/com/cognite/spark/datasource/TimeSeriesRelationTest.scala
hakontro: Several of the tests in `TimeSeriesRelationTest` relied on data from previous tests still existing, and being in the state left from the previous test. This was annoying when wishing to rerung a single test, so I decided to fix it for all of them.

src/main/scala/com/cognite/spark/datasource/CdpRelation.scala
hakontro: Looks like this is actually not quite right, we're handling `<=` and `<` in the same way now. I'll fix it.

src/main/scala/com/cognite/spark/datasource/EventsRelationV1.scala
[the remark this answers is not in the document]
hakontro: This did not age well, considering I was wrong here! The adjustment should be the other way, and also be avoided for `LessThanOrEqual` and `GreaterThanOrEqual`.

PR #307, opened 2019-09-27
Support v1 for DataPoints

hakontro:
A crack at DataPointsV1, would be grateful for some early feedback. Not sure about schema, should we have different schemas for filters on id and externalId? Name is not returned from the API anymore. The easiest would be to just not have it in the DataFrame, but not sure if the users will like that.


src/main/scala/com/cognite/spark/datasource/DataPointsRelationV1.scala
```diff
@@ -0,0 +1,299 @@
...
+        client.dataPoints
+          .queryAggregatesByExternalId(
+            extId,
+            lowerTimeLimit,
+            upperTimeLimit,
+            granularity,
+            aggregations.map(_.aggregation))
+          .map(m => m.toSeq.flatMap(t => t._2)))
+    }
+    ioFromId ++ ioFromExternalId
```
wjoel: Use `merge` to make a concurrent stream, otherwise it will start fetching from `ioFromExternalId` only after `ioFromId` has been emptied.

src/main/scala/com/cognite/spark/datasource/DataPointsRelationV1.scala
```diff
@@ -0,0 +1,299 @@
...
+
+    if (aggregations.isEmpty && granularities.nonEmpty) {
+      sys.error(s"Granularity specified but no aggregation requested")
+    }
+
+    (aggregations, granularities)
+  }
+
+  // scalastyle:off cyclomatic.complexity
+  def toAggregationFilter(aggregation: String): AggregationFilter =
```
wjoel: Abbreviations like `avg`, `cv` etc. are no longer supported in V1, so this can (and must be) be simplified.

src/main/scala/com/cognite/spark/datasource/DataPointsRelationV1.scala
hakontro: This merges two `Seq[(DataPointsFilter, IO[stuff])` that are picked by index in compute, similar to the way `EventsRelationV1` uses a `Seq[Stream]`. Should be ok this way?

wjoel: Ah, my bad, I misunderstood the usage.

PR #314, opened 2019-10-07
Support multiple streams per compute task

hakontro:


src/main/scala/cognite/spark/SdkV1Rdd.scala
```diff
@@ -24,16 +29,55 @@ case class SdkV1Rdd[A](
...
+        val freshItems = s.toVector.filterNot(i => processedItems.contains(uniqueId(i)))
+        processedItems ++= freshItems.map(uniqueId)
+        Stream.eval(IO(queue.put(freshItems)))
+      }
+    } yield put
+
+    val f: Future[Unit] = Future {
+      pushIterator.compile.drain.unsafeRunSync()
+    }(drainPool)
+    new Iterator[Row] {
```
wjoel: I think we should move this logic to a separate helper class or function. It's quite hard to follow what's going on. It deserves a few comments, imo.

src/test/scala/cognite/spark/EventsRelationTest.scala
```diff
@@ -64,6 +64,22 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
...
 
+  it should "handle duplicates in a pushdown filter scenario" taggedAs WriteTest in {
+    val metricsPrefix = "single.pushdown.filter.duplicates"
+    val df = spark.read
+      .format("cognite.spark")
+      .option("apiKey", writeApiKey.apiKey)
+      .option("type", "events")
+      .option("collectMetrics", "true")
+      .option("metricsPrefix", metricsPrefix)
+      .option("partitions", "1")
```
wjoel: Try this with several partitions.

src/main/scala/cognite/spark/SdkV1Rdd.scala
```diff
@@ -1,21 +1,26 @@
...
 import org.apache.spark.sql.Row
-import fs2.Stream
+import fs2._
+
+import scala.collection.mutable
+import scala.concurrent.{ExecutionContext, Future}
 
 case class CdfPartition(index: Int) extends Partition
 
 case class SdkV1Rdd[A](
```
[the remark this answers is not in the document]
wjoel: Or we make the `getIOs` interface be "give me IOs for each partition" and let `getIOs` do merging as it pleases, and have number of partitions be = length of `getIOs` return value.

Would that work?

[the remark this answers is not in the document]
hakontro: That actually sounds preferable, since we need to make sure that filters against a single partition are computed in the same place (to remove the possible duplicates).

src/main/scala/cognite/spark/SdkV1Rdd.scala
```diff
@@ -24,16 +29,55 @@ case class SdkV1Rdd[A](
...
+      pushIterator.compile.drain.unsafeRunSync()
+    }(drainPool)
+    new Iterator[Row] {
+      var nextItems: Iterator[A] = Iterator.empty
+
+      override def hasNext: Boolean =
+        if (nextItems.hasNext) { true } else {
+          nextItems = iteratorFromQueue()
+          while (nextItems.isEmpty && !f.isCompleted) {
+            Thread.sleep(1)
```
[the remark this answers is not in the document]
hakontro: Thought about this for a while, but couldn't find a strictly better way. The queue is often cleared completely before all the streams have been completely drained, and since draining the queue and draining the streams is done by two different threads I don't see a way to do this with blocking.

src/main/scala/cognite/spark/SdkV1Rdd.scala
```diff
@@ -24,16 +29,55 @@ case class SdkV1Rdd[A](
...
+    implicit val contextShift: ContextShift[IO] =
+      IO.contextShift(ExecutionContext.fromExecutor(Executors.newFixedThreadPool(1)))
+
+    val processedItems = mutable.Set[Long]()
+
+    val streams = getStreams(client, config.limit, config.partitions, config.partitionsPerCompute)
+      .drop(split.index)
+      .next
+      .reduce(_.merge(_))
+    val queue = new ArrayBlockingQueue[Vector[A]](20)
```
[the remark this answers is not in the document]
hakontro: Should have been dynamic! The queue never seems to fill, so we can probably go pretty low here. Tested a bit and set it to `config.parallelismPerPartition * 2`.

src/test/scala/cognite/spark/AssetsRelationTest.scala
```diff
@@ -93,7 +93,24 @@ class AssetsRelationTest extends FlatSpec with Matchers with SparkTest {
...
+  it should "handle duplicates in a pushdown filter scenario" taggedAs ReadTest in {
+    val metricsPrefix = "pushdown.filters.assets.duplicates"
+    val df = spark.read
+      .format("cognite.spark")
+      .option("apiKey", writeApiKey.apiKey)
+      .option("type", "assets")
+      .option("collectMetrics", "true")
+      .option("metricsPrefix", metricsPrefix)
+      .option("limit", "1000")
+      .option("partitions", "1")
```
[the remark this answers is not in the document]
hakontro: Good call, discovered a bug in the implementation when I tried higher values.

src/test/scala/cognite/spark/EventsRelationTest.scala
hakontro: Good call, discovered a bug in the implementation when I tried higher values.

src/main/scala/cognite/spark/SdkV1Rdd.scala
wjoel: Much better, thanks.

src/main/scala/cognite/spark/AssetsRelation.scala
```diff
@@ -34,9 +34,15 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
 
-    pushdownFilters.flatMap { f =>
-      client.assets.filterPartitions(f, numPartitions, limit)
-    }
+    val streamsPerFilter = pushdownFilters
+      .map { f =>
+        client.assets.filterPartitions(f, numPartitions, limit)
+      }
+
+    // Merge streams related to each partition to make sure duplicate values are read into
```
wjoel: Nice.

src/main/scala/cognite/spark/DefaultSource.scala
```diff
@@ -95,18 +95,25 @@ class DefaultSource
         .getOrElse(throw new IllegalArgumentException(
           s"$onConflictName is not a valid onConflict option. Please choose one of the following options instead: ${OnConflict.values
             .mkString(", ")}"))
+    val parallelismPerPartition = {
+      val inputOrDefault = toPositiveInt(parameters, "parallelismPerPartition").getOrElse(
+        Constants.DefaultParallelismPerPartition)
+      if (inputOrDefault > partitions) { inputOrDefault } else { partitions }
```
wjoel: Hmm, why?

hakontro: Now that all filters for a partition are merged into one stream we'll get one or more empty `Iterator`s when grouping the streams in `.compute` if `parallelismPerPartition > partitions`. This avoids that. I suppose `max(parallelismPerPartitions, partitions) is nicer though, and I'll add a comment.

hakontro: I realize it's also pretty confusing that `partitions` is related to the partitions of a CDF endpoint, while `parallelismPerPartition` is related to Spark partitions. We'll need to use another name for both.

PR #319, opened 2019-10-10
Expose errors in reads

hakontro:
- Pass on errors and exceptions thrown by CDF streams
- The set handling duplicates when pushdown filters have overlap is now
threadsafe

### DIFF
--- a/src/main/scala/cognite/spark/SdkV1Rdd.scala
+++ b/src/main/scala/cognite/spark/SdkV1Rdd.scala
@@ -3,13 +3,14 @@ package cognite.spark
 import java.util.concurrent.{ArrayBlockingQueue, Executors}
 
 import cats.effect.{ContextShift, IO}
+import cats.implicits._
 import com.cognite.sdk.scala.v1.GenericClient
 import com.cognite.sdk.scala.common.Auth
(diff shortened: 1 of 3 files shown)


src/main/scala/cognite/spark/SdkV1Rdd.scala
```diff
@@ -52,10 +57,11 @@ case class SdkV1Rdd[A](
       .next
       .reduce(_.merge(_))
 
-    val queue: ArrayBlockingQueue[Vector[A]] =
-      new ArrayBlockingQueue[Vector[A]](config.parallelismPerPartition * 2)
+    val queue =
+      new EitherQueue(config.parallelismPerPartition * 2)
```
wjoel: Why "* 2"?

hakontro: "Seems to work" is the answer here, I guess. The queue never fills up, during testing at least, as Spark drains the queue quite a lot faster than the API delivers new data. I figured `*2` was an all right setting here, in the worst case it would store 2 chunks per stream before blocking. If the queue is getting filled up we might as well start blocking, but at least for the test setup we get the same performance by just using a constant value like 5.

PR #320, opened 2019-10-11
Add v1 support for RAW

hakontro:


src/main/scala/cognite/spark/AssetsRelation.scala
```diff
@@ -113,5 +113,5 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
 
   override def toRow(a: Asset): Row = asRow(a)
 
-  override def uniqueId(a: Asset): Long = a.id
+  override def uniqueId(a: Asset): String = a.id.toString
```
wjoel: Feels like we shouldn't need this if we just generalize over the return value type, somewhere?

hakontro: We could add it as a generic type to `SdkV1Relation` and `SdkV1Rdd`, but wanted to avoid another generic there for something so small. Long can always be cast to String, which would you prefer?

src/main/scala/cognite/spark/SdkV1Rdd.scala
```diff
@@ -20,7 +20,7 @@ case class SdkV1Rdd[A](
     @transient override val sparkContext: SparkContext,
     config: RelationConfig,
     toRow: A => Row,
-    uniqueId: A => Long,
+    uniqueId: A => String,
```
wjoel: Could we generalize to `SdkV1Rdd[A, I]` and have this be `uniqueId: A => I`, `processedItems: Set[I]` etc.?

src/main/scala/cognite/spark/SdkV1Rdd.scala
```diff
@@ -46,7 +46,7 @@ case class SdkV1Rdd[A](
     implicit val singleThreadedCs: ContextShift[IO] =
       IO.contextShift(ExecutionContext.fromExecutor(Executors.newFixedThreadPool(1)))
 
-    val processedItems = mutable.Set[Long]()
+    val processedItems = mutable.Set[String]()
```
wjoel: `processedIds` with this change.

src/main/scala/cognite/spark/SdkV1Rdd.scala
```diff
@@ -79,7 +79,7 @@ case class SdkV1Rdd[A](
   def enqueueStreamResults(
       stream: Stream[IO, A],
       queue: EitherQueue,
-      processedItems: mutable.Set[Long],
+      processedItems: mutable.Set[String],
```
wjoel: Do we have to pass this? It's already in scope and it's a mutable, so passing the reference doesn't seem necessary.

src/main/scala/cognite/spark/StringDataPointsRelation.scala
```diff
@@ -26,6 +26,8 @@ case class StringDataPointsFilter(
 class StringDataPointsRelationV1(config: RelationConfig)(override val sqlContext: SQLContext)
     extends DataPointsRelationV1[StringDataPointsItem](config)(sqlContext) {
 
+  import PushdownUtilities.filtersToTimestampLimits
```
wjoel: Unless we need implicits or something this should be move to the top alongside other imports.

src/main/scala/cognite/spark/ThreeDModelRevisionMappingsRelation.scala
```diff
@@ -15,7 +15,7 @@ class ThreeDModelRevisionMappingsRelation(config: RelationConfig, modelId: Long,
 
   override def toRow(t: ThreeDAssetMapping): Row = asRow(t)
 
-  override def uniqueId(a: ThreeDAssetMapping): Long = a.nodeId + a.assetId
+  override def uniqueId(a: ThreeDAssetMapping): String = a.nodeId.toString + a.assetId.toString
```
wjoel: If we decide to keep it as a `Long`: This seems possibly not very unique, we should do `a.nodeId << 32 + a.assetId` or something... which means we may be better off just going to string in this case like you do here.

src/test/scala/cognite/spark/SparkTest.scala
```diff
@@ -51,7 +51,7 @@ trait SparkTest extends CdpConnector {
         actionValue
       },
       Constants.DefaultInitialRetryDelay,
-      Constants.DefaultMaxRetries
+      Constants.DefaultMaxRetries*2
```
wjoel: ? :)

src/main/scala/cognite/spark/SdkV1Rdd.scala
hakontro: For sure! Wanted to avoid another Generic for `SdkV1Relation` and `SdkV1Rdd`, but seems like it's better than the alternative.

src/main/scala/cognite/spark/SdkV1Rdd.scala
hakontro: Seems like we do, at least according to IntelliJ. `enqueueStreamResults` is under `SdkV1Rdd`, not compute, to avoid the limit of methods > 50 lines.

src/main/scala/cognite/spark/StringDataPointsRelation.scala
hakontro: Makes sense!

src/main/scala/cognite/spark/SdkV1Rdd.scala
hakontro: Done.

src/main/scala/cognite/spark/ThreeDModelRevisionMappingsRelation.scala
hakontro: True, true. I kept is as a String, doubt this endpoint will ever hold a lot of results anyways(?).

src/test/scala/cognite/spark/SparkTest.scala
hakontro: Seeing quite a lot of retry until failure, so wanted to try out increasing the timeout. Didn't seem to help, I guess the retries until failure were because of errors related to EC giving the wrong results.

scalastyle-config.xml
```diff
@@ -49,11 +49,11 @@
...
- <check level="warning" class="org.scalastyle.file.RegexChecker" enabled="true">
-  <parameters>
-   <parameter name="regex"><![CDATA[println]]></parameter>
-  </parameters>
- </check>
+<!-- <check level="warning" class="org.scalastyle.file.RegexChecker" enabled="true">-->
+<!--  <parameters>-->
+<!--   <parameter name="regex"><![CDATA[println]]></parameter>-->
+<!--  </parameters>-->
+<!-- </check>-->
```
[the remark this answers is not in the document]
hakontro: This was just for debugging the dates issue, I've commented it in again now :)

src/main/scala/cognite/spark/SdkV1Rdd.scala
```diff
@@ -42,11 +42,11 @@ case class SdkV1Rdd[A](
     val drainPool =
       ExecutionContext.fromExecutor(Executors.newFixedThreadPool(1))
 
-    // Do not increase number of threads, as this will make processedItems non-blocking
+    // Do not increase number of threads, as this will make processedIds non-blocking
     implicit val singleThreadedCs: ContextShift[IO] =
       IO.contextShift(ExecutionContext.fromExecutor(Executors.newFixedThreadPool(1)))
 
-    val processedItems = mutable.Set[Long]()
+    val processedIds = mutable.Set[I]()
```
wjoel: `mutable.Set.empty[I]` is possibly more idiomatic but this is fine as well.

src/main/scala/cognite/spark/ThreeDModelRevisionMappingsRelation.scala
wjoel: If it's automated it could end up being a lot more, but that may never happen and if it does we'll take care of it then.

PR #323, opened 2019-10-25
Reintroduce support for setting the baseUrl

hakontro:


src/main/scala/cognite/spark/v1/SdkV1Relation.scala
```diff
@@ -27,7 +27,8 @@ abstract class SdkV1Relation[A <: Product, I](config: RelationConfig, shortName:
 
   import CdpConnector.sttpBackend
   implicit val auth: Auth = config.auth
-  @transient lazy val client = new GenericClient[IO, Nothing](Constants.SparkDatasourceVersion)
+  @transient lazy val client =
```
wjoel: We should have tests for this, I think we used to have some...?
A simple test of retrieving the first N items of some resource from Greenfield, and doing the same for non-Greenfield, and ensuring they don't have the same IDs, would suffice.

PR #324, opened 2019-10-29
Reduce default maximum timestamp

hakontro:
The default for the upper timestamp filter on datapoints was too high,
this is now reduced to year 2100.


src/main/scala/cognite/spark/v1/PushdownUtilities.scala
```diff
@@ -109,7 +109,7 @@ object PushdownUtilities {
         .getOrElse(Instant.ofEpochMilli(0)),
       Try(timestampLimits.filter(_.isInstanceOf[Max]).min).toOption
         .map(_.value)
-        .getOrElse(Instant.ofEpochMilli(Long.MaxValue))
+        .getOrElse(Instant.ofEpochMilli(4102448400000L)) // Year 2100 should be sufficient
```
wjoel: Move this to `Constants`?

PR #328, opened 2019-11-01
Databricks fix

hakontro:


hakontro:
This did not work unfortunately.

PR #331, opened 2019-11-06
Release 1.1.0, rename to "cdf-spark-datasource"

wjoel:


build.sbt
```diff
@@ -78,7 +78,7 @@ lazy val library = (project in file("."))
   .dependsOn(macroSub % "compile-internal, test-internal")
   .settings(
     commonSettings,
-    name := "cdp-spark-datasource",
+    name := "cdf-spark-datasource",
```
hakontro: Could also add this as a breaking change, I suppose it will break builds that reference the cdp-spark-datasource.

wjoel: They will continue working with the old version, but you're right, it should be in the release notes.

PR #332, opened 2019-11-06
Read time series by partitions

hakontro:
Changes to datapoints are because of the new SDK version.


src/main/scala/cognite/spark/v1/StringDataPointsRelation.scala
```diff
@@ -86,14 +86,16 @@ class StringDataPointsRelationV1(config: RelationConfig)(override val sqlContext
       (
         StringDataPointsFilter(Some(id), None),
         client.dataPoints
-          .queryStringsById(id, lowerTimeLimit, upperTimeLimit, config.limit))
+          .queryStringsById(id, lowerTimeLimit, upperTimeLimit, config.limit)
```
wjoel: Might be nicer with a for comprehension?

src/test/scala/cognite/spark/v1/TimeSeriesRelationTest.scala
```diff
@@ -34,6 +44,7 @@ class TimeSeriesRelationTest extends FlatSpec with Matchers with SparkTest {
       .option("type", "timeseries")
       .option("collectMetrics", "true")
       .option("metricsPrefix", metricsPrefix)
+      .option("partitions", "1")
```
wjoel: Shouldn't it work without this as well?

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
```diff
@@ -132,7 +132,9 @@ class AssetsRelationTest extends FlatSpec with Matchers with SparkTest {
         |'$assetsTestSource' as source,
         |10 as id,
         |0 as createdTime,
-        |0 as lastUpdatedTime
+        |0 as lastUpdatedTime,
```
wjoel: Shouldn't `null` be allowed for this, as well as for `rootId`?

src/test/scala/cognite/spark/v1/TimeSeriesRelationTest.scala
```diff
@@ -26,6 +26,16 @@ class TimeSeriesRelationTest extends FlatSpec with Matchers with SparkTest {
 
   val testDataUnit = "time-series-test-data"
 
+  it should "read all data regardless of the number of partitions" taggedAs ReadTest in {
+    val dfreader = spark.read.format("cognite.spark.v1")
+      .option("apiKey", readApiKey.apiKey)
+      .option("type", "timeseries")
+    for (np <- Seq(1, 4, 8, 12)){
+      val df = dfreader.option("partitions", np.toString).load()
```
wjoel: Nice.

src/test/scala/cognite/spark/v1/TimeSeriesRelationTest.scala
hakontro: Absolutely, this is to make sure it works _with_ it!

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
hakontro: It should be, but it isn't now because it's using the sdk case class as schema (updateTimes aren't Options).

wjoel: Can we fix that? Inserts should be based on the `Create` classes, fields that are in the read class but not in the create class should be made optional. Maybe a separate PR though.

src/main/scala/cognite/spark/v1/StringDataPointsRelation.scala
hakontro: Couldn't find a nicer one, unless you meant the `ids.map{ id` part, but that's pretty much the same. Gets tricky with `IO`, do you have a suggestion?

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
hakontro: Absolutely. I agree, let's do it in a different PR, and do it for all the resource types.

andrews47288: Emil was helping me use the `Create` classes for this earlier, still working on it then will make pr

src/main/scala/cognite/spark/v1/StringDataPointsRelation.scala
wjoel: Not really nicer, you can keep it as is.
```
val itemsIOFromId = for {
      id <- ids
    } yield (
      StringDataPointsFilter(Some(id), None),
      for {
        queryResults1 <- client.dataPoints.queryStringsById(id, lowerTimeLimit, upperTimeLimit, config.limit)
      } yield queryResults1.flatMap(_.datapoints)
    )
```

src/main/scala/cognite/spark/v1/NumericDataPointsRelation.scala
```diff
@@ -172,7 +175,10 @@ class NumericDataPointsRelationV1(config: RelationConfig)(sqlContext: SQLContext
             granularity,
             agg,
             config.limit
-          ))
+          )
+          .map(_.map {
```
wjoel: Give the parameter a name here, it's difficult to read with more than one `_`

PR #334, opened 2019-11-07
Change upsert schemas for events, time series, asset, (string) datapoints

andrews47288:


src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -88,22 +89,22 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
     val (assetsToUpdate, assetsToCreate) = assets.partition(
       p => if (p.externalId.isEmpty) { false } else { existingExternalIds.contains(p.externalId.get) }
     )
 
     val idMap = client.assets
       .retrieveByExternalIds(existingExternalIds)
-      .unsafeRunSync()
-      .map(a => a.externalId -> a.id)
-      .toMap
+      .map(_.map(e => e.externalId -> e.id).toMap)
```
hakontro: We should probably use a for Asset here.

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -88,22 +89,22 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
 
     val create =
-      if (assetsToCreate.isEmpty) IO.unit else client.assets.createFromRead(assetsToCreate)
+      if (assetsToCreate.isEmpty) IO.unit else client.assets.create(assetsToCreate)
```
hakontro: I suggest `{ IO.unit }` for consistency.

src/main/scala/cognite/spark/v1/EventsRelation.scala
```diff
@@ -2,15 +2,15 @@ package cognite.spark.v1
 
 import cats.effect.{ContextShift, IO}
 import cats.implicits._
-import com.cognite.sdk.scala.v1.{Event, EventUpdate, EventsFilter, GenericClient, TimeRange}
+import com.cognite.sdk.scala.v1.{Event, EventCreate, EventUpdate, EventsFilter, GenericClient, TimeRange}
```
hakontro: Looks like `TimeRange` isn't used anymore. Not related to this PR, but let's remove it anyways.

src/test/scala/cognite/spark/v1/EventsRelationTest.scala
```diff
@@ -487,6 +487,69 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
     spark.sparkContext.setLogLevel("WARN")
   }
 
+  it should "allow upsert in savemode" taggedAs WriteTest in {
+    val source = "spark-events-test"
+
+    // Cleanup events
```
hakontro: These comments look like they should be deleted or uncommented.

src/test/scala/cognite/spark/v1/EventsRelationTest.scala
```diff
@@ -487,6 +487,69 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
...
+              |lastUpdatedTime
+              |from sourceEvent
+              |limit 100
+   """.stripMargin)
+      .write
+      .format("cognite.spark.v1")
+      .option("apiKey", writeApiKey.apiKey)
+      .option("onconflict", "upsert")
+      .option("type", "events")
+      .save()
```
hakontro: Should probably check if the result of the upsert is as expected,

src/test/scala/cognite/spark/v1/EventsRelationTest.scala
andrews47288: these comments are copied over from the  previous test (it should support upserts), so should I remove them there as well?

src/main/scala/cognite/spark/v1/AssetsRelation.scala
andrews47288: I think Emil is going to add updating by external ID in the scala sdk, so then we can just use external ID without needing to do this at all, right?

hakontro: Yep!

hakontro:
Looks great now, feel free to rebase and merge.

src/main/scala/cognite/spark/v1/StringDataPointsRelation.scala
```diff
@@ -103,3 +111,7 @@ class StringDataPointsRelationV1(config: RelationConfig)(override val sqlContext
     Row.fromSeq(indicesOfRequiredFields.map(idx => rowOfAllFields.get(idx)))
   }
 }
+
+object StringDataPointsRelation extends UpsertSchema {
+  val upsertSchema = structType[InsertStringDataPointsItem]
```
wjoel: Wouldn't `structType[StringDataPointsItem]` produce the same schema?

src/main/scala/cognite/spark/v1/TimeSeriesRelation.scala
```diff
@@ -47,13 +47,12 @@ class TimeSeriesRelation(config: RelationConfig, useLegacyName: Boolean)(val sql
 
   override def getFromRowAndCreate(rows: Seq[Row]): IO[Unit] = {
     val timeSeriesSeq = rows.map { r =>
-      val timeSeries = fromRow[TimeSeries](r)
-      val timeSeriesWithMetadata = timeSeries.copy(metadata = filterMetadata(timeSeries.metadata))
-      val timeSeriesCreate = timeSeriesWithMetadata.transformInto[TimeSeriesCreate]
+      val timeSeries = fromRow[TimeSeriesCreate](r)
+      timeSeries.copy(metadata = filterMetadata(timeSeries.metadata))
```
wjoel: `.copy` does not modify the item in place, it will return a new `TimeSeriesCreate` item with the modified metadata. You need to keep `timeSeriesWithMetadata` as it was.

We should have a test for this if we didn't catch it in one of the existing ones.

andrews47288: Test was correctly failing with this change, I just missed it. I changed it back to the way it was and the test is passing

src/main/scala/cognite/spark/v1/StringDataPointsRelation.scala
```diff
@@ -24,6 +27,13 @@ case class StringDataPointsFilter(
     externalId: Option[String]
 )
 
+case class InsertStringDataPointsItem(
```
wjoel: Probably don't need this, results in the same schema as `StringDataPointsItem` and if we were reading back `java.sql.Timestamp` from the row we should probably continue to use `StringDataPointsItem` on line 53?

src/test/scala/cognite/spark/v1/EventsRelationTest.scala
```diff
@@ -483,6 +483,84 @@ class EventsRelationTest extends FlatSpec with Matchers with SparkTest {
...
+              |least(startTime, endTime) as startTime,
+              |greatest(startTime, endTime) as endTime,
+              |type,
+              |subtype,
+              |null as assetIds,
+              |bigint(0) as id,
+              |map("foo", null, "bar", "test") as metadata,
+              |"$source" as source,
+              |string(id) as externalId,
+              |createdTime,
```
wjoel: Test with `null as createdTime` since it should be optional (and ignored) now.

src/test/scala/cognite/spark/v1/TimeSeriesRelationTest.scala
```diff
@@ -280,6 +280,8 @@ class TimeSeriesRelationTest extends FlatSpec with Matchers with SparkTest {
     assert(initialDescriptionsAfterUpsert.length == 0)
   }
 
+
```
wjoel: Remove added newlines.

PR #335, opened 2019-11-07
Save mode for numeric and string datapoints

andrews47288:
-- datapoints and stringdatapoints only support upserts (deleted the part about deletes bc we are not sure how it will look in jetfire for now)

### DIFF
--- a/src/main/scala/cognite/spark/v1/DefaultSource.scala
+++ b/src/main/scala/cognite/spark/v1/DefaultSource.scala
@@ -180,7 +180,6 @@ class DefaultSource
       data: DataFrame): BaseRelation = {
     val config = parseRelationConfig(parameters, sqlContext)
     val resourceType = parameters.getOrElse("type", sys.error("Resource type must be specified"))
-
     val relation = resourceType match {
       case "events" =>
         new EventsRelation(config)(sqlContext)
@@ -189,6 +188,10 @@ class DefaultSource
         new TimeSeriesRelation(config, useLegacyName)(sqlContext)
       case "assets" =>
         new AssetsRelation(config)(sqlContext)
+      case "datapoints" =>
+        new NumericDataPointsRelationV1(config)(sqlContext)
+      case "stringdatapoints" =>
+        new StringDataPointsRelationV1(config)(sqlContext)
       case _ => sys.error(s"Resource type $resourceType does not support save()")
     }
 
--- a/src/main/scala/cognite/spark/v1/NumericDataPointsRelation.scala
+++ b/src/main/scala/cognite/spark/v1/NumericDataPointsRelation.scala
@@ -102,9 +102,21 @@ object Granularity {
 }
 
 class NumericDataPointsRelationV1(config: RelationConfig)(sqlContext: SQLContext)
-    extends DataPointsRelationV1[DataPointsItem](config)(sqlContext) {
+    extends DataPointsRelationV1[DataPointsItem](config)(sqlContext)
+    with WritableRelation {
 
   import PushdownUtilities.filtersToTimestampLimits
+
+  override def insert(rows: Seq[Row]): IO[Unit] =
+    throw new RuntimeException("Insert not supported for datapoints. Use upsert instead.")
+
+  override def upsert(rows: Seq[Row]): IO[Unit] = insertSeqOfRows(rows)
+
+  override def update(rows: Seq[Row]): IO[Unit] =
+    throw new RuntimeException("Update not supported for datapoints. Use upsert instead.")
+
+  override def delete(rows: Seq[Row]): IO[Unit] =
+    throw new RuntimeException("Delete not supported for datapoints.")
 
   override def schema: StructType =
     StructType(
--- a/src/main/scala/cognite/spark/v1/SdkV1Relation.scala
+++ b/src/main/scala/cognite/spark/v1/SdkV1Relation.scala
@@ -88,20 +88,11 @@ abstract class SdkV1Relation[A <: Product, I](config: RelationConfig, shortName:
       val rowOfAllFields = toRow(item)
       Row.fromSeq(indicesOfRequiredFields.map(idx => rowOfAllFields.get(idx)))
     }
+}
 
-  def insert(rows: Seq[Row]): IO[Unit] =
-    throw new IllegalArgumentException(
-      s"""$shortName does not support the "onconflict" option "abort".""")
-
-  def upsert(rows: Seq[Row]): IO[Unit] =
-    throw new IllegalArgumentException(
-      s"""$shortName does not support the "onconflict" option "upsert".""")
-
-  def update(rows: Seq[Row]): IO[Unit] =
-    throw new IllegalArgumentException(
-      s"""$shortName does not support the "onconflict" option "update".""")
-
-  def delete(rows: Seq[Row]): IO[Unit] =
-    throw new IllegalArgumentException(
-      s"""$shortName does not support the "onconflict" option "delete".""")
+trait WritableRelation {
+  def insert(rows: Seq[Row]): IO[Unit]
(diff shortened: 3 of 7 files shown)


src/main/scala/cognite/spark/v1/NumericDataPointsRelation.scala
```diff
@@ -36,11 +36,44 @@ case class InsertDataPointsItem(
...
+    extends DataPointsRelationV1[DataPointsItem](config)(sqlContext)
+    with WritableRelation {
 
   import PushdownUtilities.filtersToTimestampLimits
 
+  override def insert(rows: Seq[Row]): IO[Unit] = insertSeqOfRows(rows)
+
+  override def upsert(rows: Seq[Row]): IO[Unit] = insert(rows)
+
+  override def update(rows: Seq[Row]): IO[Unit] = insert(rows)
```
hakontro: Please throw an error on `insert` and `update` that says they're not supported for datapoints, and advise using upsert instead. The way we describe these in the Readme are that insert will stop upon finding a match, while update will stop when not finding a match, and I'd prefer making it clear that we don't do that for datapoints.

src/main/scala/cognite/spark/v1/StringDataPointsRelation.scala
```diff
@@ -27,8 +29,41 @@ case class StringDataPointsFilter(
...
 class StringDataPointsRelationV1(config: RelationConfig)(override val sqlContext: SQLContext)
-    extends DataPointsRelationV1[StringDataPointsItem](config)(sqlContext) {
+    extends DataPointsRelationV1[StringDataPointsItem](config)(sqlContext)
+    with WritableRelation {
+
+  override def insert(rows: Seq[Row]): IO[Unit] = insertSeqOfRows(rows)
+
+  override def upsert(rows: Seq[Row]): IO[Unit] = insert(rows)
+
+  override def update(rows: Seq[Row]): IO[Unit] = insert(rows)
```
hakontro: Please throw an error on `insert` and `update` that says they're not supported for datapoints, and advise using upsert instead. The way we describe these in the Readme are that insert will stop upon finding a match, while update will stop when not finding a match, and I'd prefer making it clear that we don't do that for datapoints.

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -193,6 +193,10 @@ class DefaultSource
         new TimeSeriesRelation(config)(sqlContext)
       case "assets" =>
         new AssetsRelation(config)(sqlContext)
+      case "datapoints" =>
+        new NumericDataPointsRelationV1(config)(sqlContext)
+      case "string_datapoints" =>
```
wjoel: We call this `stringdatapoints`. Since we're using those strings in multiple places (and to avoid errors like this) we should probably move them to Constants and define them once instead.

src/main/scala/cognite/spark/v1/StringDataPointsRelation.scala
```diff
@@ -27,8 +29,41 @@ case class StringDataPointsFilter(
     externalId: Option[String]
 )
 
+case class StringDataPointsDelete(
```
wjoel: This looks identical to `NumericDataPointsDelete` so we can probably use just one `DataPointsDelete`.

Very cool that it works.

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
```diff
@@ -88,20 +88,11 @@ abstract class SdkV1Relation[A <: Product, I](config: RelationConfig, shortName:
...
-      s"""$shortName does not support the "onconflict" option "upsert".""")
-
-  def update(rows: Seq[Row]): IO[Unit] =
-    throw new IllegalArgumentException(
-      s"""$shortName does not support the "onconflict" option "update".""")
-
-  def delete(rows: Seq[Row]): IO[Unit] =
-    throw new IllegalArgumentException(
-      s"""$shortName does not support the "onconflict" option "delete".""")
+trait WritableRelation {
```
wjoel: Why are we moving this to a trait? The default methods seemed reasonable, what will happen instead now that some of them aren't implemented at all?

src/main/scala/cognite/spark/v1/NumericDataPointsRelation.scala
```diff
@@ -36,11 +36,48 @@ case class InsertDataPointsItem(
...
+
+  override def upsert(rows: Seq[Row]): IO[Unit] = insertSeqOfRows(rows)
+
+  override def update(rows: Seq[Row]): IO[Unit] = {
+    throw new RuntimeException("Update not supported for datapoints. Use upsert instead.")
+  }
+
+  override def delete(rows: Seq[Row]): IO[Unit] = {
+    val (dataPointsWithId, dataPointsWithExternalId) =
+      rows.map(r => fromRow[DataPointsDelete](r)).partition(p => p.id.isDefined)
```
wjoel: I'm a bit confused by this. What will this look like for the end user? How is this specified?

wjoel: I would have thought we'd need to get start and end from some filter.

andrews47288: The user could do a query of the form "select id, externalId, inclusiveBegin, exclusiveEnd from table where id = timeSeriesId" and somehow specify a delete operation on the front-end

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
andrews47288: If they are not implemented for a specific resource type, then those resources will throw their own exception. Moved it into a trait just to make default source cleaner and since each resource will ultimately implement the 4 operations

src/main/scala/cognite/spark/v1/NumericDataPointsRelation.scala
andrews47288: The user can do a query specifying the time series id and / or external id, start and end of the time range to delete, and the table to delete from on some condition. When supported on the front end they could select delete as the operation and then just run the transform

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
wjoel: But this default implementation takes care of throwing an appropriate exception? Isn't this cleaner than each resource implementing its own "throw new IllegalArgumentException" for operations it doesn't support?

src/main/scala/cognite/spark/v1/NumericDataPointsRelation.scala
wjoel: That's a strange and non-intuitive way of handling this. It's selecting column names that don't exist in the table, right? While technically possible and allowed by Spark (I guess? I find even that surprising, but I guess it doesn't check the schema at all) this is not how we should expose it, I think.

Something like "select idToDelete from datapoints where timestamp >= deleteFromHere and timestamp <= deleteUntilHere" would be more natural.

I don't know how easy that would be to implement, but I think we should perhaps leave deletes out of this for now.

wjoel: @Tapped any thoughts on what this would look like in Jetfire?

andrews47288: removing delete for datapoints and stringdatapoints until we support deleting in jetfire. Need to come up with a good way to represent it.

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
andrews47288: We decided to keep it in a trait bc datapoints and stringdatapoints are the only two that don't implement all 4 operations

PR #336, opened 2019-11-11
Avoid global execution context

wjoel:
Spark uses the global execution context for a lot of things, and depleting it gives some horrendous error messages and breaks the reporting of the original exception back to the driver.

So don't use it.

### DIFF
--- a/src/main/scala/cognite/spark/v1/AssetsRelation.scala
+++ b/src/main/scala/cognite/spark/v1/AssetsRelation.scala
@@ -11,14 +11,12 @@ import PushdownUtilities._
 import com.cognite.sdk.scala.common.CdpApiException
 import fs2.Stream
 
-import scala.concurrent.ExecutionContext
 import io.scalaland.chimney.dsl._
 
 class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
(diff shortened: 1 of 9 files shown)


wjoel:
@Tapped do you think we should be using a cached thread pool instead for this?

hakontro:
If I understand correctly, the upside of the cached thread pool is that
>These pools will typically improve the performance of programs that execute many short-lived asynchronous tasks

with the drawback being possibly too many threads. Given the limit from `parallelismPerPartition` this may not be a problem for us? Here's some light performance testing, reading all assets from a large tenant with each pool type. Forgot to note the runtime for 10 on the first run, hence the `_`.

|Parallelism|Fixed|Cached|
|---|---|---|
|1|70s|72s|
|2|39s|40s|
|5|20s|20s|
|10|_, 18.3s|20s, 22.4s|
|20|15.9s, 17.8s|14.9s, 16.0s|
|40|18.8s, 20.7s|15.4s, 15.8s|

Not very significant, but it does seem that cached is actually faster for a large number of threads.

wjoel:
Cached thread pool is also nice in that it doesn't need some arbitrary number like `numProcessors * 2`. I'll switch to using it instead. Thanks for testing it out!

PR #337, opened 2019-11-11
Fix flaky tests

hakontro:
Having all tests write and delete data based on the same unit turned out to be a bad idea, since this increases the number of operations that happen on the same data. This PR decouples the tests, which should fix a lot of the flaky behavior we're seeing now.

### DIFF
--- a/src/main/scala/cognite/spark/v1/PushdownUtilities.scala
+++ b/src/main/scala/cognite/spark/v1/PushdownUtilities.scala
@@ -119,11 +119,11 @@ object PushdownUtilities {
       case _ => {
         val minimumTimeAsInstant =
           minTime
-            .map(java.sql.Timestamp.valueOf(_).toInstant.plusMillis(1))
+            .map(java.sql.Timestamp.valueOf(_).toInstant)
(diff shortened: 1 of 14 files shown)


src/main/scala/cognite/spark/v1/PushdownUtilities.scala
```diff
@@ -119,11 +119,11 @@ object PushdownUtilities {
       case _ => {
         val minimumTimeAsInstant =
           minTime
-            .map(java.sql.Timestamp.valueOf(_).toInstant.plusMillis(1))
+            .map(java.sql.Timestamp.valueOf(_).toInstant)
```
wjoel: We probably had a reason for the old `plusMillis`, any idea why we did that? :/

hakontro: I _think_ that's because we may initially have coded this without pushdowns for `gt/ltOrEqual`. As far as I can see the endpoints are inclusive, so this is the right way to always get the datapoints (and filter later if < or >).

PR #344, opened 2019-11-20
Decouple datasource schemas from scala sdk

andrews47288:
-- add upsert / insert / read schemas explicitly in spark datasource rather than relying on scala sdk schemas

upsert schemas have optional id


src/main/scala/cognite/spark/v1/TimeSeriesRelation.scala
```diff
@@ -1,8 +1,10 @@
 package cognite.spark.v1
 
+import java.time.Instant
+
 import cats.effect.IO
 import cats.implicits._
-import com.cognite.sdk.scala.common.CdpApiException
+import com.cognite.sdk.scala.common.{CdpApiException, NonNullableSetter, Setter}
```
hakontro: These aren't needed, I think.

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -116,5 +117,42 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
   override def uniqueId(a: Asset): Long = a.id
 }
 object AssetsRelation extends UpsertSchema {
-  val upsertSchema = structType[AssetCreate]
+  val upsertSchema = structType[assetsUpsertSchema]
+  val insertSchema = structType[assetsCreateSchema]
+  val readSchema = structType[assetsReadSchema]
 }
+
+case class assetsUpsertSchema(
```
hakontro: Please change to upper case first letter for case class names.

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
```diff
@@ -159,7 +159,7 @@ class AssetsRelationTest extends FlatSpec with Matchers with SparkTest {
               |'asset name' as name,
               |null as parentId,
               |'asset description' as description,
-              map("foo", null, "bar", "test") as metadata,
+              |map("foo", null, "bar", "test") as metadata,
```
hakontro: Strange, was this working earlier?

andrews47288: Not sure what is happening here. The test was failing when I came in this morning, so I looked at it and saw the pipe wasn't there, then I added the pipe and the test started passing. It's also missing on the master branch, although it built successfully anyway.

wjoel: Both work, the output just gets ugly without the pipe, but it's still syntactically valid SQL. Most likely a random test failure? Either way it should have the pipe, so nice catch.

```
@ val assetsTestSource = "asdf"
assetsTestSource: String = "asdf"
@ s"""
                |select "1" as externalId,
                |'asset name' as name,
                |null as parentId,
                |map("foo", null, "bar", "test") as metadata,
                |bigint(10) as id
        """.stripMargin
res1: String = """
select "1" as externalId,
'asset name' as name,
null as parentId,
'asset description' as description,
map("foo", null, "bar", "test") as metadata,
'asdf' as source,
bigint(10) as id
      """
@ s"""
                |select "1" as externalId,
                |'asset name' as name,
                |null as parentId,
                map("foo", null, "bar", "test") as metadata,
                |bigint(10) as id
        """.stripMargin
res2: String = """
select "1" as externalId,
'asset name' as name,
null as parentId,
'asset description' as description,
              map("foo", null, "bar", "test") as metadata,
'asdf' as source,
bigint(10) as id
      """
```

andrews47288:
I suppose we will have to change how we handle updates whenever there is a new release of the scala sdk, now that we changed `update(Seq[xyUpdate])` to `updateById(Map[Long, xyUpdate]` and `updateByExternalId(Map[Long, xyUpdate])`

PR #347, opened 2019-11-21
Update update logic

andrews47288:
-- Update the spark datasource to use updatebyId() and updateByExternalId() from the scala sdk
-- Test that we can update resources by specifying externalId

### DIFF
--- a/src/main/scala/cognite/spark/v1/AssetsRelation.scala
+++ b/src/main/scala/cognite/spark/v1/AssetsRelation.scala
@@ -10,7 +10,8 @@ import cognite.spark.v1.SparkSchemaHelper._
 import org.apache.spark.sql.types._
 import cats.implicits._
 import PushdownUtilities._
-import com.cognite.sdk.scala.common.CdpApiException
+import com.cognite.sdk.scala.common.{CdpApiException, WithExternalId, WithId}
+import com.cognite.sdk.scala.v1.resources.Assets
 import fs2.Stream
 import io.scalaland.chimney.dsl._
 
@@ -58,8 +59,11 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
   }
 
   override def update(rows: Seq[Row]): IO[Unit] = {
-    val assetUpdates = rows.map(r => fromRow[AssetUpdate](r))
-    client.assets.update(assetUpdates) *> IO.unit
+    val assetUpdates = rows.map(r => fromRow[AssetsUpsertSchema](r))
+    updateByIdOrExternalId[AssetsUpsertSchema, AssetUpdate, Assets[IO], Asset](
+      assetUpdates,
+      client.assets
+    )
   }
 
   override def delete(rows: Seq[Row]): IO[Unit] = {
@@ -89,26 +93,11 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
       } *> IO.unit
   }
 
-  def resolveConflict(existingExternalIds: Seq[String], assets: Seq[AssetCreate]): IO[Unit] = {
-    val (assetsToUpdate, assetsToCreate) = assets.partition(
-      p => if (p.externalId.isEmpty) { false } else { existingExternalIds.contains(p.externalId.get) }
-    )
-
-    val idMap = client.assets
-      .retrieveByExternalIds(existingExternalIds)
-      .map(_.map(e => e.externalId -> e.id).toMap)
-
-    val create =
-      if (assetsToCreate.isEmpty) { IO.unit } else client.assets.create(assetsToCreate)
-    val update =
-      if (assetsToUpdate.isEmpty) { IO.unit } else {
-        idMap.flatMap(idMap =>
-          client.assets.update(assetsToUpdate.map(a =>
-            a.into[AssetUpdate].withFieldComputed(_.id, x => idMap(x.externalId)).transform)))
-      }
-
-    (create, update).parMapN((_, _) => ())
-  }
+  def resolveConflict(existingExternalIds: Seq[String], assets: Seq[AssetCreate]): IO[Unit] =
+    upsertAfterConflict[Asset, AssetUpdate, AssetCreate, Assets[IO]](
+      existingExternalIds,
+      assets,
+      client.assets)
 
   override def schema: StructType = structType[Asset]
 
@@ -131,7 +120,8 @@ case class AssetsUpsertSchema(
     metadata: Option[Map[String, String]] = None,
     parentId: Option[Long] = None,
     parentExternalId: Option[String] = None
-)
+) extends WithExternalId
+    with WithId[Option[Long]]
 
 case class AssetsInsertSchema(
     name: String,
--- a/src/main/scala/cognite/spark/v1/EventsRelation.scala
+++ b/src/main/scala/cognite/spark/v1/EventsRelation.scala
@@ -9,8 +9,9 @@ import cognite.spark.v1.SparkSchemaHelper.{asRow, fromRow, structType}
 import org.apache.spark.sql.{Row, SQLContext}
 import org.apache.spark.sql.sources.{Filter, InsertableRelation}
 import org.apache.spark.sql.types.{DataTypes, StructType}
-import com.cognite.sdk.scala.common.CdpApiException
+import com.cognite.sdk.scala.common.{CdpApiException, WithExternalId, WithId}
 import PushdownUtilities._
+import com.cognite.sdk.scala.v1.resources.Events
 import fs2.Stream
 import io.scalaland.chimney.dsl._
 
@@ -78,8 +79,11 @@ class EventsRelation(config: RelationConfig)(val sqlContext: SQLContext)
   }
 
   override def update(rows: Seq[Row]): IO[Unit] = {
-    val events = rows.map(r => fromRow[EventUpdate](r))
-    client.events.update(events) *> IO.unit
+    val eventUpdates = rows.map(r => fromRow[EventsUpsertSchema](r))
+    updateByIdOrExternalId[EventsUpsertSchema, EventUpdate, Events[IO], Event](
+      eventUpdates,
+      client.events
+    )
   }
 
   override def delete(rows: Seq[Row]): IO[Unit] = {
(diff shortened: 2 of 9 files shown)


src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -57,8 +58,22 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
   }
 
   override def update(rows: Seq[Row]): IO[Unit] = {
-    val assetUpdates = rows.map(r => fromRow[AssetUpdate](r))
-    client.assets.update(assetUpdates) *> IO.unit
+    val assetUpdates = rows.map(r => fromRow[AssetsUpsertSchema](r))
+    val (updatesById, updatesByExternalId) = assetUpdates.partition(
+      p => p.id.isDefined
+    )
+    val id = if (updatesById.isEmpty) { IO.unit } else {
```
wjoel: `val updateIds` because this is an effect that does something when you run it.

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -57,8 +58,22 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+      client.assets.updateById(updatesById.map(a => a.id.get -> a.transformInto[AssetUpdate]).toMap)
+    }
+    val extId = if (updatesByExternalId.isEmpty) { IO.unit } else {
```
wjoel: `val updateExternalds` because this is an effect that does something when you run it.

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -57,8 +58,22 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+      client.assets.updateByExternalId(
+        updatesByExternalId
+          .map(a =>
+            a.externalId
+              .getOrElse("") -> a.into[AssetUpdate].withFieldComputed(_.externalId, _ => None).transform)
```
wjoel: I think this should be `.get`, actually: it should never happen due to the `.partition` above, and if it does happen we should crash (as `.get` does on `None`).

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -93,17 +108,18 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
       if (assetsToUpdate.isEmpty) { IO.unit } else {
-        idMap.flatMap(idMap =>
-          client.assets.update(assetsToUpdate.map(a =>
-            a.into[AssetUpdate].withFieldComputed(_.id, x => idMap(x.externalId)).transform)))
+        client.assets.updateByExternalId(
+          assetsToUpdate
+            .map(
+              a =>
+                (
+                  a.externalId.getOrElse(""),
```
wjoel: Same thing here, those should always have an `.externalId` if we do `.partition` correctly, right? So better to use `.get` and crash if they don't.

src/main/scala/cognite/spark/v1/EventsRelation.scala
```diff
@@ -76,8 +78,21 @@ class EventsRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+      p => p.id.isDefined
+    )
+    val id = if (updatesById.isEmpty) { IO.unit } else {
+      client.events.updateById(updatesById.map(ev => ev.id.get -> ev.transformInto[EventUpdate]).toMap)
+    }
+    val extId = if (updatesByExternalId.isEmpty) { IO.unit } else {
+      client.events.updateByExternalId(updatesByExternalId
+        .map(ev =>
+          ev.externalId
+            .getOrElse("") -> ev.into[EventUpdate].withFieldComputed(_.externalId, _ => None).transform)
```
wjoel: And again about `.getOrElse`. We seem to be doing this "partition, then map to `externalId -> ...into[EventUpdate].withFieldComputed(...)`" several times, so maybe we can move that into a utility function?

src/main/scala/cognite/spark/v1/EventsRelation.scala
```diff
@@ -112,17 +127,18 @@ class EventsRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
       if (eventsToUpdate.isEmpty) { IO.unit } else {
-        idMap.flatMap(idMap =>
-          client.events.update(eventsToUpdate.map(e =>
-            e.transformInto[EventUpdate].copy(id = idMap(e.externalId)))))
+        client.events.updateByExternalId(
+          eventsToUpdate
+            .map(
+              e =>
+                (
+                  e.externalId.getOrElse(""),
```
wjoel: And again :)

src/main/scala/cognite/spark/v1/TimeSeriesRelation.scala
```diff
@@ -33,8 +35,26 @@ class TimeSeriesRelation(config: RelationConfig, useLegacyName: Boolean)(val sql
...
+    val id = if (updatesById.isEmpty) { IO.unit } else {
+      client.timeSeries.updateById(
+        updatesById.map(ts => ts.id.get -> ts.transformInto[TimeSeriesUpdate]).toMap)
+    }
+    val extId = if (updatesByExternalId.isEmpty) { IO.unit } else {
+      client.timeSeries.updateByExternalId(
+        updatesByExternalId
+          .map(
+            ts =>
+              ts.externalId.getOrElse("") -> ts
```
wjoel: And again

src/main/scala/cognite/spark/v1/TimeSeriesRelation.scala
```diff
@@ -80,19 +100,20 @@ class TimeSeriesRelation(config: RelationConfig, useLegacyName: Boolean)(val sql
...
       if (timeSeriesToUpdate.isEmpty) { IO.unit } else {
-        idMap.flatMap(idMap =>
-          client.timeSeries.update(timeSeriesToUpdate.map(ts =>
-            ts.transformInto[TimeSeriesUpdate].copy(id = idMap(ts.externalId)))))
+        client.timeSeries.updateByExternalId(
+          timeSeriesToUpdate
+            .map(
+              ts =>
+                (
+                  ts.externalId.getOrElse(""),
```
wjoel: And again

src/main/scala/cognite/spark/v1/AssetsRelation.scala
andrews47288: my thought was that we partition on if id is non-null, so theoretically I guess externalId could be null here. In updateByExternalId() in the scala sdk, we require that the externalId is > "", so it ends up giving the error message "Updating by externalId requires externalId to be set".

andrews47288: but if crashing here is more helpful, then can definitely change that easily

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -58,8 +59,23 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+      p => p.id.isDefined
+    )
+    val updateIds = if (updatesById.isEmpty) { IO.unit } else {
+      client.assets.updateById(updatesById.map(a => a.id.get -> a.transformInto[AssetUpdate]).toMap)
+    }
+    val updateExternalIds = if (updatesByExternalId.isEmpty) { IO.unit } else {
+      updateByExternalIdHandler[AssetUpdate, AssetsUpsertSchema, Assets[IO]](
+        updatesByExternalId,
+        client.assets)
+
```
wjoel: Remove this empty line.

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
```diff
@@ -88,6 +90,17 @@ abstract class SdkV1Relation[A <: Product, I](config: RelationConfig, shortName:
       val rowOfAllFields = toRow(item)
       Row.fromSeq(indicesOfRequiredFields.map(idx => rowOfAllFields.get(idx)))
     }
+
+  // scalastyle:off no.whitespace.after.left.bracket
+  def updateByExternalIdHandler[
```
wjoel: `updateByExternalId` is better, that's what it does and the name of the method it's using. I think it's clearer. "handler" is usually used for callbacks, and this isn't that.

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
```diff
@@ -88,6 +90,17 @@ abstract class SdkV1Relation[A <: Product, I](config: RelationConfig, shortName:
...
+      T <: WithSetExternalId,
+      R <: WithExternalId,
+      U <: UpdateByExternalId[A, T, IO]](updates: Seq[R], resource: U)(
```
wjoel: I'm confused by the names :)
`R` should be the "read" type, for example `Asset`, returned from listing.
`U` should be the "update" type, for example `AssetUpdate`.
`C` (which we don't have here) is the "create" type, `AssetCreate`.

Can we find better names? May even have to use longer names to make it clear, even if that can also be confusing, for example `Resource <: UpdateByExternalId[...]`. However, couldn't we just do `resource: UpdateByExternalId[...]`?

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -58,8 +59,23 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
   }
 
   override def update(rows: Seq[Row]): IO[Unit] = {
-    val assetUpdates = rows.map(r => fromRow[AssetUpdate](r))
-    client.assets.update(assetUpdates) *> IO.unit
+    val assetUpdates = rows.map(r => fromRow[AssetsUpsertSchema](r))
+    require(
+      assetUpdates.forall(au => au.id.isDefined || au.externalId.isDefined),
+      "Updating requires id or externalId to be set for each row.")
+    val (updatesById, updatesByExternalId) = assetUpdates.partition(
```
wjoel: `assetUpdates.partition(_.id.isDefined)` looks better and is clear enough.

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
wjoel: Even better would be `updateByIdOrExternalId` if we can abstract out the other common things:
Ensuring id or externalId is set, partitioning based on this, updating by id or external id, doing the parMapN.

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
andrews47288: I tried doing `resource: UpdateByExternalId[...]` without having U already as a parametrization of the method, but it was not happy with that. I am not sure if it is possible to define a generic in the parameter list?

andrews47288: I have changed it to be that U is xyUpdate, C is xyCreate or xyUpsertSchema, and T is the resource type

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
andrews47288: In `resolveConflict` we partition on `if (p.externalId.isEmpty) { false } else { existingExternalIds.contains(p.externalId.get)` and not `p.id.isDefined`. Also, the updates can either be a sequence of xyUpsertSchema, which do have id, or xyCreate, which don't. So I am not sure if there is a way to unify these, although I could factor them into two different methods if we want (one for update() and one for resolveConflict())

PR #349, opened 2019-11-28
Support ignoreUnknownIds for asset and event deletes

andrews47288:
- support ignoreUnknownIds when doing delete
- user can specify `true as ignoreUnknownIds` in delete queries to not error out when encountering unknown ids
- `false as ignoreUnknownIds` or not specifying any value for ignoreUnknownIds will error out when encountering unknown ids

### DIFF
--- a/src/main/scala/cognite/spark/v1/AssetsRelation.scala
+++ b/src/main/scala/cognite/spark/v1/AssetsRelation.scala
@@ -67,8 +67,8 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
   }
 
   override def delete(rows: Seq[Row]): IO[Unit] = {
-    val ids = rows.map(r => fromRow[DeleteItem](r).id)
-    client.assets.deleteByIds(ids)
+    val deletes = rows.map(r => fromRow[DeleteItem](r))
+    deleteWithIgnoreUnknownIds(client.assets, deletes, config.ignoreUnknownIds)
   }
 
   override def upsert(rows: Seq[Row]): IO[Unit] = getFromRowAndCreate(rows)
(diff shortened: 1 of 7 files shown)


src/main/scala/cognite/spark/v1/PushdownUtilities.scala
```diff
@@ -8,7 +8,7 @@ import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types.{DataTypes, StructField, StructType}
 import scala.util.Try
 
-case class DeleteItem(id: Long)
+case class DeleteItem(id: Long, ignoreUnknownIds: Option[Boolean] = None)
```
wjoel: I don't think we need to bother with making this optional... can't think of any use case for when we'd like deletes on missing items to give errors, and the code will be so much simpler if we just always use this flag in the Spark data source.

wjoel: We can support it as an option on the relation, `.option("ignoreUnknownIds", "false")` (with a default of true).

andrews47288: why should the default be true? In the scala sdk it defaults to false, plus that was the behavior before this was added.

wjoel: In the SDK(s) we try to stay as close to the API itself as possible, where ignoreUnknownIds is false by default.

For users of the Spark data source we expect that almost always they'd rather not have their job fail if something they're trying to delete has already been deleted. We're adding this feature to the Spark data source *because* users have been disappointed when deletes fail because something has already been deleted. ;)

wjoel:
The failing tests are a good example of how annoying it is to have deletes fail when something has already been deleted. Happens quite often due to eventual consistency, where you read back data you want to delete but that data may be stale and partly deleted already, and then you do a delete on the ids you read.

PR #354, opened 2019-12-11
Set CDF base URL in StringDataPointsRdd

sebbert:


sebbert:
Already fixed in #353

PR #355, opened 2019-12-16
[DT-1239] Add metrics for insert, update, and support for upsert by id

wjoel:


wjoel:
I'll make some additional cleanup and add tests for other resource types as well, but this is ready for a first review.

src/main/scala/cognite/spark/v1/EventsRelation.scala
```diff
@@ -101,23 +101,12 @@ class EventsRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
-          if (e.code == 409) {
-            val existingExternalIds =
-              e.duplicated.get.map(j => j("externalId").get.asString.get)
-            resolveConflict(existingExternalIds, events)
-          } else { IO.raiseError(e) }
-      } *> IO.unit
-  }
-
-  def resolveConflict(existingExternalIds: Seq[String], events: Seq[EventCreate]): IO[Unit] =
     upsertAfterConflict[Event, EventUpdate, EventCreate, Events[IO]](
```
sebbert: I don't think the name `upsertAfterConflict` makes much sense anymore, since now seems to be the primary entrypoint for doing upserts?

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
```diff
@@ -123,14 +116,37 @@ abstract class SdkV1Relation[A <: Product, I](config: RelationConfig, shortName:
...
+          IO {
+            if (config.collectMetrics) {
+              itemsCreated.inc(resourcesToCreate.size)
+            }
+          }
+        }
+        .handleErrorWith {
+          case e: CdpApiException if doUpsert && e.code == 409 =>
+            val existingExternalIds =
+              e.duplicated.get.map(j => j("externalId").get.asString.get)
```
hakontro: Could this be improved by parsing the Json directly to a case class, either here or in the SDK?

project/plugins.sbt
```diff
@@ -5,4 +5,5 @@ addSbtPlugin("com.geirsson" % "sbt-scalafmt" % "1.5.1")
 addSbtPlugin("com.thesamet" % "sbt-protoc" % "0.99.19")
 addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.1.1")
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.9.0")
+addSbtPlugin("au.com.onegeek" %% "sbt-dotenv" % "2.1.146")
```
hakontro: Nice! I suppose it should be mentioned as an option in the Readme.

src/main/scala/cognite/spark/v1/TimeSeriesRelation.scala
```diff
@@ -77,7 +77,8 @@ class TimeSeriesRelation(config: RelationConfig, useLegacyName: Boolean)(val sql
     upsertAfterConflict[TimeSeries, TimeSeriesUpdate, TimeSeriesCreate, TimeSeriesResource[IO]](
       existingExternalIds,
```
hakontro: It looks to me like this could have the same implementation as Assets and Events here, where the `externalIds` are handled in `upsertAfterConflict`, but I may be wrong. If it can, we can split `upsertAfterConflict` in two and avoid the recursion.

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -80,24 +80,12 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
-  }
-
-  def resolveConflict(existingExternalIds: Seq[String], assets: Seq[AssetCreate]): IO[Unit] =
     upsertAfterConflict[Asset, AssetUpdate, AssetCreate, Assets[IO]](
-      existingExternalIds,
+      Seq.empty,
       assets,
-      client.assets)
+      client.assets,
+      doUpsert = true)
```
sebbert: Perhaps we should give a default value to this parameter, since it's only set to `false` when handling duplicates?

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
```diff
@@ -123,14 +116,37 @@ abstract class SdkV1Relation[A <: Product, I](config: RelationConfig, shortName:
...
+              doUpsert = false)
+        }
     }
-    val update = if (resourcesToUpdate.isEmpty) { IO.unit } else {
+    val update = if (resourcesToUpdate.isEmpty) {
+      IO.unit
+    } else {
       resource.updateByExternalId(
         resourcesToUpdate
           .map(u => u.externalId.get -> u.into[U].withFieldComputed(_.externalId, _ => None).transform)
```
sebbert: Shouldn't we be reporting metrics for updates here? I thought that was one of the main points of this PR?

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
```diff
@@ -104,37 +106,72 @@ abstract class SdkV1Relation[A <: Product, I](config: RelationConfig, shortName:
...
+      IO.unit
+    } else {
+      resource
+        .create(resourcesToCreate)
+        .flatTap(_ => incMetrics(itemsCreated, resourcesToCreate.size))
+        .handleErrorWith {
+          case e: CdpApiException if doUpsert && e.code == 409 =>
+            assertNoLegacyNameConflicts(e)
+            val existingExternalIds =
+              e.duplicated.get.flatMap(j => j("externalId")).map(_.asString.get)
```
sebbert: Are we sure that `duplicated` is always set here? According to the docs, it can be null, so we might want to map its absence to an empty list instead

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
```diff
@@ -104,37 +106,72 @@ abstract class SdkV1Relation[A <: Product, I](config: RelationConfig, shortName:
...
     )
-    val create = if (resourcesToCreate.isEmpty) { IO.unit } else {
-      resource.create(resourcesToCreate)
+    val create = if (resourcesToCreate.isEmpty) {
+      IO.unit
+    } else {
+      resource
+        .create(resourcesToCreate)
+        .flatTap(_ => incMetrics(itemsCreated, resourcesToCreate.size))
+        .handleErrorWith {
```
sebbert: You can use `recoverWith` here to avoid having to re-raise

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
```diff
@@ -104,37 +106,72 @@ abstract class SdkV1Relation[A <: Product, I](config: RelationConfig, shortName:
...
+          case e: CdpApiException if doUpsert && e.code == 409 =>
+            assertNoLegacyNameConflicts(e)
```
sebbert: Just checking, it's safe to throw an exception here, right?

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
wjoel: That's the one I was looking for. Thanks! Couldn't quite remember the name right.

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
wjoel: Fixed by matching on it. Thanks!

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
wjoel: Yes, it's safe. It will abort the Spark job and show the exception.

sebbert: Okay, great 👍

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
```diff
@@ -102,39 +83,78 @@ abstract class SdkV1Relation[A <: Product, I](config: RelationConfig, shortName:
...
-      resource.create(resourcesToCreate)
+    val create = if (resourcesToCreate.isEmpty) {
+      IO.unit
+    } else {
+      resource
+        .create(resourcesToCreate)
+        .flatTap(_ => incMetrics(itemsCreated, resourcesToCreate.size))
+        .map(_ => ())
+        .recoverWith {
+          case CdpApiException(_, 409, _, _, Some(duplicated), _, _) if doUpsert =>
```
sebbert: Agh, I really wish Scala had named pattern matching, but it's fine.

src/main/scala/cognite/spark/v1/TimeSeriesRelation.scala
```diff
@@ -55,29 +73,12 @@ class TimeSeriesRelation(config: RelationConfig, useLegacyName: Boolean)(val sql
...
-  def resolveConflict(existingExternalIds: Seq[String], timeSeriesSeq: Seq[TimeSeriesCreate]): IO[Unit] =
-    upsertAfterConflict[TimeSeries, TimeSeriesUpdate, TimeSeriesCreate, TimeSeriesResource[IO]](
-      existingExternalIds,
+    createOrUpdateByExternalId[TimeSeries, TimeSeriesUpdate, TimeSeriesCreate, TimeSeriesResource[IO]](
+      Seq.empty,
       timeSeriesSeq,
-      client.timeSeries)
+      client.timeSeries,
+      doUpsert = doUpsert)
+  }
```
sebbert: Seems like we're no longer handling legacyName conflicts here, is that intentional?

sebbert: Seems like we're no longer handling legacyName conflicts here, is that intentional?

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
sebbert: We could do this though, but I don't consider it a blocker for this PR

```
case e: CdpApiException if doUpsert && e.status == 409 && e.duplicated.isDefined =>
  let duplicated = e.duplicated.get
```

src/main/scala/cognite/spark/v1/TimeSeriesRelation.scala
wjoel: `createOrUpdateByExternalId` now always asserts that there are no legacy name conflicts (as we did before but only here), so it's dealt with in that method instead.

wjoel:
@sebbert / @hakontro: Feel free to merge if it looks ok, and maybe make a new release once https://github.com/cognitedata/cdp-spark-datasource/pull/357 is merged.

PR #357, opened 2019-12-17
Add isStep and isString to TimeSeriesUpsertSchema

sebbert:


sebbert:
It... worked? On the first try? Something is clearly wrong here.

src/main/scala/cognite/spark/v1/TimeSeriesRelation.scala
```diff
@@ -132,7 +132,9 @@ case class TimeSeriesUpsertSchema(
     metadata: Option[Map[String, String]] = None,
     assetId: Option[Long] = None,
     description: Option[String] = None,
-    securityCategories: Option[Seq[Long]] = None
+    securityCategories: Option[Seq[Long]] = None,
```
wjoel: Does the order matter? It's quite unsorted, if it matches what we use at doc.cognitedata.com it's ok I suppose. Otherwise id, name, description, metadata should all be closer to the top instead. Not sure if that breaks backwards compatibility but probably worth changing now rather than later even so.

src/main/scala/cognite/spark/v1/TimeSeriesRelation.scala
```diff
@@ -137,6 +137,19 @@ case class TimeSeriesUpsertSchema(
...
+case class TimeSeriesUpsertSchema(
+    id: Option[Long] = None,
+    name: Option[String] = None,
+    externalId: Option[String] = None,
+    metadata: Option[Map[String, String]] = None,
+    assetId: Option[Long] = None,
+    description: Option[String] = None,
+    securityCategories: Option[Seq[Long]] = None,
+    isStep: Boolean = false,
+    isString: Boolean = false
```
sebbert: These should be optional, no?

PR #362, opened 2020-01-20
Handle duplicates when using upsert setting for onConflict

wjoel:


src/main/scala/cognite/spark/v1/CdpConnector.scala
```diff
@@ -44,4 +32,14 @@ object CdpConnector {
...
+
+  // null values aren't allowed according to our schema, and also not allowed by CDP, but they can
+  // still end up here. Filter them out to avoid null pointer exceptions from Circe encoding.
+  // Since null keys don't make sense to CDP either, remove them as well.
+  // Additionally, values are limited to 512 characters, yet we still have data where values have
+  // more characters than that, so truncate them to the valid length if required: it's necessary for
+  // copying existing data, and probably for upserts as well.
+  def filterMetadata(metadata: Option[Map[String, String]]): Option[Map[String, String]] =
+    metadata.map(_.filter { case (k, v) => k != null && v != null }
+      .mapValues(_.slice(0, Constants.MetadataValuePostMaxLength)))
```
sebbert: Unrelated to this PR, but shouldn't we throw an error if a metadata value does not fit into 512 chars? Seems like something I would like to know upfront, not after I cannot find the data I'm looking for.

wjoel: Good question! The problem is that in some projects we have old data which has more than 512 chars, from before the rules were enforced. Such projects would be unable to safely process items without implementing something similar to this everywhere they read and write metadata, so we made the choice to do this silent truncation here and silently "fix" their existing item metadata values.

PR #367, opened 2020-01-24
[WIP] Prototype for Asset loader

hakontro:
Note: There are some inconsistencies in the updating part of the algorithm, leading to test errors and sometimes GC overload. I'm working on it!


src/test/scala/cognite/spark/v1/AssetsHierarchyBuilderTest.scala
```diff
@@ -0,0 +1,152 @@
+package cognite.spark.v1
+
+import cognite.spark.v1.SparkSchemaHelper.fromRow
+import com.cognite.sdk.scala.v1.AssetCreate
+import org.apache.spark.sql.Row
+import org.scalatest.{FlatSpec, Matchers}
+
+class AssetsHierarchyBuilderTest  extends FlatSpec with Matchers with SparkTest {
```
wjoel: Add tests for an empty input, and an input where things aren't connected at all (lots of null/None values), and an input where there is a cycle.

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
```diff
@@ -0,0 +1,209 @@
...
+      visited
+    } else {
+      visitKids(input, kids.map(_.externalId), visited ++ kids)
+    }
+  }
+
+  override def schema: StructType = structType[AssetsIngestSchema]
+}
+
+case class AssetsIngestSchema(
```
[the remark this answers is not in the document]
hakontro: Roger!

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -185,49 +190,54 @@ class DefaultSource
...
-        case OnConflict.UPDATE =>
-          batches.grouped(Constants.MaxConcurrentRequests).foreach { batchGroup =>
-            batchGroup.parTraverse(relation.update).unsafeRunSync()
-          }
-
-        case OnConflict.DELETE =>
-          batches.grouped(Constants.MaxConcurrentRequests).foreach { batchGroup =>
-            batchGroup.parTraverse(relation.delete).unsafeRunSync()
-          }
+    if (resourceType == "assetshierarchybuilder") {
```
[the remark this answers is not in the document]
hakontro: We could! I didn't want to rewrite the logic for batching since I wasn't sure this was going to go into the Data source, but I'll have a crack at it.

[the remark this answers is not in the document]
hakontro: Actually, no. The `AssetHierarchyBuilder` is a CdfRelation, but it doesn't have all the traits that the others have. `relation` is a `CdfRelation with TableScan with PrunedFilteredScan with InsertableRelation with WritableRelation`, and I don't want to add all that to `AssetHierarchyBuilder` just for this.

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
```diff
@@ -0,0 +1,221 @@
...
+
+    val root = maybeRoots.toList match {
+      case singleRoot :: Nil => singleRoot
+      case Nil => throw NoRootException()
+      case _ => throw MultipleRootsException()
+    }
+    (root, leaves)
+  }
+
+  def isTree(root: AssetsIngestSchema, leaves: Array[AssetsIngestSchema]): Set[AssetsIngestSchema] = {
```
wjoel: This should be named something else considering it returns a set of nodes. "is" should be for boolean return values.

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -185,49 +190,54 @@ class DefaultSource
...
-        case OnConflict.UPDATE =>
-          batches.grouped(Constants.MaxConcurrentRequests).foreach { batchGroup =>
-            batchGroup.parTraverse(relation.update).unsafeRunSync()
-          }
-
-        case OnConflict.DELETE =>
-          batches.grouped(Constants.MaxConcurrentRequests).foreach { batchGroup =>
-            batchGroup.parTraverse(relation.delete).unsafeRunSync()
-          }
+    if (resourceType == "assetshierarchy") {
```
wjoel: Still don't see why we couldn't just make this part of the match? Apologies if I missed your explanation on the previous review.

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -174,6 +177,8 @@ class DefaultSource
         val revisionId =
           parameters.getOrElse("revisionId", sys.error("Revision id must be specified")).toLong
         new ThreeDModelRevisionNodesRelation(config, modelId, revisionId)(sqlContext)
+      case "assetshierarchybuilder" =>
```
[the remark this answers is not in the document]
wjoel: We use `assetshierarchy` below but not here.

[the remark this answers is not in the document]
hakontro: Good catch, this shouldn't even be here.

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
```diff
@@ -0,0 +1,221 @@
...
+
+  def getCdfRootAndTree(sourceRootExternalId: CogniteExternalId): IO[(Option[Asset], List[Asset])] = {
+    val cdfTree = client.assets
+      .filter(AssetsFilter(rootIds = Some(Seq(sourceRootExternalId))))
+      .compile
+      .toList
+
+    for {
+      tree <- cdfTree
+      root <- IO(tree.find(_.externalId.contains(sourceRootExternalId.externalId)))
```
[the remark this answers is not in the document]
hakontro: Unfortunately not, we'd get an `Option[(Asset, List[Asset])]` instead.

src/main/scala/cognite/spark/v1/DefaultSource.scala
hakontro: There's probably a way, but the Asset hierarchy builder needs to work on the entire dataset, not a batched subset. We could move the batching into each of the implementations of the `WritetableResource` methods for the other resources, but I don't think it's worth it.

wjoel: Ah, I missed the `data.foreachPartition` part below the match. Alrightie then.

src/test/scala/cognite/spark/v1/SparkTest.scala
```diff
@@ -87,7 +87,8 @@ trait SparkTest {
       OnConflict.ABORT,
       spark.sparkContext.applicationId,
       Constants.DefaultParallelismPerPartition,
-      ignoreUnknownIds = true
+      ignoreUnknownIds = true,
+      false
```
wjoel: Should name this boolean variable.

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
```diff
@@ -0,0 +1,223 @@
...
+final case class InvalidTreeException(message: String = s"The tree is has an invalid structure.")
+    extends Exception(message)
+
+class AssetsHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
+    extends CdfRelation(config, "assetshierarchybuilder") {
+
+  def build(df: DataFrame, deleteMissingAssets: Boolean = false): IO[Unit] = {
+    val sourceTree = df.collect.map(r => fromRow[AssetsIngestSchema](r))
+
+    val (root, leaves) = validateAndOrderInput(sourceTree)
```
wjoel: "children" rather than leaves, leaves are nodes without children but `validateAndOrderInput` returns all nodes except the root.

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
```diff
@@ -0,0 +1,223 @@
...
+  def needsUpdate(leaf: AssetsIngestSchema, cdfTree: Seq[Asset]): Boolean = {
+    // Find the matching Asset in CDF
+    val cdfAsset = cdfTree.find(a => a.externalId.contains(leaf.externalId))
+    cdfAsset match {
+      case Some(asset) =>
+        // Find the parent Asset in CDF to be able to get the externalId of the parent
+        cdfTree.find(a => asset.parentId.contains(a.id)) match {
+          case Some(parentAsset) =>
+            !isMostlyEqual(leaf, asset) ||
+              !parentAsset.externalId.contains(leaf.parentExternalId)
```
wjoel: Should be on the same line as `||` now

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
```diff
@@ -0,0 +1,223 @@
...
+    val (root, leaves) = splitIntoRootAndLeaves(tree)
+
+    (root, getValidatedTree(root, leaves))
+  }
+
+  def splitIntoRootAndLeaves(
+      tree: Array[AssetsIngestSchema]): (AssetsIngestSchema, Array[AssetsIngestSchema]) = {
+    val (maybeRoots, leaves) = tree.partition(_.parentExternalId == "")
+
+    val root = maybeRoots.toList match {
```
wjoel: Ah, so it's an array. We can pattern match on those as well and avoid the `.toList`: https://stackoverflow.com/questions/6647166/how-do-i-pattern-match-arrays-in-scala

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  def isMostlyEqual(leaf: AssetsIngestSchema, asset: Asset): Boolean =
+    leaf.description == asset.description &&
+      leaf.metadata == asset.metadata &&
+      leaf.name == asset.name &&
+      leaf.source == asset.source
+
+  def validateAndOrderInput(
+      tree: Array[AssetsIngestSchema]): (AssetsIngestSchema, Set[AssetsIngestSchema]) = {
```
wjoel: This is likely my fault after bad comments on a previous review :( but there's no order to a `Set` so it should probably be a `Seq` (or `List`, or whatever).
Our tests should catch that though, right?

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
```diff
@@ -0,0 +1,223 @@
...
+      root: AssetsIngestSchema,
+      leaves: Array[AssetsIngestSchema]): Set[AssetsIngestSchema] = {
+    val visited = Set[AssetsIngestSchema]()
+    val insertableTree = iterateChildren(leaves, Array(root.externalId), visited)
+
+    if (insertableTree.map(_.externalId) == leaves.map(_.externalId).toSet) {
+      insertableTree
+    } else {
+      val assetsNotInTree =
+        leaves.filter(a => insertableTree.map(_.externalId).contains(a.externalId)).map(_.externalId)
```
wjoel: `leaves.map(_.externalId).filter(externalId => insertableTree.map(_.externalId).contains(externalId))`
but since we don't want to have to do this mapping on insertableTree for ever element of leaves we should do that once at the top: `val insertableTreeExternalIds = insertableTree.map(_.externalId)`

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
```diff
@@ -0,0 +1,223 @@
...
+      case singleRoot :: Nil => singleRoot
+      case Nil => throw NoRootException()
+      case _ => throw MultipleRootsException()
+    }
+    (root, leaves)
+  }
+
+  def getValidatedTree(
+      root: AssetsIngestSchema,
+      leaves: Array[AssetsIngestSchema]): Set[AssetsIngestSchema] = {
```
wjoel: `children` instead.

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
hakontro: Decided to go for the `toList` for niceness, but turns out it was even nicer with Arrays regardless!

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
hakontro: This applied a lot of places, particularly to everyone `leaf/leaves` had been used.

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
hakontro: Tests don't catch it yet, I suppose because the asset trees are small enough to always fit in a single batch. We need to ensure that the batching happens in the right places though, or this won't work for asset trees with more than 1,000 assets.

hakontro:
I made `batchSize` configurable, but mostly just to set it when testing to avoid the need for writing huge test trees.

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
```diff
@@ -0,0 +1,230 @@
...
+final case class MultipleRootsException(message: String = "Tree has more than one root.")
+    extends Exception(message)
+final case class NoRootException(
+    message: String = """Tree has no root. Set parentExternalId to "" for the root Asset.""")
+    extends Exception(message)
+final case class InvalidTreeException(message: String = s"The tree is has an invalid structure.")
+    extends Exception(message)
+
+class AssetsHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
+    extends CdfRelation(config, "assetshierarchybuilder") {
```
wjoel: Shouldn't this also be `"assetshierarchy"`?

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
```diff
@@ -0,0 +1,230 @@
...
+      _ <- insert(toInsert, batchSize)
+      _ <- update(toUpdate, batchSize)
+      _ <- delete(toDelete, deleteMissingAssets, batchSize)
+    } yield ()
+  }
+
+  def insert(toInsert: Seq[AssetsIngestSchema], batchSize: Int): IO[Unit] =
+    if (toInsert.nonEmpty) {
+      val assetCreatesToInsert = toInsert.map(AssetsIngestSchema.toAssetCreate).grouped(batchSize)
+      assetCreatesToInsert.toList.traverse(client.assets.create) *> IO.unit
```
wjoel: Maybe parTraverse instead?

wjoel: For consistency with `delete` and `update` we should format it the same
```
  toInsert...
    .grouped(batchSize)
    .toList
    .parTraverse(client.assets.create) *> IO.unit
```

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
```diff
@@ -0,0 +1,230 @@
...
+
+  def upsertRoot(
+      root: AssetsIngestSchema,
+      sourceRoot: Option[Asset]
+  ): IO[Unit] =
+    sourceRoot match {
+      case None =>
+        client.assets.create(
+          Seq(
+            AssetCreate(
```
wjoel: Could we use `AssetsIngestSchema.toAssetCreate` here or is it doing something special?

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+  def nodesToDeleteInsertUpdate(
+      root: AssetsIngestSchema,
+      children: Seq[AssetsIngestSchema],
+      cdfTree: Seq[Asset]): (Seq[Asset], Seq[AssetsIngestSchema], Seq[AssetsIngestSchema]) = {
+    // Insert assets that are not present in CDF
+    val toInsert = children.filter(a => !cdfTree.map(_.externalId.get).contains(a.externalId))
+
+    // Keep assets that are in both source and cdf, delete those only in CDF
```
wjoel: "and CDF" (all caps)

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
hakontro: I used `.traverse` here since it (to my understanding) preserves order, which is necessary to avoid writing children before their parents exist.

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
hakontro: I let `parentExternalId == ""` rather than `null` denote the root Asset, to let the schema have `parentExternalId: String` rather than `parentExternalId: Option[String]`. This handles that by setting `parentExternalId` to `None` in the `AssetCreate`.

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
hakontro: I should!

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
wjoel: Hmm, so it is. Add a comment about that, please.

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
wjoel: `AssetsIngestSchema.toAssetCreate(root).copy(parentExternalId = None)`?

hakontro: Yes!

PR #371, opened 2020-02-03
Fix camel case typos in aggregation options

hakontro:


sebbert:
This seems like it would be a breaking change? If so, I think we should either bump the major version, or make it backwards compatible somehow.

hakontro:
> This seems like it would be a breaking change? If so, I think we should either bump the major version, or make it backwards compatible somehow.

This is actually a bug that nobody had reported yet - turns out it wasn't possible to read any of the camelCased aggregations before.

PR #373, opened 2020-02-03
Speed up files test by filtering with SDK

hakontro:
Files update is struggling with slow Elastic Search results. This uses the SDK filtering method, and runs a lot faster.

### DIFF
--- a/src/test/scala/cognite/spark/v1/FilesRelationTest.scala
+++ b/src/test/scala/cognite/spark/v1/FilesRelationTest.scala
@@ -1,6 +1,6 @@
 package cognite.spark.v1
 
-import org.apache.spark.sql.Row
+import com.cognite.sdk.scala.v1.{File, FilesFilter}
 import org.apache.spark.sql.functions.col
 import org.scalatest.{FlatSpec, Matchers}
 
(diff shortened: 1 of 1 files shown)


src/test/scala/cognite/spark/v1/FilesRelationTest.scala
```diff
@@ -72,9 +70,9 @@ class FilesRelationTest extends FlatSpec with Matchers with SparkTest {
       .write
       .insertInto("filesSource")
 
-    val dfWithTestDirectory = retryWhile[Array[Row]](
-      spark.sql(s"select * from filesSource where assetIds = Array(1739842716040355)").collect,
-      rows => rows.length < 10)
+    val dfWithTestDirectory = retryWhile[List[File]](
+      writeClient.files.filter(FilesFilter(assetIds = Some(Seq(1739842716040355L)))).compile.toList,
```
wjoel: Is it slow because we don't do pushdown on this, or because Spark overhead is large?

src/test/scala/cognite/spark/v1/FilesRelationTest.scala
```diff
@@ -72,9 +70,9 @@ class FilesRelationTest extends FlatSpec with Matchers with SparkTest {
...
+      files => files.length != 10)
```
wjoel: "< 10" is better, because if someone creates a file matching this at least the test will fail sooner on the next line instead of waiting for the full timeout.

src/test/scala/cognite/spark/v1/FilesRelationTest.scala
hakontro: This was slow because `files/list` was slow. We don't have pushdowns for Files metadata, but filter just happened to be a lot faster than list and this was the easiest option.

PR #374, opened 2020-02-04
Throw error in Asset builder on empty externalId

hakontro:


src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
```diff
@@ -45,6 +45,8 @@ final case class NoRootException(
     extends Exception(message)
 final case class InvalidTreeException(message: String = s"The tree is has an invalid structure.")
     extends Exception(message)
+final case class EmptyExternalIdException(message: String = s"ExternalId cannot be an empty String.")
```
wjoel: We should include some info about which row caused this, "Found empty externalId for child with externalId = ${externalId}" (or other identifier, if we have one)

src/test/scala/cognite/spark/v1/FilesRelationTest.scala
```diff
@@ -72,9 +70,9 @@ class FilesRelationTest extends FlatSpec with Matchers with SparkTest {
       .write
       .insertInto("filesSource")
 
-    val dfWithTestDirectory = retryWhile[Array[Row]](
-      spark.sql(s"select * from filesSource where assetIds = Array(1739842716040355)").collect,
-      rows => rows.length < 10)
+    val dfWithTestDirectory = retryWhile[List[File]](
+      writeClient.files.filter(FilesFilter(assetIds = Some(Seq(1739842716040355L)))).compile.toList,
+      files => files.length != 10)
```
wjoel: This will get stuck for a while if someone else is creating files like this. Then it will fail on the next line anyway. `< 10` is better imo (though of course it will get stuck if someone is deleting files, but that seems to be more rare)

src/main/scala/cognite/spark/v1/AssetsHierarchyBuilder.scala
hakontro: This is only used when `externalId == ""`, so doesn't make much sense returning it, but we can return the name.

wjoel: I thought it might be a parent with external id "". But if the child itself has "" as the external id, perhaps we can display something else about the row to help the user identify it?

hakontro: For sure, we can go with the `name`. If that's `""` as well I suppose people will have to filter their data to find the culprits, but that should be fine.

src/test/scala/cognite/spark/v1/FilesRelationTest.scala
hakontro: Agreed, will fix this in #373 and rebase when that's merged.

PR #377, opened 2020-02-04
[DT-1337] Support setting legacyName from externalId for time series

sebbert:
Make sure to merge #382 before merging this PR.

Previously, when the `useLegacyName` option was set to "true",
we would set `legacyName` based on the `name` of the time series.

This extends the `useLegacyName` option to allow the following:

```scala
  .option("useLegacyName", false)        // legacyName = null
  .option("useLegacyName", true)         // legacyName = name
  .option("useLegacyName", "name")       // legacyName = name
  .option("useLegacyName", "externalId") // legacyName = externalId
```

Setting this to "true" has the same effect as "name", so that we
maintain backwards compatibility.

This PR also fixes inconsistent naming in the OnConflict enumeration.


src/main/scala/cognite/spark/v1/TimeSeriesRelation.scala
```diff
@@ -133,6 +137,19 @@ object TimeSeriesRelation extends UpsertSchema {
   val readSchema = structType[TimeSeriesReadSchema]
 }
 
+object LegacyNameSource extends Enumeration {
+  type LegacyNameSource = Value
+  val None, Name, ExternalId = Value
```
hakontro: We used uppercase Enumerations for the `OnConflict` handling, but I see Spark used Pascal for their `SaveMode` Enumerations. Perhaps you could change the existing ones while you're at it for consistency.

PR #381, opened 2020-02-07
Document asset hierarchy builder

hakontro:


README.md
```diff
@@ -133,6 +137,75 @@ Assets and events will ignore existing ids on deletes. If you prefer to abort th
 when an attempt to delete an unknown id, use `.option("ignoreUnknownIds", "true")`
 for those resources types.
 
+## Asset hierarchy (beta)
+Note: The asset hierarchy utility is currently in beta, and has not been sufficiently tested to be used
```
sebbert: Maybe change "The asset hierarchy utility" to "Asset hierarchy ingestion"? Not sure though, I think it's fine as is really.

hakontro: I was conflicted here as well, what should we call it? Ingestion sounds like it's a first time thing, while the tool is actually most useful when there are updates.

README.md
```diff
@@ -133,6 +137,75 @@ Assets and events will ignore existing ids on deletes. If you prefer to abort th
 when an attempt to delete an unknown id, use `.option("ignoreUnknownIds", "true")`
 for those resources types.
 
+## Asset hierarchy (beta)
+Note: The asset hierarchy utility is currently in beta, and has not been sufficiently tested to be used 
+on production data.
+
+The `.option("type", "assethierarchy")` lets you write new asset hierarchies, or update existing ones,
+using the Spark data source. It currently supports writing a single root asset and any number of children.
```
sebbert: Spark Data Source should be Capitalized

README.md
```diff
@@ -133,6 +137,75 @@ Assets and events will ignore existing ids on deletes. If you prefer to abort th
...
+The asset hierarchy utility can ingest entire trees, as long as all nodes are connected to the root asset
+through the `externalId`/`parentExternalId` relationship. If input contains an update to data that already exists,
+i.e there's a match on `externalId` and there's a change to one of the other fields, the asset will be updated.
+There's also an option to delete assets from CDF that are not referenced in the input data.
+
+### Requirements
+- The source data must contain a single root, denoted by setting its `parentExternalId` to the empty string `""`.
+- All assets, except root, must be connected to another asset in the input data by `parentExternalId`.
+- The input data must not have loops, to ensure all asset trees are fully connected.
```
sebbert: Are we doing anything to detect cycles?

hakontro: We are requiring all assets to be connected to the root, which is not connected to anything, which will discover cycles (at least I can't think of a counter-example)

README.md
```diff
@@ -133,6 +137,75 @@ Assets and events will ignore existing ids on deletes. If you prefer to abort th
...
+- `externalId` can not be the empty string `""`.
+
+### Options
+|    Option                  | Default | Description |
+| -------------------------- | --------|----------------------------------------------------------------------------------------------------------- |
+| `deleteMissingAssets`      | `false` | Whether or not you would like assets under the root to be deleted if they're not present in the input data |
+| `ignoreDisconnectedAssets` | `false` | This will ignore assets that are not connected to root, as opposed to throw an error and stop the program. |
```
sebbert:
```suggestion
| `ignoreDisconnectedAssets` | `false` | This will ignore assets that are not connected to root, as opposed to throwing an error and stopping the program. |
```

README.md
```diff
@@ -133,6 +137,75 @@ Assets and events will ignore existing ids on deletes. If you prefer to abort th
...
+  Seq("second_child", "root_asset", "manual_input", "second_child", Some("This is the second_child"), None),
+  Seq("grandchild", "", "manual_input", "first_child", Some("This is the root asset"), None)
+)
+
+val assetTreeDataFrame = spark
+  .sparkContext
+  .parallelize(assetTree)
+  .toDF(assetHierarchySchema) // The column names
+
+// Insert the assets with the asset hierarchy tool
```
sebbert: I see we're using the term "asset hierarchy utility" elsewhere in this article, so let's be consistent about what we call it :)

README.md
```diff
@@ -133,6 +137,75 @@ Assets and events will ignore existing ids on deletes. If you prefer to abort th
 when an attempt to delete an unknown id, use `.option("ignoreUnknownIds", "true")`
 for those resources types.
 
+## Asset hierarchy (beta)
```
sebbert: I don't exactly know why, but the title "Asset hierarchy" feels a bit off to me. Maybe "Asset hierarchies" or "Asset hierarchy ingestion"? Feel free to ignore this, as it's only my gut feeling :)

README.md
```diff
@@ -133,6 +137,75 @@ Assets and events will ignore existing ids on deletes. If you prefer to abort th
...
+  .toDF(assetHierarchySchema) // The column names
+
+// Insert the assets with the asset hierarchy tool
+assetTreeDataFrame
+  .format("cognite.spark.v1")
+  .option("apiKey", "myApiKey")
+  .option("type", "assethierarchy")
+  .save
+
+// Have a look at your new asset tree
```
sebbert:
```suggestion
// Have a look at your new asset hierarchy
```

README.md
hakontro: I don't really like utility either, I just chose it because I thought builder _could_ be misleading. In retrospect I like builder more anyways, so let's go with that.

README.md
hakontro: I'll change this everywhere to avoid multiple terms for the same thing.

README.md
hakontro: Will go for builder here as well

README.md
sebbert: I agree, "ingestion" is not the right word here. I've heard the term "asset syncing" a lot, but I'm not sure if that's the right term either. I'll leave it up to you.

README.md
sebbert: That's true, good point 👍

README.md
```diff
@@ -13,6 +13,10 @@ The instructions below explain how to read from, and write to, the different res
     - [Read data](#read-data)
     - [Write data](#write-data)
     - [Delete data](#delete-data)
+- [Asset hierarchy (beta)](#asset-hierarchy-beta)
```
sebbert:
```suggestion
- [Asset hierarchy builder (beta)](#asset-hierarchy-builder-beta)
```

README.md
```diff
@@ -133,6 +137,75 @@ Assets and events will ignore existing ids on deletes. If you prefer to abort th
...
+### Example
+```scala
+val assetHierarchySchema = Seq("externalId", "parentExternalId", "description", "name", "metadata")
+
+// Manually create some assets that satisfy the requirements of the asset hierarchy builder
+val assetHierarchy = Seq(
+  Seq("root_asset", "", "manual_input", "root_asset", Some("This is the root asset"), None),
+  Seq("first_child", "root_asset", "manual_input", "first_child", Some("This is the first_child"), None),
+  Seq("second_child", "root_asset", "manual_input", "second_child", Some("This is the second_child"), None),
+  Seq("grandchild", "", "manual_input", "first_child", Some("This is the root asset"), None)
```
sebbert: Shouldn't this have `first_child` or `second_child` as its `parentExternalId`? Seems like we have two root assets here

hakontro: You're correct, I see I changed the `source` instead of the `parentExternalId`. I'll run the code to verify it.

sebbert:
@hakontro Let me know when you have fixed the remaining issues, and I think we're good to go 👍

hakontro:
> @hakontro Let me know when you have fixed the remaining issues, and I think we're good to go

Should be good now, behaves as expected when testing.

README.md
hakontro: Fixed

README.md
```diff
@@ -133,6 +137,78 @@ Assets and events will ignore existing ids on deletes. If you prefer to abort th
...
+- The source data must contain a single root, denoted by setting its `parentExternalId` to the empty string `""`.
+- All assets, except root, must be connected to another asset in the input data by `parentExternalId`.
+- The input data must not have loops, to ensure all asset hierarchies are fully connected.
+- `externalId` can not be the empty string `""`.
+
+### Options
+|    Option                  | Default | Description |
+| -------------------------- | --------|------------------------------------------------------------------------------------------------------------ |
+| `deleteMissingAssets`      | `false` | Whether or not you would like assets under the root to be deleted if they're not present in the input data. |
+| `ignoreDisconnectedAssets` | `false` | This will ignore assets that are not connected to root, as opposed to throw an error and stop the program.  |
```
sebbert:
```suggestion
| `ignoreDisconnectedAssets` | `false` | This will ignore assets that are not connected to root, as opposed to throwing an error and stopping the program.  |
```

hakontro: Good catch! I actually edited this after the previous comment, but I simply replaced it with itself which doesn't help much.

PR #386, opened 2020-02-17
Add data set id fields

exyi:
Adds dataSetId Read/Insert/Upsert schemas of the data source.

I have only fixed existing tests to count with the `dataSetId` field, not added any new tests. I think the existing tests cover well that the mapping actually works.

Note that filtering based on `dataSetId` is not supported, I don't know if that's something we'd like to have.


exyi:
> I would like to see some tests that attempt to read the resources back again, and assert that the dataSetId matches what you provide during creation.

Sure, I have added a dataSet into the `jetfiretest2` db and reused Asset test to check that the dataSetId is inserted correctly.

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
```diff
@@ -145,9 +146,13 @@ class AssetsRelationTest extends FlatSpec with Matchers with SparkTest {
     val assetsCreated = getNumberOfRowsCreated(metricsPrefix, "assets")
     assert(assetsCreated == 1)
 
-    retryWhile[Array[Row]](
+    val Array(createdAsset) = retryWhile[Array[Row]](
```
sebbert: Pro tip: You can use [`loneElement`](http://doc.scalatest.org/3.0.1-2.12/org/scalatest/LoneElement.html) to get better error messages when you expect a collection to contain a single element. I don't consider that a blocker for this PR though.

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
```diff
@@ -145,9 +146,13 @@ class AssetsRelationTest extends FlatSpec with Matchers with SparkTest {
...
       spark.sql(s"select * from assets where source = '$assetsTestSource'").collect,
       rows => rows.length < 1)
+
+    assert(createdAsset.getAs[String]("name") == "asset name")
+    assert(createdAsset.getAs[Long]("dataSetId") == testDataSetId)
+    assert(createdAsset.getAs[String]("externalId") == "1")
```
sebbert: We should use the `x should be(y)` or `x shouldBe y` matchers when asserting equality.

```suggestion
    createdAsset.getAs[String]("name") shouldBe "asset name"
    createdAsset.getAs[Long]("dataSetId") shouldBe testDataSetId
    createdAsset.getAs[String]("externalId") shouldBe "1"
```

sebbert: We should also perform the same checks for other resource types (events, time series, etc.).

exyi: If you mean... However, I managed to find a disadvantage of `shouldBe` - when I get the types wrong (for example `.getAs[Option[Long]]("x") == 1`), Scala compiler complain. It just compiles and fails the test when I use use `shouldBe` :/

sebbert: Agh, that's a good point. Feel free to disregard this comment then.

sebbert: I would still like to see tests for other resource types though :)

PR #387, opened 2020-02-20
[DT-1376] Support upserts when useLegacyName = externalId

sebbert:


BergsethCognite:
Bug is blocking Jetfire migration @ Lundin

PR #388, opened 2020-02-21
Add metrics to AssetHierarchyBuilder

exyi:


src/main/scala/cognite/spark/v1/CdfRelation.scala
```diff
@@ -17,6 +17,10 @@ abstract class CdfRelation(config: RelationConfig, shortName: String)
     MetricsSource.getOrCreateCounter(config.metricsPrefix, s"$shortName.created")
   @transient lazy protected val itemsUpdated: Counter =
     MetricsSource.getOrCreateCounter(config.metricsPrefix, s"$shortName.updated")
+  @transient lazy protected val itemsIgnored: Counter =
+    MetricsSource.getOrCreateCounter(config.metricsPrefix, s"$shortName.ignored")
```
wjoel: I don't think we should have this as a metric, it will show up in the Jetfire UI and I think it will be confusing for the user to see "ignored". @sebbert what do you think?

sebbert: I agree that it will look confusing, but at the same time I can imagine users would like to know if there are disconnected assets in their source data.

sebbert: Then again, you'll be able to tell if the number of reads != number of writes, so I'd say we remove it for now.

exyi: > you'll be able to tell if the number of reads != number of writes, so I'd say we remove it for now.

That could have quite a lot of possible explanations, especially for more complex queries (suboptimal filter pushdown, ...). I don't really know what do our users do when the assets will be ignored. Does it occur often?

sebbert: That's a very good point. Assets are ignored when they have no connection to the root asset, and the `ignoreDisconnectedAssets` option is enabled. I don't really see the harm in displaying that metric in Jetfire, as long as it's documented.

exyi: Let's include it for now, maybe we will rename it to `disconnected` as it explains why is the item actually ignored. For now, `ignored` is at least indication that something strange happened. And it's better if users will be asking "what is this `ignored`" than "why is my data missing"

PR #389, opened 2020-02-21
[DT-1255] Delete metrics

exyi:
Adds metrics for delete operations.


exyi:
@wjoel
> what about events?

As most resources, events rely on delete code from `SdkV1Relation`.

PR #390, opened 2020-02-24
Add support for dataSetId in AssetHierarchyBuilder

exyi:


src/test/scala/cognite/spark/v1/AssetHierarchyBuilderTest.scala
```diff
@@ -123,14 +125,15 @@ class AssetHierarchyBuilderTest extends FlatSpec with Matchers with SparkTest {
       .save
 
     val result = retryWhile[Array[Row]](
-      spark.sql(s"select * from assets where source = '$testName'").collect,
+      spark.sql(s"select * from assets where source = '$testName' and dataSetId = $testDataSetId").collect,
```
wjoel: Could you add a test with `dataSetId = null`? It should ignore the null in that case (ie. not set dataSetId at all).

wjoel: Or `dataSetId is null`

PR #391, opened 2020-02-24
[DT-900] Add support for empty updates

exyi:
This fixes a bug, where only some updated items have non-null fields.

### DIFF
--- a/src/main/scala/cognite/spark/v1/AssetsRelation.scala
+++ b/src/main/scala/cognite/spark/v1/AssetsRelation.scala
@@ -61,11 +61,14 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
       .flatTap(_ => incMetrics(itemsCreated, assets.size)) *> IO.unit
   }
 
+  private def isUpdateEmpty(u: AssetUpdate): Boolean = u == AssetUpdate()
+
   override def update(rows: Seq[Row]): IO[Unit] = {
     val assetUpdates = rows.map(r => fromRow[AssetsUpsertSchema](r))
     updateByIdOrExternalId[AssetsUpsertSchema, AssetUpdate, Assets[IO], Asset](
       assetUpdates,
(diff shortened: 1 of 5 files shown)


src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -61,11 +61,14 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
       .flatTap(_ => incMetrics(itemsCreated, assets.size)) *> IO.unit
   }
 
+  def isUpdateEmpty(u: AssetUpdate): Boolean = u == TimeSeriesUpdate()
```
wjoel: Make this private and should probably use `===` from https://typelevel.org/cats/typeclasses/eq.html to catch this type error ;)

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
```diff
@@ -77,17 +77,26 @@ abstract class SdkV1Relation[A <: Product, I](config: RelationConfig, shortName:
...
     require(
       updates.forall(u => u.id.isDefined || u.externalId.isDefined),
       "Update requires an id or externalId to be set for each row.")
-    val (updatesById, updatesByExternalId) = updates.partition(u => u.id.exists(_ > 0))
+    val (rawUpdatesById, updatesByExternalId) = updates.partition(u => u.id.exists(_ > 0))
+    val updatesById =
+      rawUpdatesById
+        .map(u => u.id.get -> u.transformInto[U])
+        .filter({
+          case (_id, update) => !isUpdateEmpty(update)
```
wjoel: Use `_` only if you want to ignore id, instead of `_id`. In this case, you also don't need both `(` and `{`: Use only `{` instead.
```scala
.filter { case (_, update) => !isUpdateEmpty(update) }
 ```

PR #392, opened 2020-02-25
[DT-1275] Improve error message for column with invalid type or missing value

exyi:
Before this change, Spark failed with message like "failed due to missing value column",
when the type was not matching the expected one, or the value was null.
Now it prints which column has invalid type and also on which row the error occurs

### DIFF
--- a/src/test/scala/cognite/spark/v1/DataPointsRelationTest.scala
+++ b/src/test/scala/cognite/spark/v1/DataPointsRelationTest.scala
@@ -502,6 +502,64 @@ class DataPointsRelationTest extends FlatSpec with Matchers with SparkTest {
     assert(dataPointsAfterPostByExternalId.length == timestamps.length)
   }
 
+  it should "fail reasonably when datapoint values are invalid" taggedAs WriteTest in {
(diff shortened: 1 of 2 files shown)


src/test/scala/cognite/spark/v1/DataPointsRelationTest.scala
```diff
@@ -402,6 +402,62 @@ class DataPointsRelationTest extends FlatSpec with Matchers with SparkTest {
...
+      .save()
+
+    val Array(tsId) = retryWhile[Array[Row]](
+      spark
+        .sql(s"""select id from destinationTimeSeries where externalId = '$tsName'""")
+        .collect,
+      df => df.isEmpty)
+      .map(r => r.getLong(0).toString)
+
+    val exception = intercept[SparkException] {
```
wjoel: Disable Spark logging before this, re-enable after the expected exception.

wjoel:
@exyi time for a rebase and re-test?

exyi:
Merging. The test failing is "EventsRelation: should allow simple reads" which has nothing to do with this change, so we blame a production issue in the Events database.

PR #396, opened 2020-03-04
Release version 1.2.15

hakontro:


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 1.2.15
+
+## Fixes
+
+* Fixes a bug where not all datapoints would be read if a timeseries had infrequent datapoints.
```
wjoel: "data points", "time series" is how we write it.

Tapped:
Should we also just fix the issue where we are not setting the limit for Raw? It will improve perf a lot I assume.

hakontro:
> Should we also just fix the issue where we are not setting the limit for Raw? It will improve perf a lot I assume.

Feel free to sneak it in here if it's a quick fix, but I'd like to get this released today.

Tapped:
Should be a quick fix, low effort and very high value

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 1.2.15
+
+## Fixes
+
+* Fixes a bug where not all data points would be read if a time series had infrequent data points.
```
wjoel: Maybe we should be more specific about how infrequent they had to be for this bug to occur?

hakontro:
Note: I'm disabling one of the time series tests, since an old time series has been stuck in time series limbo. Will re-enable tomorrow to see if the issue resolves by itself.

PR #397, opened 2020-03-05
Increase number of data points to write per batch

hakontro:
Before this we were only writing batches of 1,000 data points. This will increase this to 10,000 data points.

### DIFF
--- a/src/main/scala/cognite/spark/v1/Constants.scala
+++ b/src/main/scala/cognite/spark/v1/Constants.scala
@@ -5,6 +5,7 @@ import BuildInfo.BuildInfo
 import scala.concurrent.duration._
 
 object Constants {
+  val CreateDataPointsLimit = 100000
   val DefaultMaxRetries = 10
   val DefaultBatchSize = 1000
   val DefaultRawBatchSize = 10000
--- a/src/main/scala/cognite/spark/v1/DefaultSource.scala
+++ b/src/main/scala/cognite/spark/v1/DefaultSource.scala
@@ -233,28 +233,34 @@ class DefaultSource
           new StringDataPointsRelationV1(config)(sqlContext)
         case _ => sys.error(s"Resource type $resourceType does not support save()")
       }
+      val batchSize = relation match {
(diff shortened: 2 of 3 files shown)


src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -233,28 +233,33 @@ class DefaultSource
           new StringDataPointsRelationV1(config)(sqlContext)
         case _ => sys.error(s"Resource type $resourceType does not support save()")
       }
+      val batchSize = relation match {
+        case n: NumericDataPointsRelationV1 => Constants.CreateDataPointsLimit
```
exyi: I think this should also apply to string data points, does it?

src/main/scala/cognite/spark/v1/Constants.scala
```diff
@@ -5,6 +5,7 @@ import BuildInfo.BuildInfo
 import scala.concurrent.duration._
 
 object Constants {
+  val CreateDataPointsLimit = 10000
```
exyi: Why 10,000 when the limit is higher?

https://docs.cognite.com/api/v1/#operation/postMultiTimeSeriesDatapoints
>  The limit per request is 100000 datapoints.

hakontro: Ohh, good catch! I thought it was 10k from this part
>Array of DatapointsWithInternalId (object) or DatapointsWithExternalId (object) (DatapointsPostDatapoint) [ 1 .. *10000* ] items

src/main/scala/cognite/spark/v1/DefaultSource.scala
hakontro: It should, I just missed it. I'll implement it, thanks for pointing it out.

wjoel:
> Before this we were only writing batches of 1,000 data points. This will increase this to 10,000 data points.

For the record, it seems to actually increase it to 100,000 (which is good).

PR #399, opened 2020-03-09
[DT-1424] Only infer schema on a small subset of the table

Tapped:
Before we ended up reading 200 (numPartitions) * 10000 (defaultInferenceLimit) just to infer schema. This PR sets num partitions to 1 during schema inference.

I don't think we've a good way of testing this unfortunately.


wjoel:
@Tapped we need to fix the RAW schema inference test.
```
[info] Infer Schema
[info] - should use a different limit for infer schema *** FAILED ***
[info]   1 was not equal to 10 (RawTableRelationTest.scala:239)
```

Tapped:
I'll fix thanks

PR #400, opened 2020-03-09
[DT-1371][DT-1373] Asset Hierarchy generalization

exyi:
Add support for inserting subtrees (i.e. nodes rooted outside of the inserted tree) and multiple trees at once.

The builder now works like this:
* Group the input data into a set of trees (connected components)
   - and validate that it does not contain cycle. Note that previously cycles were allowed when `ignoreDisconnectedAssets` was set
* If `allowSubtreeIngestion` is `false`
    - drop trees without root when if `ignoreDisconnectedAssets` is set
    - otherwise raise an error
* If `allowMultipleRoots`, check that there is only one tree
* Empty insertions are now allowed (this will quite certainly if you are doing incremental updates). Only when the input set was non-empty and all nodes were dropped, the error is raised.
* Check that all subtree have an existing parent.
* The trees are inserted one-by-one (*1):
    - The corresponding subtree is downloaded from CDF for comparison
    - Which nodes are to be inserted, updated, deleted or kept is decided (based on matching `externalId`)
    - The insert, update, delete operations are performed (without parallelism)

Notes:

I think I have removed quadratic complexity, so it should now handle non-trivial asset hierarchies just fine. It still has to fit into RAM of one executor, and subtrees cannot be larger than 100,000 items.

*1: I don't want to make it parallel to prevent races. In case two trees overlap - one is on the path to root to another tree and `deleteMissingAssets` is set, then the result (if it error or not) depends on the tree order.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ i.e there's a match on `externalId` and there's a change to one of the other fie
 There's also an option to delete assets from CDF that are not referenced in the input data.
 
 ### Requirements
-- The root is denoted by setting its `parentExternalId` to the empty string `""`.
+- Root assets are denoted by setting their `parentExternalId` to the empty string `""`.
 - The input data must not have loops, to ensure all asset hierarchies are fully connected.
 - `externalId` can not be the empty string `""`.
 
--- a/README.md
+++ b/README.md
@@ -160,9 +160,7 @@ There's also an option to delete assets from CDF that are not referenced in the 
 |    Option                  | Default | Description |
 | -------------------------- | --------|------------------------------------------------------------------------------------------------------------------ |
 | `deleteMissingAssets`      | `false` | Whether or not you would like assets under the root to be deleted if they're not present in the input data.       |
-| `ignoreDisconnectedAssets` | `false` | This will ignore assets that are not connected to root, as opposed to throwing an error and stopping the program. |
-| `allowMultipleRoots`       | `true`  | Whether it's allowed to ingest more than one (sub)tree. |
-| `allowSubtreeIngestion`    | `true`  | Whether ingestion of trees without a root node is allowed. Otherwise, an error is raised or the items are ignored when `ignoreDisconnectedAssets` is set |
+| `subtrees` | `ingest` | Controls what should happen with subtrees without a root node in the input data. `ingest` says they will be processed and loaded into CDF, `ignore` will ignore all of them and `error` will stop the execution and raise an error (nothing will be ingested). |
 | `batchSize`                | 1000    | The number of assets to write per API call.                                                                       |
 
 ### Example
--- a/README.md
+++ b/README.md
@@ -145,15 +145,14 @@ Note: The asset hierarchy builder is currently in beta, and has not been suffici
 on production data.
 
 The `.option("type", "assethierarchy")` lets you write new asset hierarchies, or update existing ones,
-using the Spark Data Source. It currently supports writing a single root asset and any number of children.
-The asset hierarchy builder can ingest entire hierarchies, as long as all nodes are connected to the root asset
+using the Spark Data Source.
+The asset hierarchy builder can ingest entire hierarchies of nodes connected
 through the `externalId`/`parentExternalId` relationship. If input contains an update to data that already exists,
 i.e there's a match on `externalId` and there's a change to one of the other fields, the asset will be updated.
 There's also an option to delete assets from CDF that are not referenced in the input data.
 
 ### Requirements
-- The source data must contain a single root, denoted by setting its `parentExternalId` to the empty string `""`.
-- All assets, except root, must be connected to another asset in the input data by `parentExternalId`.
+- The root is denoted by setting its `parentExternalId` to the empty string `""`.
 - The input data must not have loops, to ensure all asset hierarchies are fully connected.
 - `externalId` can not be the empty string `""`.
 
@@ -162,6 +161,8 @@ There's also an option to delete assets from CDF that are not referenced in the 
 | -------------------------- | --------|------------------------------------------------------------------------------------------------------------------ |
 | `deleteMissingAssets`      | `false` | Whether or not you would like assets under the root to be deleted if they're not present in the input data.       |
 | `ignoreDisconnectedAssets` | `false` | This will ignore assets that are not connected to root, as opposed to throwing an error and stopping the program. |
+| `allowMultipleRoots`       | `true`  | Whether it's allowed to ingest more than one (sub)tree. |
+| `allowSubtreeIngestion`    | `true`  | Whether ingestion of trees without a root node is allowed. Otherwise, an error is raised or the items are ignored when `ignoreDisconnectedAssets` is set |
 | `batchSize`                | 1000    | The number of assets to write per API call.                                                                       |
 
 ### Example
--- a/src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
+++ b/src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
@@ -52,8 +52,6 @@ private case class AssetSubtree(
     nodes: Array[AssetsIngestSchema]
 )
 
-final case class MultipleRootsException(roots: Seq[String])
-    extends Exception(s"Tree has more than one root: ${roots.mkString(", ")}")
 final case class NoRootException(
     message: String = """Tree has no root. Set parentExternalId to "" for the root Asset.""")
     extends Exception(message)
@@ -80,9 +78,7 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
   private val batchSize = config.batchSize.getOrElse(Constants.DefaultBatchSize)
 
   private val deleteMissingAssets = config.deleteMissingAssets
-  private val ignoreDisconnectedAssets = config.ignoreDisconnectedAssets
-  private val allowMultipleRoots = config.allowMultipleRoots
-  private val allowSubtreeIngestion = config.allowSubtreeIngestion
+  private val subtreeMode = config.subtrees
 
   def build(df: DataFrame): IO[Unit] = {
     val sourceTree = df.collect.map(r => fromRow[AssetsIngestSchema](r))
@@ -91,13 +87,13 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
       validateAndOrderInput(sourceTree)
         .sortBy(_.root.externalId) // make potential errors deterministic
 
-    val subtrees2 =
-      if (allowSubtreeIngestion) {
+    val subtrees2 = subtreeMode match {
+      case AssetSubtreeOption.Ingest =>
         subtrees
-      } else if (ignoreDisconnectedAssets) {
+      case AssetSubtreeOption.Ignore =>
         // only take trees with proper roots
         subtrees.filter(t => t.root.parentExternalId.isEmpty)
-      } else {
+      case AssetSubtreeOption.Error =>
         val nonRoots = subtrees.filter(t => t.root.parentExternalId.nonEmpty)
         if (nonRoots.nonEmpty) {
           throw InvalidTreeException(
@@ -106,16 +102,12 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
           )
         }
         subtrees
-      }
+    }
 
     if (subtrees.nonEmpty && subtrees2.isEmpty) {
       // in case all nodes are dropped due to ignoreDisconnectedAssets
       throw NoRootException()
       // we can allow empty source data set, that won't really happen by accident
-    }
-
-    if (!allowMultipleRoots && subtrees2.size > 1) {
-      throw MultipleRootsException(subtrees2.map(_.root.externalId))
     }
 
     for {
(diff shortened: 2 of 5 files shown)


src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -118,112 +196,173 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
...
 
-  def getCdfRootAndTree(sourceRootExternalId: CogniteExternalId): IO[(Option[Asset], List[Asset])] = {
-    val cdfTree = client.assets
-      .filter(AssetsFilter(rootIds = Some(Seq(sourceRootExternalId))))
-      .compile
-      .toList
+  def fetchCdfRoot(sourceRootExternalId: String): IO[Option[Asset]] =
+    client.assets
+      .retrieveByExternalId(sourceRootExternalId)
+      .map(Some(_).asInstanceOf[Option[Asset]])
```
sebbert: Use a type ascription here `Some(_): Option[Asset]` to avoid a runtime cast, or use `_.some` from cats

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -118,112 +196,173 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
...
+      .handleError {
+        case e: CdpApiException if e.code == 400 && e.missing.isDefined =>
+          None.asInstanceOf[Option[Asset]]
```
sebbert:
```suggestion
          None: Option[Asset]
```

README.md
```diff
@@ -139,15 +139,14 @@ Note: The asset hierarchy builder is currently in beta, and has not been suffici
...
+using the Spark Data Source.
+The asset hierarchy builder can ingest entire hierarchies of nodes connected
 through the `externalId`/`parentExternalId` relationship. If input contains an update to data that already exists,
 i.e there's a match on `externalId` and there's a change to one of the other fields, the asset will be updated.
 There's also an option to delete assets from CDF that are not referenced in the input data.
 
 ### Requirements
-- The source data must contain a single root, denoted by setting its `parentExternalId` to the empty string `""`.
-- All assets, except root, must be connected to another asset in the input data by `parentExternalId`.
+- The root is denoted by setting its `parentExternalId` to the empty string `""`.
```
sebbert: If we now support multiple roots, maybe change this to

- Root assets are denoted by setting their `parentExternalId` to the empty string `""`.

wjoel:
I'm a bit troubled by the very many options we're ending up with for asset hierarchy ingestion. Do we really need to support all of them? @hakontro @linda-cognite @BergsethCognite

BergsethCognite:
So what are all the options we have today:
- SDK (need this one)
- Jetfire (need this one)
-Spark Connector (not sure if we need this one)?

- any more?

linda-cognite:
deleteMissingAssets - Idk, I haven't had a case where I needed it, but I can see that it might be needed

ignoreDisconnectedAssets - I will always set this to true. I would like a warning containing the assets that were not ingested due to this. (In Jetfire I would actually love, if you would create an assets-error table and put those there.)

allowMultipleRoots - This alone doesn't make much sense to me, because the alternative is not complete.  As is, I'd always set it to true. If we had an alternative, where I specify the root node I care about and you ignore the rest, I would decide case by case.

allowSubtreeIngestion - I'm not sure about this one. I guess it is to add a subtree to an existing tree? We can ask people to union their tables first, right?

wjoel:
> So what are all the options we have today:
> 
>     * SDK (need this one)
> 
>     * Jetfire (need this one)
>       -Spark Connector (not sure if we need this one)?
> 
>     * any more?

@BergsethCognite I meant the multiple options added in this pull request, as listed in the description. We now have the following options for `assethierarchy`: ignoreDisconnectedAssets, allowMultipleRoots, allowSubtreeIngestion, deleteMissingAssets.

exyi:
> `deleteMissingAssets` - Idk, I haven't had a case where I needed it, but I can see that it might be needed

I mean it's already implemented, but removing this option (and seeting to `false` always), we could do upserts in a better way and maybe it could be faster (no need to download the CDF :smile:)

> allowMultipleRoots - This alone doesn't make much sense to me, because the alternative is not complete. As is, I'd always set it to true. If we had an alternative, where I specify the root node I care about and you ignore the rest, I would decide case by case.

That's right. This option only exists to cause errors when you have more root than expected. I wanted to preserve the tests we had for that :)

> allowSubtreeIngestion - I'm not sure about this one. I guess it is to add a subtree to an existing tree? We can ask people to union their tables first, right?

Yea, it's for updating/inserting a subtree. You can do unions, but this way does not need to download all the assets and should be easier to use. Also I guess simple union is not gonna be enough for updates at it would fail at duplicate entries.

hakontro:
> > `deleteMissingAssets` - Idk, I haven't had a case where I needed it, but I can see that it might be needed
> 
> I mean it's already implemented, but removing this option (and seeting to `false` always), we could do upserts in a better way and maybe it could be faster (no need to download the CDF )

For the record this was requested by @kjetilh33, and is apparently needed for a quite common way to work with asset hierarchies:
Re-ingest an already existing tree, a) do updates to existing assets, b) insert new assets, c) delete assets that are present in the old tree, but not in the new tree.

exyi:
> For the record this was requested by @kjetilh33
> <https://github.com/kjetilh33>, and is apparently needed for a quite
> common way to work with asset hierarchies:
> Re-ingest an already existing tree, a) do updates to existing assets, b)
> insert new assets, c) delete assets that are present in the old tree,
> but not in the new tree.
Ok, we have the use case so we can keep it as it is :)

kjetilh33:
> > > `deleteMissingAssets` - Idk, I haven't had a case where I needed it, but I can see that it might be needed
> > 
> For the record this was requested by @kjetilh33, and is apparently needed for a quite common way to work with asset hierarchies:
> Re-ingest an already existing tree, a) do updates to existing assets, b) insert new assets, c) delete assets that are present in the old tree, but not in the new tree.

Yes, we need to be able to remove/delete assets. If not, the hierarchy would "drift" and no longer be an accurate representation of the source.

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -118,112 +196,173 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
...
-            .updateByExternalId(
-              Seq(
-                root.externalId -> AssetsIngestSchema
-                  .toAssetUpdate(root)
-                  .copy(parentExternalId = None)).toMap)
-            .flatTap(x => incMetrics(itemsUpdated, x.size)) *> IO.unit
+          .create(Seq(AssetsIngestSchema.toAssetCreate(root).copy(parentExternalId = parentId)))
+          .flatTap(x => incMetrics(itemsCreated, x.size))
+          .map(x => x.head)
+      case Some(asset) => {
```
hakontro: Brackets can be omitted here as well.

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -118,112 +196,173 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
...
       sourceRoot: Option[Asset]
-  ): IO[Unit] =
+  ): IO[Unit] = {
+    val parentId = if (root.parentExternalId.isEmpty) {
+      None
+    } else {
+      Some(root.parentExternalId)
+    }
+
+    // TODO: don't do the update when nothing is changed (waiting for Scala SDK parentExternalId)
```
hakontro: Looks like this can be implemented now that `1.2.2` of the Scala SDK is out!

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -118,112 +196,173 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
...
+          .map(x => x.head)
+      case Some(asset) => {
+        val parentIdUpdate = (parentId, asset.parentId) match {
+          case (None, Some(oldParent)) =>
+            throw CdfDoesNotSupportException(
+              s"Can not make root from asset '${asset.externalId.get}' which is already inside asset '${oldParent}."
+              // TODO: include parentExternalId in the message
+                + " You might need to remove the asset and insert it again.")
+          case (None, None) => None
+          case (Some(newParent), _) => Some(SetValue(newParent))
```
hakontro: `if (newParent == asset.parentExternalId)`..

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -118,112 +196,173 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
...
+  def fetchCdfSubtree(root: AssetsIngestSchema): IO[List[Asset]] =
+    if (root.parentExternalId.isEmpty) {
+      // proper root
+      client.assets
+        .filter(AssetsFilter(rootIds = Some(Seq(CogniteExternalId(root.externalId)))))
+        .compile
+        .toList
+    } else {
+      // This API will error out if the subtree contains more than 100,000 items
+      // If that's a problem, we'll have to workaround by downloading it's entire hierarchy
```
hakontro:
```suggestion
      // The API will error out if the subtree contains more than 100,000 items
      // If that's a problem, we'll have to do a workaround by downloading its entire hierarchy
```

src/test/scala/cognite/spark/v1/AssetHierarchyBuilderTest.scala
```diff
@@ -105,8 +117,80 @@ class AssetHierarchyBuilderTest extends FlatSpec with Matchers with SparkTest {
...
+      AssetCreate("testNode1", None, None, Some(testName),Some("testNode1"), None, Some("nonExistentNode-jakdhdslfskgslfuwfvbnvwbqrvotfeds")),
+      AssetCreate("testNode2", None, None, Some(testName),Some("testNode2"), None, Some(""))
+    )
+
+    val ex = intercept[Exception] { ingest(tree) }
+    ex shouldBe an[InvalidNodeReferenceException]
+    ex.getMessage shouldBe "Node 'nonExistentNode-jakdhdslfskgslfuwfvbnvwbqrvotfeds' referenced from 'testNode1' does not exist."
+  }
+
+  it should "fail when subtree is updated into being root" in {
```
hakontro: Did they explain why this isn't allowed? It's something I suppose people might want to do, even if it's strange

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
exyi: You mean I should only do the update when it actually changes? That makes sense, however we don't do that for all the other fields, so we should do there too?

hakontro: That is the overall intention - assets that exist, but haven't changed, don't need to be updated. Maybe you can reuse `isMostlyEqual` here, or have it include `parentExternalId`

src/test/scala/cognite/spark/v1/AssetHierarchyBuilderTest.scala
exyi: I don't really know, the API simply does not allow you to specify `setNull` to `parentId`.

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
exyi: > Maybe you can reuse isMostlyEqual here

Thats already the handled in the `case` above. I was just waiting for the `parentExternalId` and then forgot to push yesterday :/

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -41,53 +45,127 @@ object AssetsIngestSchema {
...
+    extends Exception(s"Asset tree contains asset '$id' multiple times.")
 final case class EmptyExternalIdException(message: String = s"ExternalId cannot be an empty String.")
     extends Exception(message)
+final case class CdfDoesNotSupportException(message: String) extends Exception(message)
+final case class InvalidNodeReferenceException(nodeIds: Seq[String], referencedFrom: Seq[String])
+    extends Exception({
+      val plural = (sg: String, pl: String) => if (nodeIds.length == 1) sg else pl
+      val nodes = nodeIds.map(x => s"'$x'").mkString(", ")
+      val refNodes = referencedFrom.map(x => s"'$x'").mkString(", ")
+      s"${plural("Node", "Nodes")} $nodes referenced from $refNodes ${plural("does", "do")} not exist."
```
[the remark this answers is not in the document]
exyi: Good point! I have changed "Node" to "Parent"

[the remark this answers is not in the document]
hakontro: Nice, but looks like will print either "Parent nodes" or "Parents nodes" as opposed to "parent node/parent nodes" now, should probably move that around a bit!

PR #402, opened 2020-03-10
Adjust datapoints reads for exlusive endpoint

hakontro:


src/main/scala/cognite/spark/v1/NumericDataPointsRdd.scala
```diff
@@ -95,7 +95,7 @@ case class NumericDataPointsRdd(
       case _ =>
         countStart match {
           case Some(start) =>
-            val end = countEnd.getOrElse(start.plus(granularity.amount, granularity.unit))
+            val end = countEnd.getOrElse(start.plus(granularity.amount, granularity.unit)).plusMillis(1)
```
wjoel: The comment about this being exclusive is helpful, as it's not obvious why we're doing `.plusMillis(1)` otherwise.

src/main/scala/cognite/spark/v1/NumericDataPointsRdd.scala
```diff
@@ -337,7 +337,9 @@ case class NumericDataPointsRdd(
       .parEvalMapUnordered(50) {
         case (id, Some(first), Some(latest)) =>
           val aggStart = Instant.ofEpochMilli(floorToNearest(first.toEpochMilli, granularityMillis))
-          val aggEnd = Instant.ofEpochMilli(ceilToNearest(latest.toEpochMilli, granularityMillis))
+          val aggEnd = Instant
+            .ofEpochMilli(ceilToNearest(latest.toEpochMilli, granularityMillis))
+            .plusMillis(1) //endpoint is exclusive
```
wjoel: `// end is exclusive` (to match the name `end`) in https://docs.cognite.com/api/v1/#operation/getMultiTimeSeriesDatapoints

src/main/scala/cognite/spark/v1/NumericDataPointsRdd.scala
hakontro: Good catch!

hakontro:
> Those things are really tricky, but I trust you know what you're doing. I wonder though, how did we find this mistake? What are the consequences for the users? We should include a description of the latter in the change log.

Tricky indeed! We found this because aggregates weren't returning the very last data point, but it's hard to find because it only applies when the last data point is the only one in its aggregation interval. The result for the users is having a wrong aggregate, or entirely missing aggregate, for the final interval.

src/main/scala/cognite/spark/v1/NumericDataPointsRdd.scala
```diff
@@ -95,7 +95,9 @@ case class NumericDataPointsRdd(
       case _ =>
         countStart match {
           case Some(start) =>
-            val end = countEnd.getOrElse(start.plus(granularity.amount, granularity.unit))
+            val end = countEnd
+              .getOrElse(start.plus(granularity.amount, granularity.unit))
+              .plusMillis(1) //Adding 1 millisecond as end is exclusive
```
wjoel: Missing space after `//` ;)

PR #403, opened 2020-03-11
[DT-1449] Initial implementation of performance benchmarking

Tapped:
This is just the skeleton of the performance benchmarking application. Packaging and deployment is not implemented yet, and there are no way of configuring api url, number of items to write / read etc.


build.sbt
```diff
@@ -7,8 +7,10 @@ val sttpVersion = "1.6.3"
 val Specs2Version = "4.2.0"
 val artifactory = "https://cognite.jfrog.io/cognite/"
 val cogniteSdkVersion = "1.2.1"
+val prometheusVersion = "0.8.1"
+val log4sVersion = "1.8.2"
 
-resolvers += "libs-release" at artifactory + "libs-release/"
+resolvers += "libs-release".at(artifactory + "libs-release/")
```
wjoel: Not sure if it's your IntelliJ or if we need to change some config file to exclude build.sbt, but for better or worse this just isn't the formatting used for SBT files, and it's inconsistent with most of the rest of the file. The same applies for several other changes in this diff.

performancebench/src/main/scala/cognite/spark/performancebench/Metrics.scala
```diff
@@ -0,0 +1,13 @@
+package cognite.spark.performancebench
+
+import io.prometheus.client.Summary
+
+object Metrics {
+  // We don't really know the buckets, so summary is a better metric type than histogram
```
wjoel: If we only run the tests once, maybe we can use a counter and measure the number of seconds each test takes? If we run them once every once in a while, a gauge describing the time it took for the last test run might be appropriate.

performancebench/src/main/scala/cognite/spark/performancebench/Metrics.scala
```diff
@@ -0,0 +1,13 @@
...
+  val testTimeSummary = Summary
+    .build()
+    .name("cognite_cdf_spark_test_seconds")
+    .labelNames("testName")
```
wjoel: Prometheus labels are snake_cased, so "test_name" instead.

performancebench/src/main/scala/cognite/spark/performancebench/PerformanceSuite.scala
```diff
@@ -0,0 +1,39 @@
...
+      val testResult = for {
+        beforeTestResult <- Try(perfTest.beforeTest())
+        testResult <- {
+          logger.info(s"${perfTest.testName}: starting")
+          val startTime = System.currentTimeMillis()
+          val res = Try(
+            Metrics.testTimeSummary
+              .labels(perfTest.testName)
+              .time(new Runnable() { def run() = perfTest.test(beforeTestResult) }))
+          logger.info(s"${perfTest.testName}: finished after ${(System.currentTimeMillis() - startTime) / 1000}")
```
wjoel: Better to include "seconds" (or "s", or "sec") in the log message as well.

build.sbt
Tapped: Yes saw it. I thought maybe this was something you wanted since we don't ignore sbt files from formatting. We should exclude it from scalafmt.

performancebench/src/main/scala/cognite/spark/performancebench/Metrics.scala
Tapped: Yes, so I've not decided yet if I want this as a cron job or as a deployment. So it depends a bit, one can specify the sliding window of summary as well.

performancebench/src/main/scala/cognite/spark/performancebench/PerformanceSuite.scala
Tapped: I agree 👍

performancebench/src/main/scala/cognite/spark/performancebench/Metrics.scala
Tapped: Good catch

performancebench/src/main/scala/cognite/spark/performancebench/Metrics.scala
Tapped: I went for the gauge method, so that we got more flexibility.

build.sbt
Tapped: Okey, it was just me using a new intellij version, and had to set scalafmt as formatter again.

performancebench/src/main/scala/cognite/spark/performancebench/PerformanceSuite.scala
```diff
@@ -0,0 +1,42 @@
...
+            .set(duration)
+          val durationAsString = f"$duration%1.2f"
+          logger.info(s"${perfTest.testName}: finished after ${durationAsString} secs")
+          res
+        }
+      } yield testResult
+
+      testResult match {
+        case Success(_) => ()
+        case Failure(exception) => logger.error(exception)(s"${perfTest.testName}: failed to run")
```
wjoel: Could be useful to have a counter for those failures.

Tapped:
See 116, it's not published anywhere.

Tapped:
I think it's easiest if it's part of the repo, as having it's close to the rest of the code makes it easier to update etc. We will just deploy it from our Jenkinsfile.

Tapped:
@wjoel Added a counter for failed tests

PR #404, opened 2020-03-12
[DT-1378] Add support for ignoreUnknownIds to timeseries

sush03:


src/test/scala/cognite/spark/v1/TimeSeriesRelationTest.scala
```diff
@@ -936,6 +936,83 @@ class TimeSeriesRelationTest extends FlatSpec with Matchers with SparkTest with
...
+      .insertInto("destinationTimeSeries")
+
+    // Check if insert worked
+    val idsAfterInsert =
+      retryWhile[Array[Row]](
+        spark
+          .sql(s"select id from destinationTimeSeries where unit = '$unit'")
+          .collect,
+        df => df.length < 1)
+    assert(idsAfterInsert.length == 1)
```
exyi: It seems to me that you don't need to insert the test data at all.

sush03: I tried following the format used for Assets and events deletion (It had insertion of test data following the deletes).
Also, I don't know why "timeseries not found " error ie I guess error 400 is not thrown if there's no insertion.

wjoel: The tests in assets and events for ignoreUnknownIds are not very good, it would be useful if they tested a delete on the created id+unknown id (which should fail without the option, and leave created id present), then did a delete on with ignoreUnknownIds set (which should succeed, and the created id should be gone), but they're not doing that.

You can just keep the other tests (attempt to delete a non-existing id with ignoreUnknownIds should be ok, should throw an exception without the setting). You can change the non-existing id to something more obviously non-existing, like 123.

Other than that, this PR looks good to me, assuming it's using the correct SDK version.

exyi:
On 2020-03-12 16:11, Joel Wilsson wrote:
> The tests in assets and events for ignoreUnknownIds are not very good,
> it would be useful if they tested a delete on the created id+unknown id
> (which should fail without the option, and leave created id present),
> then did a delete on with ignoreUnknownIds set.

Is is however not how it works (AFAIK, at least), if the delete
operation fails, the destiny of the valid deletes is undefined. The
delete operation is performed in batches and in parallel. When one
fails, the already deleted items are not going to re-appear.

wjoel:
> On 2020-03-12 16:11, Joel Wilsson wrote: The tests in assets and events for ignoreUnknownIds are not very good, it would be useful if they tested a delete on the created id+unknown id (which should fail without the option, and leave created id present), then did a delete on with ignoreUnknownIds set.
> Is is however not how it works (AFAIK, at least), if the delete operation fails, the destiny of the valid deletes is undefined. The delete operation is performed in batches and in parallel. When one fails, the already deleted items are not going to re-appear.

Good point! In the general case you're right, the deletes could end up being executed by different executors, but in the test case they will end up on the same executor (not sure if it's guaranteed, or just incredibly likely) and sent in the same request, in which case the API does guarantee that no change will be made to the destination data.

src/test/scala/cognite/spark/v1/TimeSeriesRelationTest.scala
sush03: What I understand from the conversation that the deletes are performed by different executors and hence performing deletes with (created + unknown)id with and without (ie twice) unknown id setting won't make sense because the created id would be already deleted in the first go, right?

Also, I guess I have not updated the SDK version to the one which would be released now ( I guess we should release the updated scala SDK)

hakontro:
Looks like these tests are failing because the new Scala SDK version has added the field `parentExternalId` to the `AssetCreate` class. This, in turn, requires `parentExternalId` to be present in the SQL statements we use in our tests.

wjoel:
Thanks, I'll make that change. Maybe we should just change them, or at least most of them, to use `save` instead of `insertInto`.

PR #405, opened 2020-03-13
[DT-1412] Support externalId for upserts in assets, events and timeseries

sush03:


src/test/scala/cognite/spark/v1/TimeSeriesRelationTest.scala
```diff
@@ -936,6 +936,83 @@ class TimeSeriesRelationTest extends FlatSpec with Matchers with SparkTest with
...
+      .insertInto("destinationTimeSeries")
+
+    // Check if insert worked
+    val idsAfterInsert =
+      retryWhile[Array[Row]](
+        spark
+          .sql(s"select id from destinationTimeSeries where unit = '$unit'")
+          .collect,
+        df => df.length < 1)
+    assert(idsAfterInsert.length == 1)
```
wjoel: I think we can remove all of this. Just makes the tests slower, and isn't really used nor necessary to test this functionality.

sush03:
> Tests aren't passing with this change.

They are just events and datapoints test right ?

wjoel:
@sush03 yes, but the events failures are not due to inconsistency ("simple reads" might be, that one can be ignored), but something else, and so is the data points test failing as far as I can tell.

PR #406, opened 2020-03-13
[DT-1450] Deploy performance bench image

Tapped:


project/plugins.sbt
```diff
@@ -6,4 +6,5 @@ addSbtPlugin("com.thesamet" % "sbt-protoc" % "0.99.19")
 addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.1.1")
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.9.0")
 addSbtPlugin("au.com.onegeek" %% "sbt-dotenv" % "2.1.146")
+addSbtPlugin("com.typesafe.sbt" % "sbt-native-packager" % "1.3.15")
```
wjoel: Latest release is 1.7.0 (13 hours old!) with a very nice feature to layer the application, let's give it a go?

Tapped: So the docker layering cause issues for us (at least before) since `docker:stage` needs to know which docker version to target, unless docker is installed which it isn't in the sbt jenkins image. However we can work around it by setting the docker version explicitly in Jenkins.

Tapped: I guess it's called docker multi-stage actually

Tapped: Okey, right in 1.7.0, they have fixed the requirement of setting a docker version. It will default to 1.2.15. But last week when I checked it had to be explicitly set, if docker wasn't installed.

Jenkinsfile
```diff
@@ -3,39 +3,57 @@
...
+                    ttyEnabled: true,
+                    command: '/bin/cat -'
+                ),
+                containerTemplate(
+                    name: 'docker',
+                    image: 'docker:19.03',
+                    command: '/bin/cat -',
+                    resourceRequestCpu: '200m',
+                    resourceLimitCpu: '2000m',
+                    resourceLimitMemory: '300Mi',
```
wjoel: Possibly a bit too low? Can use more, no problem.

Tapped: Maybe, this is the setting we use other places and seems to work fine

PR #410, opened 2020-03-17
Drain the read queue on separate thread

hakontro:


src/main/scala/cognite/spark/v1/SdkV1Rdd.scala
```diff
@@ -43,6 +43,8 @@ case class SdkV1Rdd[A, I](
 
   override def compute(_split: Partition, context: TaskContext): Iterator[Row] = {
     val split = _split.asInstanceOf[CdfPartition]
+    val drainPool =
```
Tapped: Lets make a comment here, to make it clear why we need this. Also add shutdown of this threadpool

PR #413, opened 2020-03-19
Close spark bench when done

Tapped:


performancebench/src/main/scala/cognite/spark/performancebench/Main.scala
```diff
@@ -11,7 +11,7 @@ object Main extends App {
     eventPerf.run()
   } finally {
     logger.info("Sleeping for 20 seconds before exiting, so that all metrics are scraped.")
-    Thread.sleep(20*1000) /* Sleep for 20 seconds to make sure metrics are scraped */
-    metricsServer.stop()
+    Thread.sleep(20*1000) // Sleep for 20 seconds to make sure metrics are scraped
```
wjoel: Shut down metricsServer as well?

Tapped: Nah, we just let the OS clean that up. Same procedure as Chrome, that's how they get the fast shutdown speed. I've also done `fork := true` so this works fine in sbt as well.

wjoel: It's still good practice to clean up the resources you create, and I think it would be better to fix the issue with background threads rather than doing a hard exit, but alright.

Tapped:
Yes I'll get another PR in after daemon PR is merged (PR not created yet)

PR #414, opened 2020-03-20
Change Upsert method (it should try to update and then insert if nece…

sush03:
…ssary)


sush03:
https://github.com/cognitedata/cdp-spark-datasource/blob/Upsert-method/src/test/scala/cognite/spark/v1/AssetsRelationTest.scala#L271

So, we won't be able to support upserting through insertInto (There are some tests in each resource like this)

Also, I'm not quite sure how to include the legacyName check !

PR #416, opened 2020-03-20
Report test failures to Jenkins

Tapped:
The test failures will then be easier accessible through the Jenkins UI.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -68,7 +68,11 @@ podTemplate(label: label,
                     if (env.BRANCH_NAME == 'master') {
                         test = "+test"
                     }
-                    sh("sbt -Dsbt.log.noformat=true scalastyle scalafmtCheck coverage $test coverageReport")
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -68,7 +68,11 @@ podTemplate(label: label,
                     if (env.BRANCH_NAME == 'master') {
                         test = "+test"
                     }
-                    sh("sbt -Dsbt.log.noformat=true scalastyle scalafmtCheck coverage $test coverageReport")
+                    testStatus = sh(returnStatus: true, "sbt -Dsbt.log.noformat=true scalastyle scalafmtCheck coverage $test coverageReport")
+                    junit(allowEmptyResults: false, testResults: '**/target/test-reports/*.xml')
+                    if (testStatus != 0) {
+                        error("Tests failed")
+                    }
```
wjoel: Cool, maybe we also want
```
} else {
  summarizeTestResults()
}
```
so we don't have to open the Jenkins UI as often.

PR #419, opened 2020-03-20
[DT-1412] Make asset name optional for asset update via upsert

wjoel:


wjoel:
@Tapped We added `it should "be possible to make an upsert on externalId without name"` to do that.

Tapped:
Ah sorry, I missed the file as it was collapsed by github since the diff was to big.

PR #420, opened 2020-03-24
Add warmup logic to perf. bench

Tapped:
Read some events to warmup the Spark cluster, since without warmup, we get biased metric results.

### DIFF
--- a/performancebench/src/main/scala/cognite/spark/performancebench/Main.scala
+++ b/performancebench/src/main/scala/cognite/spark/performancebench/Main.scala
@@ -9,6 +9,8 @@ object Main extends App {
   val metricsServer = new HTTPServer(8123, true)
   PrometheusDefaultExports.initialize()
   try {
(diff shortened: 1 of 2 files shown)


performancebench/src/main/scala/cognite/spark/performancebench/Main.scala
```diff
@@ -1,12 +1,16 @@
...
+import io.prometheus.client.hotspot.{DefaultExports => PrometheusDefaultExports}
 import org.log4s._
 
 object Main extends App {
   val logger = getLogger
   val metricsServer = new HTTPServer(8123, true)
+  PrometheusDefaultExports.initialize()
   try {
+    // We need to warm-up our Spark cluster to get fair metrics
+    Warmup.run()
```
wjoel: Isn't it a bit overkill to make this an object instead of just a local method?

Tapped: Probably 🤷‍♂

PR #422, opened 2020-03-30
Gzip compression

wjoel:


src/main/scala/cognite/spark/v1/SdkV1Rdd.scala
```diff
@@ -81,33 +71,27 @@ case class SdkV1Rdd[A, I](
...
-      // Ensure that only one chunk is processed at a time
-      s <- stream.chunks.flatMap(c => Stream.eval(singleThreadedCs.shift *> IO.pure(c)))
-      put <- {
-        // Filter out and keep track of already seen items
-        // since overlapping pushdown filters may read duplicates.
-        val freshIds = s.toVector.filterNot(i => processedIds.contains(uniqueId(i)))
-        processedIds ++= freshIds.map(i => uniqueId(i))
-        Stream.eval(IO(queue.put(Right(freshIds))))
+      processedIds: ConcurrentHashMap[I, Unit]): Stream[IO, Unit] =
+    stream.chunks.parEvalMapUnordered(config.parallelismPerPartition * 2) { chunk =>
```
Tapped: Nice this should work fine!

src/main/scala/cognite/spark/v1/SdkV1Rdd.scala
```diff
@@ -81,33 +71,27 @@ case class SdkV1Rdd[A, I](
...
+      val freshIds = chunk.toVector.filterNot(i => processedIds.containsKey(uniqueId(i)))
```
Tapped: I know we did a `toVector` before, however I'm unsure if that's needed, and only ends up being an unnecessary copy operation (though only a 10k * IntPtr.size).

wjoel: Good point, it's easy to work with Chunks directly.

src/main/scala/cognite/spark/v1/SdkV1Rdd.scala
wjoel: It does! :) Unfortunately I haven't been able to prove any performance benefit on my machine, but it's also not worse, and the code is shorter and easier to understand.

PR #424, opened 2020-04-01
Stream data points and generalize Stream to Iterator

wjoel:


src/main/scala/cognite/spark/v1/StreamIterator.scala
```diff
@@ -0,0 +1,105 @@
...
+import cats.effect.{Concurrent, IO}
+import fs2.{Chunk, Stream}
+import org.apache.spark.sql.Row
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object StreamIterator {
+  type EitherQueue = ArrayBlockingQueue[Either[Throwable, Chunk[Row]]]
+
+  def apply(stream: Stream[IO, Row], queueBufferSize: Int, toId: Option[Row => String] = None)(
```
Tapped, line 14: `queueBufferSize` = `numBatchesInQueue` so it should be closer to the number of batches running at any time.

wjoel: @Tapped not sure if I understood what you meant by this, maybe I'm doing it already, otherwise could you suggest what we should change it to?

src/main/scala/cognite/spark/v1/NumericDataPointsRdd.scala
```diff
@@ -422,54 +434,81 @@ case class NumericDataPointsRdd(
...
+    for (f <- rowIndices) {
+      f match {
+        case 0 => array(i) = id.left.toOption
+        case 1 => array(i) = id.right.toOption
+        case 2 => array(i) = java.sql.Timestamp.from(dataPoint.timestamp)
+        case 3 => array(i) = dataPoint.value
+        case 4 | 5 => array(i) = None
+      }
+      i += 1
+    }
```
sebbert, line 458:
```suggestion
    for ((f, i) <- rowIndices.zipWithIndex) {
      f match {
        case 0 => array(i) = id.left.toOption
        case 1 => array(i) = id.right.toOption
        case 2 => array(i) = java.sql.Timestamp.from(dataPoint.timestamp)
        case 3 => array(i) = dataPoint.value
        case 4 | 5 => array(i) = None
      }
    }
```

src/main/scala/cognite/spark/v1/NumericDataPointsRdd.scala
```diff
@@ -422,54 +434,81 @@ case class NumericDataPointsRdd(
...
+      f match {
+        case 0 => array(i) = r.id.left.toOption
+        case 1 => array(i) = r.id.right.toOption
+        case 2 => array(i) = java.sql.Timestamp.from(dataPoint.timestamp)
+        case 3 => array(i) = dataPoint.value
+        case 4 => array(i) = Some(r.aggregation)
+        case 5 => array(i) = Some(r.granularity.toString)
+      }
+      i += 1
+    }
```
sebbert, line 476:
```suggestion
    for ((f, i) <- rowIndices) {
      f match {
        case 0 => array(i) = r.id.left.toOption
        case 1 => array(i) = r.id.right.toOption
        case 2 => array(i) = java.sql.Timestamp.from(dataPoint.timestamp)
        case 3 => array(i) = dataPoint.value
        case 4 => array(i) = Some(r.aggregation)
        case 5 => array(i) = Some(r.granularity.toString)
      }
    }
```

src/main/scala/cognite/spark/v1/NumericDataPointsRdd.scala
wjoel: As far as I can that leads to less efficient code (which allocates tuples, and ). It's ugly, but it's meant to be very fast and efficient since it gets called hundreds of thousands of times (once for every data point).

From "Decompile Scala to Java" in IntelliJ, the first version is mine:
```java
   private Row toRowDataPointsRange(final Either id, final DataPoint dataPoint) {
      Object[] array = new Object[this.rowIndices().length];
      IntRef i = IntRef.create(0);
      (new ofInt(scala.Predef..MODULE$.intArrayOps(this.rowIndices()))).foreach((f) -> {
         switch(f) {
         case 0:
            array[i.elem] = id.left().toOption();
            break;
         case 1:
            array[i.elem] = id.right().toOption();
            break;
         case 2:
            array[i.elem] = Timestamp.from(dataPoint.timestamp());
            break;
         case 3:
            array[i.elem] = BoxesRunTime.boxToDouble(dataPoint.value());
            break;
         case 4:
         case 5:
            array[i.elem] = scala.None..MODULE$;
            break;
         default:
            throw new MatchError(BoxesRunTime.boxToInteger(f));
         }

         ++i.elem;
      });
      return new GenericRow(array);
   }

   private Row toRowDataPointsRange(final Either id, final DataPoint dataPoint) {
      Object[] array = new Object[this.rowIndices().length];
      (new ofRef(scala.Predef..MODULE$.refArrayOps((Object[])(new ofInt(scala.Predef..MODULE$.intArrayOps(this.rowIndices()))).zipWithIndex(scala.Array..MODULE$.canBuildFrom(scala.reflect.ClassTag..MODULE$.apply(Tuple2.class)))))).withFilter((check$ifrefutable$3) -> {
         return BoxesRunTime.boxToBoolean($anonfun$toRowDataPointsRange$1(check$ifrefutable$3));
      }).foreach((x$23) -> {
         $anonfun$toRowDataPointsRange$2(array, id, dataPoint, x$23);
         return BoxedUnit.UNIT;
      });
      return new GenericRow(array);
   }

   // $FF: synthetic method
   public static final void $anonfun$toRowDataPointsRange$2(final Object[] array$1, final Either id$5, final DataPoint dataPoint$1, final Tuple2 x$23) {
      if (x$23 != null) {
         int f = x$23._1$mcI$sp();
         int i = x$23._2$mcI$sp();
         BoxedUnit var10000;
         switch(f) {
         case 0:
            array$1[i] = id$5.left().toOption();
            var10000 = BoxedUnit.UNIT;
            break;
         case 1:
            array$1[i] = id$5.right().toOption();
            var10000 = BoxedUnit.UNIT;
            break;
         case 2:
            array$1[i] = Timestamp.from(dataPoint$1.timestamp());
            var10000 = BoxedUnit.UNIT;
            break;
         case 3:
            array$1[i] = BoxesRunTime.boxToDouble(dataPoint$1.value());
            var10000 = BoxedUnit.UNIT;
            break;
         case 4:
         case 5:
            array$1[i] = scala.None..MODULE$;
            var10000 = BoxedUnit.UNIT;
            break;
         default:
            throw new MatchError(BoxesRunTime.boxToInteger(f));
         }

      } else {
         throw new MatchError(x$23);
      }
   }
```

It's always hard to predict what the JVM will turn this into once it's JITted, but I'd prefer the stupid, simple, ugly and definitely fast version in this case. ;)

src/main/scala/cognite/spark/v1/SdkV1Rdd.scala
```diff
@@ -45,82 +45,22 @@ case class SdkV1Rdd[A, I](
...
-          processedIds.put(uniqueId(i), ())
-        }
-        queue.put(Right(freshIds))
+    val processChunk = Some((chunk: Chunk[A]) => {
+      chunk.filter { i =>
+        // Use .asInstanceOf[Object] to silence warning about
+        // "comparing values of types Unit and Null"
+        // putIfAbsent returns null if the key did not exist, in which ase we
+        // should keep (and process) the item.
+        processedIds.putIfAbsent(uniqueId(i), ()).asInstanceOf[Object] == null
```
sebbert, line 60: Alternatively, there's [Option.apply](https://www.scala-lang.org/api/current/scala/Option$.html#apply[A](x:A):Option[A]), which converts a nullable object into an Option

src/main/scala/cognite/spark/v1/SdkV1Rdd.scala
```diff
@@ -45,82 +45,22 @@ case class SdkV1Rdd[A, I](
...
-  def enqueueStreamResults(stream: Stream[IO, A], queue: EitherQueue): Stream[IO, Unit] = {
     val processedIds = new ConcurrentHashMap[I, Unit]
-    stream.chunks.parEvalMapUnordered(config.parallelismPerPartition) { chunk =>
-      val freshIds = chunk.filter(i => !processedIds.containsKey(uniqueId(i)))
-      IO {
-        freshIds.foreach { i =>
-          processedIds.put(uniqueId(i), ())
-        }
-        queue.put(Right(freshIds))
+    val processChunk = Some((chunk: Chunk[A]) => {
```
Tapped, line 54: I think I would like this `Some` at the usage site eg. line 63.

src/main/scala/cognite/spark/v1/NumericDataPointsRdd.scala
```diff
@@ -422,54 +434,81 @@ case class NumericDataPointsRdd(
...
-                      p.value,
-                      None,
-                      None)
-                  }
-                  .map(toRow)
-                  .toVector)
+    val stream: Stream[IO, Stream[IO, Row]] = Stream
+      .emits(bucket.ranges.toVector)
+      .covary[IO]
+      .parEvalMapUnordered(100) {
```
Tapped, line 492: Where does this constant comes from? Should it be dynamic?

src/main/scala/cognite/spark/v1/SdkV1Rdd.scala
```diff
@@ -45,82 +45,22 @@ case class SdkV1Rdd[A, I](
...
-    val streamsToQueue: Future[Unit] = Future {
-      putOnQueueStream.compile.drain.unsafeRunSync()
-    }(drainContext)
-
-    queueIterator(queue, streamsToQueue) {
-      if (!drainPool.isShutdown) {
-        drainPool.shutdown()
-      }
+    val currentStreamsAsSingleStream = config.limitPerPartition match {
+      case Some(limit) => groupedStreams(split.index).map(_.take(limit)).reduce(_.merge(_))
```
Tapped, line 49: Is this `take` really necessary, as it's applied to each "CDF partition", however we have also set the limit at line 37.

src/main/scala/cognite/spark/v1/SdkV1Rdd.scala
wjoel: Good point, that's nicer. Done.

src/main/scala/cognite/spark/v1/NumericDataPointsRdd.scala
wjoel: We really just want it to be limited only by thread pools etc., turns out parJoinUnbounded sets this to Int.MaxValue and parEvalMapUnordered is using parJoin. We should use Int.MaxValue here as well, and I've added comments to explain this.

src/main/scala/cognite/spark/v1/SdkV1Rdd.scala
wjoel: Nope! Good catch, that already take cares of limitPerPartition for SdkV1Rdd.

PR #426, opened 2020-04-03
[DT-1118]Fix null handling in filter pushdown

exyi:
Fixes queries like `where X in (null, "a", "b")` + I have added tests and asserts for other operators when using NULLs

### DIFF
--- a/src/main/scala/cognite/spark/v1/PushdownUtilities.scala
+++ b/src/main/scala/cognite/spark/v1/PushdownUtilities.scala
@@ -59,7 +59,11 @@ object PushdownUtilities {
   // scalastyle:off
   def getFilter(filter: Filter): PushdownExpression =
     filter match {
-      case IsNotNull(colName) => NoPushdown()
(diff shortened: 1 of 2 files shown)


src/main/scala/cognite/spark/v1/PushdownUtilities.scala
```diff
@@ -59,7 +59,11 @@ object PushdownUtilities {
   // scalastyle:off
   def getFilter(filter: Filter): PushdownExpression =
     filter match {
-      case IsNotNull(colName) => NoPushdown()
+      case IsNotNull(_) | IsNull(_) | EqualNullSafe(_, null) => NoPushdown()
+      case EqualTo(_, null) | GreaterThan(_, null) | GreaterThanOrEqual(_, null) | LessThan(_, null) |
+          LessThanOrEqual(_, null) =>
+        throw new Exception(
+          "Wait, this should not happen... This is always false and Spark should just optimize it away.")
```
wjoel, line 66: But just in case ;) "Unexpected error, please contact support@cognite.com and tell them Spark is misbehaving."
After all, a user *could* end up seeing this message!

PR #427, opened 2020-04-14
Asset hierarchy: Fix glitch with duplicate root parentExternalId

exyi:


src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -124,13 +124,14 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
...
       client.assets
-        .retrieveByExternalIds(ids)
+        .retrieveByExternalIds(ids.distinct)
         .adaptError({
           case e: CdpApiException if e.code == 400 && e.missing.nonEmpty =>
             val missingNodes = e.missing.get.map(j => j("externalId").get.asString.get).take(10)
             val referencingNodes =
               missingNodes
-                .flatMap(missing => roots.find(_.parentExternalId == missing))
+              // always take the one with "lowest" externalId, so the error are deterministic
```
hakontro, line 133:
```suggestion
              // always take the one with "lowest" externalId, so the errors are deterministic
```

PR #428, opened 2020-04-22
[DT-1127] Improve error message for column incorrect type

exyi:
We now explicitly handle the case when the column has an unexpected type.

We also do implicit conversions to Double and Long from other numeric types.

Minor improvements: shorter type names, hint for parentExternalId=NULL

### DIFF
--- a/macro/src/main/scala/SparkSchemaHelper.scala
+++ b/macro/src/main/scala/SparkSchemaHelper.scala
@@ -6,7 +6,7 @@ import org.apache.spark.sql.types.StructType
 import com.cognite.sdk.scala.common.{NonNullableSetter, Setter}
 
 import scala.reflect.macros.blackbox.Context
-import scala.util.{Try, Failure, Success}
+import scala.util.{Failure, Success, Try}
 
 class SparkSchemaHelperImpl(val c: Context) {
   def asRow[T: c.WeakTypeTag](x: c.Expr[T]): c.Expr[Row] = {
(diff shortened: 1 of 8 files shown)


exyi:
I also had to make some  to the tests (I mean the second commit), since some data are broken again... All were also needed on master to pass.

macro/src/main/scala/SparkSchemaHelper.scala
```diff
@@ -80,16 +80,54 @@ class SparkSchemaHelperImpl(val c: Context) {
...
 
-        val column = q"scala.util.Try(Option($r.getAs[$rowType]($name))).toOption.flatten"
+        val throwError =
+          q"throw cognite.spark.v1.SparkSchemaHelper.badRowError($r, $name, ${rowType.toString}, ${structType.toString})"
+
+        def handleFieldValue(value: Tree) =
+          if (rowType <:< typeOf[Double]) {
+            // do implicit conversion to double
+            q"""($value match {
+             case x:Double => Some(x)
```
wjoel, line 90: Not super important but there should be a space after the colon ;)

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
```diff
@@ -391,14 +391,15 @@ class AssetsRelationTest extends FlatSpec with Matchers with SparkTest {
         .write
         .insertInto("destinationAssetsUpsert")
 
+      val assetsCreated = getNumberOfRowsCreated(metricsPrefix, "assets")
```
wjoel, line 394: Do you have any idea why/if moving this check here makes a difference?

exyi: It does not fix the test, just asserts the easy to  assert and easy to debug stuff first.

PR #430, opened 2020-05-04
[CDF-5654] Check Map[String, String] (a metadata field) in fromRow

exyi:
the `fromRow` macro now checks that `Map[String, String]` contains only strings, ignores null values and trims the values to the metadata size limit.

### DIFF
--- a/macro/src/main/scala/SparkSchemaHelper.scala
+++ b/macro/src/main/scala/SparkSchemaHelper.scala
@@ -35,13 +35,13 @@ class SparkSchemaHelperImpl(val c: Context) {
     import c.universe._
     val optionType = typeOf[Option[_]]
     val optionSeq = typeOf[Seq[Option[_]]]
-    val optionMap = typeOf[Map[_, Option[_]]]
     val optionSetter = typeOf[Option[Setter[_]]]
     val optionNonNullableSetter = typeOf[Option[NonNullableSetter[_]]]
     val setterType = symbolOf[Setter.type].asClass.module
     val nonNullableSetterType = symbolOf[NonNullableSetter.type].asClass.module
     val seqAny = typeOf[Seq[Any]]
     val mapAny = typeOf[Map[Any, Any]]
+    val mapString = typeOf[Map[String, String]]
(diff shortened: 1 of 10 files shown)


macro/src/main/scala/SparkSchemaHelper.scala
```diff
@@ -60,12 +60,16 @@ class SparkSchemaHelperImpl(val c: Context) {
           param.typeSignature
         }
         val isSequenceWithOptionalVal = innerType <:< optionSeq
-        val isMapWithOptionalVal = innerType <:< optionMap
         val isSequenceOfAny = innerType <:< seqAny
-        val isMapOfAny = innerType <:< mapAny
+        val isMapOfString = innerType <:< mapString
+        if (!isMapOfString && innerType.getClass == classOf[Map[_, _]]) {
+          throw new Exception(s"Only Map[String, String] is supported, not $innerType")
```
wjoel, line 66: Isn't this excluding all other types of maps? It's possible to store a map<string, long> in a RAW column, for example, and that's allowed.

wjoel: Of course, this is for deriving the schema for our own types. :facepalm: All good then.

macro/src/main/scala/SparkSchemaHelper.scala
```diff
@@ -74,38 +78,53 @@ class SparkSchemaHelperImpl(val c: Context) {
...
+             case x: Int => Some(x.toDouble)
+             case x: Float => Some(x.toDouble)
+             case x: Long => Some(x.toDouble)
              case x: BigDecimal => Some(x.toDouble)
              case x: BigInt => Some(x.toDouble)
              case _ => $throwError
            })"""
-          } else if (rowType <:< typeOf[Long]) {
+          } else if (rowType == typeOf[Float]) {
+            // do implicit conversion to double
```
wjoel, line 101: Either the comment or the code is wrong, should probably be `x.toDouble`?

exyi: Ah, sorry, the comment is lying ...

PR #434, opened 2020-05-08
[CDF-5653] Add support for files metadata

sush03:


src/main/scala/cognite/spark/v1/FilesRelation.scala
```diff
@@ -1,35 +1,144 @@
...
+    id: Option[Long] = None,
+    externalId: Option[String] = None,
+    source: Option[String] = None,
+    metadata: Option[Map[String, String]] = None,
+    assetIds: Option[Seq[Long]] = None,
+    dataSetId: Option[Long] = None,
+    mimeType: Option[String] = None,
+    sourceCreatedTime: Option[Instant] = None,
+    sourceModifiedTime: Option[Instant] = None,
+    securityCategories: Option[Seq[Long]] = None,
```
wjoel, line 107: Remove the trailing comma, it's not allowed in Scala 2.11 and scalafmt doesn't like it.

src/test/scala/cognite/spark/v1/FilesRelationTest.scala
```diff
@@ -49,64 +50,82 @@ class FilesRelationTest extends FlatSpec with Matchers with SparkTest {
...
+            |'updated-externalId-$originalSource' as externalId
      """.stripMargin)
-      .select(df.columns.map(col): _*)
       .write
-      .insertInto("filesSource")
+      .format("cognite.spark.v1")
+      .option("apiKey", writeApiKey)
+      .option("type", "files")
+      .option("onconflict", "upsert")
+      .save()
```
wjoel, line 90: We should still have at least some tests that still use `.insertInto`, as that's a different interface and uses different code. Better to just add a separate test for `"onconflict", "upsert"`, I think.

src/test/scala/cognite/spark/v1/FilesRelationTest.scala
```diff
@@ -1,20 +1,35 @@
...
+  destinationDf.createOrReplaceTempView("destinationFiles")
+
+  "FilesRelation" should "read files" taggedAs ReadTest in {
+    val res = spark.sqlContext
+      .sql("select * from sourceFiles")
+      .collect()
     // as of 2019-02-19 there are 11 files, but more might be added in the future,
     // which should not be allowed to break this test.
-    assert(df.count() >= 11)
+    assert(res.length == 18)
```
wjoel, line 32: Either update the comment or change the select to filter on creation time.

src/test/scala/cognite/spark/v1/FilesRelationTest.scala
```diff
@@ -39,74 +55,300 @@ class FilesRelationTest extends FlatSpec with Matchers with SparkTest {
...
+        .save()
 
-    val dfWithUpdatedSource = retryWhile[List[File]](
-      writeClient.files.filter(FilesFilter(assetIds = Some(Seq(176631399357975L)))).compile.toList,
-      files => files.length < 10)
-    assert(dfWithUpdatedSource.length == 10)
+      val filesCreated2 = getNumberOfRowsCreated(metricsPrefix, "files")
+      assert(filesCreated2 == 1)
+      var updatedFiles = getNumberOfRowsUpdated(metricsPrefix, "files")
+      assert(updatedFiles == 1)
```
wjoel, line 164: Don't use `var`, but if only checking the value once and then not using it again, you can do `assert(getNumberOfRowsUpdated(metricsPrefix, "files") == 1)`

src/test/scala/cognite/spark/v1/FilesRelationTest.scala
```diff
@@ -39,74 +55,300 @@ class FilesRelationTest extends FlatSpec with Matchers with SparkTest {
...
+                |null as assetIds,
+                |null as datasetId,
+                |null as sourceCreatedTime,
+                |null as sourceModifiedTime,
+                |null as securityCategories,
+                |null as uploaded,
+                |null as createdTime,
+                |null as lastUpdatedTime,
+                |null as uploadedTime,
+                |null as uploadUrl
```
wjoel, line 304: Can just use `write...save()` instead, then we won't have to specify nulls like this.

src/test/scala/cognite/spark/v1/FilesRelationTest.scala
```diff
@@ -39,74 +55,300 @@ class FilesRelationTest extends FlatSpec with Matchers with SparkTest {
...
```
wjoel, line 129: Can just use `write...save()` instead, then we won't have to specify nulls like this (we've already tested insertInto above).

src/test/scala/cognite/spark/v1/FilesRelationTest.scala
```diff
@@ -39,74 +55,300 @@ class FilesRelationTest extends FlatSpec with Matchers with SparkTest {
...
+      assert(idsAfterDelete.isEmpty)
+    } finally {
+      try {
+        cleanupFiles(source)
+      } catch {
+        case NonFatal(_) => // ignore
+      }
+    }
+  }
+
```
wjoel, line 344: Would it be possible/reasonable to have a test like https://github.com/cognitedata/cdp-spark-datasource/blob/master/src/test/scala/cognite/spark/v1/EventsRelationTest.scala#L1100 for Files schemas?

src/test/scala/cognite/spark/v1/FilesRelationTest.scala
sush03: I was checking two updates (one by id and other by externalId).
will just write assert(getNumberOfRowsUpdated(metricsPrefix, "files") == 1).

src/test/scala/cognite/spark/v1/FilesRelationTest.scala
sush03: Yeah will do that !

src/test/scala/cognite/spark/v1/FilesRelationTest.scala
sush03: Here, I would be testing insertInto, will have write..save() everywhere else

src/test/scala/cognite/spark/v1/FilesRelationTest.scala
sush03: It makes sense to check that all necessary parameters are there in each schema, but as the name is required parameter in files how should I go about it?

README.md
```diff
@@ -322,20 +322,28 @@ The schemas mirror the CDF API as closely as possible.
...
+| `lastUpdatedTime`    | `timestamp`           | No        |
+| `sourceCreatedTime`  | `timestamp`           | Yes       |
+| `sourceModifiedTime` | `timestamp`           | Yes       |
+| `securityCategories` | `array(long)`         | Yes       |
+| `uploadUrl`          | `string`              | Yes       |
+| `dataSetId`          | `long`                | Yes       |
+
+
+
+
```
wjoel, line 346: Remove those newlines ;)

wjoel: Not that important, you can just skip this if the tests pass now. The events thing was unrelated, maybe we need to adjust that test, but it's clearly not because of your changes.

PR #435, opened 2020-05-12
docs(readme): escape special character

polomani:
this breaks on our docs portal as well


polomani:
what happened to ci?

wjoel:
The broken test is due to a change in the validation of some parameters, we have updates on the way to the master branch. We can just merge this to master, as it doesn't change the code and the test failures are unrelated.

PR #437, opened 2020-05-19
Update Cognite Scala SDK to 1.3.0

wjoel:


sush03:
We need to fix those failing tests. The files would be fixed by the new changes

PR #438, opened 2020-05-19
Use ParallelTestExecution

wjoel:


src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
```diff
@@ -5,25 +5,25 @@ import io.scalaland.chimney.dsl._
...
-class AssetsRelationTest extends FlatSpec with Matchers with SparkTest {
+class AssetsRelationTest extends FlatSpec with Matchers with ParallelTestExecution with SparkTest {
 
   val sourceDf = spark.read
     .format("cognite.spark.v1")
     .option("apiKey", readApiKey)
     .option("type", "assets")
     .load()
-  sourceDf.createTempView("sourceAssets")
+  sourceDf.createOrReplaceTempView("sourceAssets")
```
exyi, line 19: I don't see why is this necessary? Would it be better to use `createTempView` to prevent races?

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
```diff
@@ -506,17 +501,24 @@ class AssetsRelationTest extends FlatSpec with Matchers with SparkTest {
         .option("onconflict", "does-not-exists")
         .save()
     }
-    enableSparkLogging()
```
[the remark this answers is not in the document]
wjoel: It went to global, so we suppress all Spark logging when running the tests now. As they execute in parallel and share the same execution context, I don't think there's any better way to do it.

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
wjoel: If the same SparkSession somehow gets reused the `createTempView` will fail, and the tests abort. Typically that shouldn't happen, but I've seen it, and in that case we do want to just replace the existing one.

PR #439, opened 2020-05-20
Fix implicit conversion from BigInteger/BigDecimal

exyi:
Spark uses `java.math.BigInteger` and `java.math.BigDecimal` instead of the Scala types, so the implicit conversions actually did not work.


wjoel:
@exyi seems someone is writing data to our project, if you could add some filters on createdTime (and if it makes sense), you can do that, otherwise I can delete the new data (I did it yesterday).

wjoel:
Some of the test failures are already taken care of in https://github.com/cognitedata/cdp-spark-datasource/pull/438 so maybe you could review that? :pray:

PR #440, opened 2020-05-23
[CDF-5528] Add delete for data points

exyi:
This is the delete schema for datapoints delete:

```scala
final case class DeleteDataPointsItem(
  id: Option[Long],
  externalId: Option[String],
  exclusiveBegin: Option[Instant],
  inclusiveBegin: Option[Instant],
  exclusiveEnd: Option[Instant],
  inclusiveEnd: Option[Instant]
)
```

Can delete ranges. If you want to delete specific points, just set `inclusiveBegin` and `inclusiveEnd` to the same value.

TODO:
- [x] test for string data points
- [x] docs to README.md

### DIFF
--- a/src/main/scala/cognite/spark/v1/DataPointsRelation.scala
+++ b/src/main/scala/cognite/spark/v1/DataPointsRelation.scala
@@ -21,6 +21,8 @@ sealed case class Max(value: Instant) extends Limit
 
 final case class AggregationFilter(aggregation: String)
 
+import cognite.spark.v1.SparkSchemaHelper.{asRow, fromRow, structType}
+
 abstract class DataPointsRelationV1[A](config: RelationConfig, shortName: String)(
     override val sqlContext: SQLContext)
     extends CdfRelation(config, shortName)
@@ -29,10 +31,19 @@ abstract class DataPointsRelationV1[A](config: RelationConfig, shortName: String
     with Serializable
     with InsertableRelation {
   import CdpConnector._
+  import DataPointsRelationV1._
 
   def toRow(a: A): Row
 
   def toRow(requiredColumns: Array[String])(item: A): Row
+
+  def delete(rows: Seq[Row]): IO[Unit] = {
+    val deleteRanges =
+      rows
+        .map(fromRow[DeleteDataPointsItem](_))
+        .map(processDeleteRow)
+    client.dataPoints.deleteRanges(deleteRanges)
+  }
 
   override def insert(data: DataFrame, overwrite: Boolean): Unit =
     data.foreachPartition((rows: Iterator[Row]) => {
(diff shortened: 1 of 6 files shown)


src/main/scala/cognite/spark/v1/DataPointsRelation.scala
```diff
@@ -172,4 +183,52 @@ object DataPointsRelationV1 {
               newAllPoints)
         }
     }
+
+  def processDeleteRow(x: DeleteDataPointsItem): DeleteDataPointsRange = {
+    val id = (x.id, x.externalId) match {
+      case (Some(_), Some(_)) =>
+        throw new IllegalArgumentException(s"Delete row for Data Points can not contain both id and externalId (on row $x)")
```
wjoel: Elsewhere we just use the id and ignore the external id if both are set, I think it may be beneficial.

src/main/scala/cognite/spark/v1/DataPointsRelation.scala
```diff
@@ -172,4 +183,52 @@ object DataPointsRelationV1 {
...
+      case (Some(id), None) => CogniteInternalId(id)
+      case (None, Some(externalId)) => CogniteExternalId(externalId)
+      case (None, None) =>
+        throw new IllegalArgumentException(s"Delete row for Data Points must contain id or externalId (on row $x)")
+    }
+
+    // The API only supports exclusiveEnd and inclusiveBegin, so we must adjust the row format for this
+    // Since the minimum datapoint precision is 1ms, we can just assume exclusiveEnd = inclusiveEnd + 1ms
+    // and inclusiveBegin = exclusiveBegin + 1ms
+    def toLowerbound(lower: Option[Instant], upper: Option[Instant], part: String): Long =
```
wjoel, line 202: Make this a `class-level private def` instead, and pass `x`

src/main/scala/cognite/spark/v1/DataPointsRelation.scala
```diff
@@ -172,4 +183,52 @@ object DataPointsRelationV1 {
...
+        case (None, Some(upper)) =>
+          upper.toEpochMilli + 1
+        case (None, None) =>
+          throw new IllegalArgumentException(s"Delete row for Data Points must contain inclusive$part or exclusive$part (on row $x)")
+      }
+
+    val inclusiveBegin = toLowerbound(x.inclusiveBegin, x.exclusiveBegin, "Begin")
+    val exclusiveEnd = toLowerbound(x.exclusiveEnd, x.inclusiveEnd, "End")
+    if (exclusiveEnd <= inclusiveBegin) {
+      throw new IllegalArgumentException(s"Delete range [$inclusiveBegin, $exclusiveEnd) is empty (on row $x)")
```
wjoel: Maybe "is invalid" instead, as it includes the case where end is before begin.

PR #444, opened 2020-06-22
[CDF-7436] Don't truncate metadata values, bump to 1.4.0

sebbert:
Some services have increased their limits, and we should not fail silently here anyway.


src/main/scala/cognite/spark/v1/Constants.scala
```diff
@@ -20,7 +20,6 @@ object Constants {
   val DefaultMaxBackoffDelay: FiniteDuration = 15.seconds
   // val DefaultBaseUrl = "http://localhost:4001"
```
wjoel, line 21: Why do we have this? 🙈

sebbert: Let's remove it while we're at it :)

PR #451, opened 2020-07-21
Add functions API as UDF

ebrattli:


ebrattli:
For now input to the function is just an empty jsonObject. Not sure how we will make user send dictionary as input to function so that we can create a JsonObject. Also, the result is made into a string. Not sure how to specify generic output from the UDF.

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -85,13 +91,67 @@ class DefaultSource
...
+    } else {
+      IO.pure(Json.fromJsonObject(JsonObject.empty))
+    }
+  }
+
+  private def callFunctionAndGetResult(config: RelationConfig, functionId: Long, data: Json): IO[Json] = {
+    implicit val backend: SttpBackend[IO, Nothing] = CdpConnector.retryingSttpBackend(config.maxRetries)
+    for {
+      client <- GenericClient.forAuth[IO](Constants.SparkDatasourceVersion, config.functionsAuth.get, config.baseUrl, apiVersion = Some("playground"))
+      call <- client.functionCalls(functionId).callFunction(Json.fromJsonObject(JsonObject.empty))
```
wjoel, line 118: Should be `callFunction(data)`?

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -85,13 +91,67 @@ class DefaultSource
...
+
+  private def initializeFunctionUdfs(config: RelationConfig, context: SQLContext): Unit = {
+    implicit val backend: SttpBackend[IO, Nothing] = CdpConnector.retryingSttpBackend(config.maxRetries)
+
+    val functionsIO = for {
+      client <- GenericClient.forAuth[IO](Constants.SparkDatasourceVersion, config.functionsAuth.get, config.baseUrl, apiVersion = Some("playground"))
+      functions <- client.functions.read()
+      l <- client.login.status()
+      f = functions.items.map(func =>
+        (func.name.replace(" ", "_"), (data: Any) =>
```
wjoel, line 131: We'd need some prefix to make this work, but this also assumes that the user has permissions to list CDF functions. I think the benefit is small compared to having a `cdf_function("my function", <args>)` (using name/externalId), `cdf_function(123, <args>)` (using id), and it has some downsides.

Any particular reason why we'd like to do it this way?

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -259,6 +319,9 @@ object DefaultSource {
     val auth = apiKey
       .orElse(bearerToken)
       .getOrElse(sys.error("Either apiKey or bearerToken is required."))
+    val functionsAuth = parameters
+      .get("functionsApiKey")
```
wjoel, line 323: Why do we need a separate one for this? Is it just for testing?

src/main/scala/cognite/spark/v1/DefaultSource.scala
ebrattli: Yes, it is just that I have not found out what the input type should be. The Json has to be a Json object like `{ "data": <the data>}`. It would be nice to have input be Any, but I have not found any AnySerializers.

src/main/scala/cognite/spark/v1/DefaultSource.scala
ebrattli: I just assumed that if a user has permissions to call a function, then he probably has permissions to list functions. Not sure if those are separate permissions, but if so, then we can refactor it.

src/main/scala/cognite/spark/v1/DefaultSource.scala
ebrattli: The functions are added in DefaultSource. During the initialization there we only have access to one apiKey which might not have permissions for functions. For example, if we read from publicdata, then we need a separate key to read from functions.

src/main/scala/cognite/spark/v1/DefaultSource.scala
wjoel: Still, I don't think registering all of them automatically is something we want to do: risk of collisions and possibly a performance hit if there are many. Doesn't seem worth it over `cdf_function("my function", <args>)` imo.

src/main/scala/cognite/spark/v1/DefaultSource.scala
wjoel: Hmm, I see, that can be useful but then I'd suggest we use `"functionsApiKey"` if set, otherwise just use `"apiKey"`, as most users won't need to use a separate key for this and having to specify the same one twice would be annoying.

src/main/scala/cognite/spark/v1/DefaultSource.scala
wjoel, line 131: For args, we can use `data: Row`. The user can then call the function as `cdf_function("foo", struct(*))` to pass all columns. Our UDF will receive the columns in a `Row` object which also includes a schema. The schema doesn't matter too much as we should just call `to_json` (that's what it's called in SQL, there should be something similar in the Scala API) to encode it as a string.

For the return type we have two options that I can see, either we make it to always be a string, and let the user decide if they should decode it as JSON, or we make it a `Map[String, String]` where the values would be JSONs encoded as strings. For simple cases the latter is a bit more convenient as the user can do `cast(cdf_function("my function", struct(*)).prediction as double)` and similar, which is probably a common use case.

wjoel: If functions are allowed to return bare numbers we'd have to go for always returning a string, though. We can add some additional methods later like `cdf_function_double` which would return a double, or something, but that's not needed for the first release.

ebrattli: I could not figure out how to convert a row into JSON. The `to_json` function accepts a `Column` type. All internet examples I found used it from SQL. So I think it is best if the user uses `to_json` from SQL.

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -85,13 +94,114 @@ class DefaultSource
     new SequenceRowsRelation(config, sequenceId)(sqlContext)
   }
 
+  private def getFunctionResult(
+      client: GenericClient[IO],
+      functionId: Long,
+      result: FunctionCall): IO[Json] = {
+    var res = result
+    while (res.status != None && res.status.get == "Running") {
```
wjoel, line 102: This is potentially infinite, maybe we can use `retryWhile` here?

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -85,13 +94,114 @@ class DefaultSource
...
+    }
+
+    if (res.status != None && res.status.get == "Completed") {
+      for {
+        functionCallResponse <- client
+          .functionCalls(functionId)
+          .retrieveResponse(res.id.getOrElse(0L))
+      } yield functionCallResponse.response.getOrElse(Json.fromJsonObject(JsonObject.empty))
+    } else {
+      IO.pure(Json.fromJsonObject(JsonObject.empty))
```
wjoel, line 116: Returning an empty Json object on errors seems a bit strange, I think it might be better to throw an exception.

src/main/scala/cognite/spark/v1/DefaultSource.scala
ebrattli: I could not find a `retryWhile` function. Instead I added a counter. 30 retries seem about right from what I have experienced in datastudio.

wjoel: It's this one: https://github.com/cognitedata/cdp-spark-datasource/blob/3db8b5401685e4d9736c9bd6fb03196fef43e3f7/src/test/scala/cognite/spark/v1/SparkTest.scala#L72
For this use case we probably wouldn't want an exponential backoff. Either way, we can change that later.

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -85,13 +94,116 @@ class DefaultSource
...
+
   // scalastyle:off cyclomatic.complexity method.length
   override def createRelation(
       sqlContext: SQLContext,
       parameters: Map[String, String],
       schema: StructType): BaseRelation = {
+
     val resourceType = parameters.getOrElse("type", sys.error("Resource type must be specified"))
     val config = parseRelationConfig(parameters, sqlContext)
+    initializeFunctionUdfs(config, sqlContext)
```
wjoel, line 205: Maybe we should do this separately and make it opt-in, we're registering those UDFs for every user of the data source like this. A separate cognite.spark.v1.udfs package would make sense, we also have some in Jetfire that we can/should share as well. Anyways, we can make that change in a subsequent PR.

src/main/scala/cognite/spark/v1/udf/CogniteUdfs.scala
```diff
@@ -0,0 +1,143 @@
...
+import com.softwaremill.sttp.SttpBackend
+import io.circe.{Json, JsonObject, parser}
+import org.apache.spark.sql.SparkSession
+
+class CogniteUdfs(sparkSession: SparkSession) extends Serializable {
+
+  def initializeUdfs(
+      apiKey: ApiKeyAuth,
+      baseUrl: Option[String] = None,
+      maxRetries: Option[Int] = None): Unit = {
```
wjoel, line 17: Easier to just do `baseUrl: String = DefaultBaseUrl, maxRetries: Int = DefaultMaxRetries)`

src/main/scala/cognite/spark/v1/udf/CogniteUdfs.scala
```diff
@@ -0,0 +1,143 @@
...
+    if (res.status.isDefined && res.status.get == "Completed") {
+      for {
+        functionCallResponse <- client
+          .functionCalls(functionId)
+          .retrieveResponse(res.id.getOrElse(0L))
+      } yield functionCallResponse.response.getOrElse(Json.fromJsonObject(JsonObject.empty))
+    } else {
+      val logs = client.functionCalls(functionId).retrieveLogs(res.id.get).unsafeRunSync()
+      throw new CdfSparkException(
+        "CDF function call failed. Res = " + res.toString + "logs = " + logs.items.toString())
```
wjoel: Missing a space in `"logs = "`, should be (adding a comma, as well) `", logs = "`

src/main/scala/cognite/spark/v1/udf/CogniteUdfs.scala
```diff
@@ -0,0 +1,143 @@
...
+import cognite.spark.v1.Constants.{DefaultBaseUrl, DefaultMaxRetries}
+import cognite.spark.v1.{CdfSparkException, CdpConnector, Constants}
+import com.cognite.sdk.scala.common.ApiKeyAuth
+import com.cognite.sdk.scala.v1.{FunctionCall, GenericClient}
+import com.softwaremill.sttp.SttpBackend
+import io.circe.{Json, JsonObject, parser}
+import org.apache.spark.sql.SparkSession
+
+class CogniteUdfs(sparkSession: SparkSession) extends Serializable {
+
```
wjoel, line 13: We can do `implicit val backend: SttpBackend[IO, Nothing] = CdpConnector.retryingSttpBackend(maxRetries)` just once here, I think? Unless we need to get a new one for each call.

src/test/scala/cognite/spark/v1/FunctionsUDFTest.scala
```diff
@@ -0,0 +1,47 @@
...
+class FunctionsUDFTest extends FlatSpec with Matchers with ParallelTestExecution with SparkTest with Inspectors {
+
+  it should "read assets with functionUDF" taggedAs ReadTest in {
+    implicit val backend: SttpBackend[IO, Nothing] = CdpConnector.retryingSttpBackend(DefaultMaxRetries)
+    val df = spark.read
+      .format("cognite.spark.v1")
+      .option("apiKey", readApiKey)
+      .option("type", "assets")
+      .option("limitPerPartition", "1")
+      .option("partitions", "1")
```
wjoel, line 22: Maybe add a `.option("limit", "50")` or something to this so we don't end up making too many (slow) calls.

src/main/scala/cognite/spark/v1/udf/CogniteUdfs.scala
ebrattli: I added logs just to debug something, and then I removed it now. Is it something we should include?

PR #452, opened 2020-07-28
Add exception base classes and use custom exceptions

ebrattli:
Exceptions in Spark must have common base class so that we can decide which to let through in Jetfire. https://cognitedata.atlassian.net/secure/RapidBoard.jspa?rapidView=303&modal=detail&selectedIssue=CDF-5649

### DIFF
--- a/src/main/scala/cognite/spark/v1/CDFSparkException.scala
+++ b/src/main/scala/cognite/spark/v1/CDFSparkException.scala
@@ -1,0 +1,23 @@
+package cognite.spark.v1
+
+class CdfSparkException(message: String) extends Exception(message) {
+  def this(message: String, cause: Throwable) {
+    this(message)
+    initCause(cause)
+  }
+
+  def this(cause: Throwable) {
+    this(Option(cause).map(_.toString).orNull, cause)
+  }
+}
+
+class CdfSparkIllegalArgumentException(message: String) extends CdfSparkException(message) {
+  def this(message: String, cause: Throwable) {
+    this(message)
+    initCause(cause)
+  }
+
+  def this(cause: Throwable) {
+    this(Option(cause).map(_.toString).orNull, cause)
+  }
+}
--- a/performancebench/src/main/scala/cognite/spark/performancebench/SparkUtil.scala
+++ b/performancebench/src/main/scala/cognite/spark/performancebench/SparkUtil.scala
@@ -1,10 +1,11 @@
 package cognite.spark.performancebench
 
+import cognite.spark.v1.CdfSparkException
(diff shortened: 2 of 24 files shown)


src/main/scala/cognite/spark/v1/CDFSparkException.scala
```diff
@@ -0,0 +1,27 @@
+package cognite.spark.v1
+
+trait CDFSparkException {}
```
wjoel, line 3: We capitalize this as `CdfSparkException` (change the file name as well), but apart from that this, looks pretty good to me!

src/main/scala/cognite/spark/v1/CDFSparkException.scala
```diff
@@ -0,0 +1,27 @@
...
+
+class CDFSparkInternalException(message: String) extends Exception(message) with CDFSparkException {
```
wjoel, line 5: I think we can drop `Internal` from this name.

ebrattli: That would conflict with the trait. We could rename the trait to something else, but I'm not sure what.

wjoel: Could we make CdfSparkException be a class and have CdfSparkIllegalArgumentException inherit from it?

ebrattli: We can do that, however, then `CdfSparkIllegalArgumentException` can not inherit from `IllegalArgumentException` due to multiple inheritance. But, we probably don't need to inherit from `IllegalArgumentException`.

wjoel: Because many of those messages like "Granularity specified but no aggregation requested" are not actually internal at all, so I'm not a big fan of that name.

ebrattli: I have removed the trait now. Had to change some tests to expect `CdfSparkIllegalArgumentException` instead of `IllegalArgumentException`.

wjoel:
Note that you have to run scalafmt as Jenkins is failing with
>  [error] /home/runner/work/cdp-spark-datasource/cdp-spark-datasource/src/main/scala/cognite/spark/v1/SequenceRowsRelation.scala isn't formatted properly!
> [error] (Compile / scalafmtCheck) /home/runner/work/cdp-spark-datasource/cdp-spark-datasource/src/main/scala/cognite/spark/v1/SequenceRowsRelation.scala isn't formatted properly!

PR #459, opened 2020-08-14
Set x-cdp headers correcly

mathialo:
Let the new sdk version set `x-cdp-sdk`, let `x-cdp-app` remain as `com.cognite.spark.datasource-VERSION`, and let user (optionally) set `x-cdp-clienttag` in options.

### DIFF
--- a/src/main/scala/cognite/spark/v1/CdpConnector.scala
+++ b/src/main/scala/cognite/spark/v1/CdpConnector.scala
@@ -7,9 +7,9 @@ import cats.effect.{ContextShift, IO, Timer}
 import com.cognite.sdk.scala.common.{GzipSttpBackend, RetryingBackend}
 import com.cognite.sdk.scala.v1.GenericClient
 import com.google.common.util.concurrent.ThreadFactoryBuilder
-import com.softwaremill.sttp.asynchttpclient.cats.AsyncHttpClientCatsBackend
 import com.softwaremill.sttp._
 import com.softwaremill.sttp.asynchttpclient.SttpClientBackendFactory
+import com.softwaremill.sttp.asynchttpclient.cats.AsyncHttpClientCatsBackend
 
 import scala.concurrent.ExecutionContext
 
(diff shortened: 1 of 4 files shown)


src/main/scala/cognite/spark/v1/CdpConnector.scala
```diff
@@ -47,7 +47,8 @@ object CdpConnector {
       Constants.SparkDatasourceVersion,
```
wjoel, line 47: It would be useful to allow setting this as well. Jetfire would like to override it, I think.

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -18,6 +18,7 @@ import scala.util.parsing.json.JSONObject
 
 final case class RelationConfig(
     auth: Auth,
+    clientTag: Option[String],
```
wjoel, line 21: Let's make it possible to set the application name as well.

mathialo: So make spark 'invisible' in terms of these headers?

src/main/scala/cognite/spark/v1/CdpConnector.scala
sebbert: Application name is required, while the client tag is optional. Would it not make more sense that we use the application name to identify the spark data source version, and use the client tag to identify e.g. jetfire? Otherwise, we would have to start requiring the user to specify the application name, which is a breaking change.

wjoel: @sebbert Client tag is meant to be used for grouping related requests, in our case it could be all requests related to one Spark job.

We can keep this (the Spark data source version) as the default application name, and allow the user to override it. Since it's optional, it wouldn't be a breaking change.

src/main/scala/cognite/spark/v1/DefaultSource.scala
wjoel: If the user overrides it with something unrelated to Spark: yes. It's more useful to have "JoelWeatherTransform" as an identifier if that's what I choose to override it with - of course I could choose "JoelWeatherTransform-Cognite-Spark-1.4.6" as well.

src/main/scala/cognite/spark/v1/CdpConnector.scala
sebbert: Ah right, that makes more sense. I'll update the PR 👍

sebbert: Nevermind, @mathialo already did :)

mathialo:
> LGTM, but I think we should squash this branch to avoid failing commits on master

My plan was to do that when merging, but I see now that squash-merging not enabled for this repo. I'll do it it manually.

sebbert:
@mathialo I already squashed it

PR #466, opened 2020-09-10
[CDF-86919] Raise a more helpful error when moving an asset to another root asset

sebbert:


src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -69,6 +69,26 @@ final case class InvalidNodeReferenceException(nodeIds: Seq[String], referencedF
       val refNodes = referencedFrom.map(x => s"'$x'").sorted.mkString(", ")
       s"${plural("Parent", "Parents")} $nodes referenced from $refNodes ${plural("does", "do")} not exist."
     })
+final case class AttemptedRootChangeException(cause: CdpApiException, newSubtreeRoot: Asset)
```
sebbert, line 72: Suggestions welcome for a different name here

wjoel: Maybe "Invalid" instead of "Attempted", but I think either choice is ok.

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -69,6 +69,26 @@ final case class InvalidNodeReferenceException(nodeIds: Seq[String], referencedF
...
+        val commaSeparatedAssets =
+          cause.duplicated.get
+            .map {
+              _.toIterable
+                .map { case (key, value) => s"$key=${value.asString.getOrElse(value.toString)}" }
+                .mkString("(", ", ", ")")
+            }
+            .mkString(", ")
+
+        val withRootAssetId = newSubtreeRoot.rootId.map(id => s" with id $id").getOrElse("")
```
wjoel, line 84: Maybe display externalId as well, if set.

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -69,6 +69,26 @@ final case class InvalidNodeReferenceException(nodeIds: Seq[String], referencedF
...
+
+        s"Attempted to move some assets to a different root asset$withRootAssetId. " +
+          "If this is intended, the assets must be manually deleted and re-created under the new root asset. " +
+          s"The following assets were attempted to be moved: $commaSeparatedAssets"
```
wjoel, line 88: Finish with a dot as the others do.

wjoel (approved):
Quite elegant! Minor comments, but LGTM.

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
sebbert: Unfortunately, we don't know the externalId of the root asset without making an extra call to CDF, as the `newSubtreeRoot` is not necessarily an actual root asset, just the "uppermost" node specified in the input data. CDF gives us a `rootId`, but not a `rootExternalId`, afaik.

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
sebbert: Good catch 👍

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
sebbert: I like "invalid", thanks!

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
wjoel: Yeah, could make such a call where we construct this exception, but perhaps not worth the effort.

PR #467, opened 2020-10-06
Don't use .handleError with non-exhaustive match

sebbert:
ApplicativeError.handleError takes a regular function (as opposed to a partial function), and expects it to always handle every error.
But we were using this function in a few places under the assumption that it accepts a partial function, so we will throw a MatchError if the exception did not match the pattern.
So let's use .adaptError when we just want to map certain exceptions to different exceptions, or .recover when we want to recover successfully from certain exceptions.

This caused jetfire to redact some safe error messages, where they should have been shown to the user.


wjoel (approved):
Nice catch!

PR #469, opened 2020-10-08
[WIP] Do batching of asset subtrees

hakontro:
Feel free to pick this up if you have the time @wjoel 
Seems to be working, but the test takes ages so I'm probably doing something stupid with the traversal here

### DIFF
--- a/src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
+++ b/src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
@@ -1,7 +1,7 @@
 package cognite.spark.v1
 
 import cats.effect.IO
-import cats.effect.syntax._
+import cats.effect.syntax.all._
 import cats.implicits._
 import cats.syntax._
 import cognite.spark.v1.PushdownUtilities.stringSeqToCogniteExternalIdSeq
@@ -148,7 +148,7 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
 
     for {
       _ <- validateSubtreeRoots(subtrees2.map(_.root))
-      _ <- subtrees2.map(t => buildSubtree(t.root, t.nodes)).sequence_ // TODO: parallel?
+      _ <- subtrees2.map(t => buildSubtree(t.root, t.nodes)).parSequence_
     } yield ()
   }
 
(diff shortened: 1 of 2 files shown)


src/test/scala/cognite/spark/v1/AssetHierarchyBuilderTest.scala
```diff
@@ -1048,6 +1049,26 @@ class AssetHierarchyBuilderTest
...
+  it should "successfully do batching of subtree roots" in {
+    val key = shortRandomString()
+
+    val dad = AssetCreate("dad", None, None, None, Some("dad"), None, Some(""))
+
+    val kids = (1 to 1100).flatMap(k => Seq(
+      AssetCreate(s"kid$k", None, None, None, Some(s"kid$k"), None, Some("dad"))))
+    val grandkids = (1 to 1100).flatMap(k => Seq(
+      AssetCreate(s"grandkid$k", None, None, None, Some(s"grandkid$k"), None, Some(s"kid$k"))
+    ))
```
sebbert, line 1061: Small nit, but you can simplify `x.flatMap(x => Seq(...))` to `x.map(x => ...)`

wjoel:
I haven't checked this in IntelliJ but might be one of the other tests which is a bit slow. The code looks good to me.

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -143,18 +143,23 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
...
-            val missingNodes = e.missing.get.map(j => j("externalId").get.asString.get).take(10)
-            val referencingNodes =
-              missingNodes
-              // always take the one with "lowest" externalId, so the errors are deterministic
-                .map(missing => roots.filter(_.parentExternalId == missing).minBy(_.externalId))
-                .map(_.externalId)
-            InvalidNodeReferenceException(missingNodes, referencingNodes)
-        }) *> IO.unit
+      ids
+        .grouped(1000)
```
lprovensi, line 147: If this is the assets api limit, should it be made into a constant?
I'm not very familiar with the code, but it would make it easier to understand why you are using 1000 here, change later and reuse...

src/test/scala/cognite/spark/v1/AssetHierarchyBuilderTest.scala
hakontro: Good catch! I initially coded it up with several things in the Seq(), hence the weird logic here.

wjoel (approved):
Good to go!

PR #470, opened 2020-10-09
Move RawTableRelation.defaultSchema to companion object

sebbert:
This makes the default RAW schema accessible without requiring an instance of RawTableRelation.
This is so that we can use it as the source of truth when querying RAW from Jetfire without schema inference, instead of copypasting the schema into the Jetfire codebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+## Enhancements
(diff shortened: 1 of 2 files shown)


sebbert:
Turned out not to be necessary, but I think it's an improvement anyways, so let's just make it part of the next release.

wjoel (approved):

PR #471, opened 2020-10-15
Add Labels support

mathialo:


src/main/scala/cognite/spark/v1/LabelsRelation.scala
```diff
@@ -0,0 +1,70 @@
...
+  override def delete(rows: Seq[Row]): IO[Unit] = {
+    val labelIds = rows.map(fromRow[LabelDeleteSchema](_)).map(_.externalId)
+    client.labels
+      .deleteByExternalIds(labelIds)
+      .flatTap(_ => incMetrics(itemsDeleted, labelIds.length))
+  }
+
+  override def upsert(rows: Seq[Row]): IO[Unit] = ???
+
+  override def update(rows: Seq[Row]): IO[Unit] = ???
```
mathialo, line 46: Not quite sure what to do with these. The labels API doesn't support updates, but having the `WritableRelation` made deletes and inserts easier

src/test/scala/cognite/spark/v1/LabelsRelationTest.scala
```diff
@@ -0,0 +1,84 @@
+package cognite.spark.v1
+
+import cognite.spark.v1.SparkSchemaHelper.fromRow
+import com.cognite.sdk.scala.v1.{Label, LabelCreate, LabelsFilter}
+import org.apache.spark.sql.DataFrame
+import org.scalatest.{FlatSpec, Inspectors, Matchers, ParallelTestExecution}
+
+class LabelsRelationTest
```
mathialo, line 8: Very basic tests for now, open to suggestions on new ones

mathialo:
@hakontro Seems the API key for the read project don't have labels access, causing the test to fail, are you able to assist on that?

hakontro:
> @hakontro Seems the API key for the read project don't have labels access, causing the test to fail, are you able to assist on that?

Not sure which service account key we use, but I tried updating some groups and rerunning the jobs.. let's see what happens

src/main/scala/cognite/spark/v1/LabelsRelation.scala
wjoel: Throw an exception saying it's not supported. We do this for some other relations as well: https://github.com/cognitedata/cdp-spark-datasource/blob/master/src/main/scala/cognite/spark/v1/NumericDataPointsRelation.scala#L116

src/main/scala/cognite/spark/v1/LabelsRelation.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  override def toRow(a: Label): Row = asRow(a)
+
+  override def uniqueId(a: Label): String = a.externalId
+
+  override def getStreams(filters: Array[Filter])(
+      client: GenericClient[IO],
+      limit: Option[Int],
+      numPartitions: Int): Seq[fs2.Stream[IO, Label]] =
+    Seq(client.labels.filter(LabelsFilter()))
```
wjoel, line 28: It would probably be easy to add support for the filter pushdown as in https://github.com/cognitedata/cdp-spark-datasource/blob/master/src/main/scala/cognite/spark/v1/EventsRelation.scala#L65

wjoel (approved):
LGTM once the `???` have been replaced with informative exceptions.

PR #472, opened 2020-10-15
[CDF-9109] Update to Spark 3, replace Option with value or null

wjoel:


src/test/scala/cognite/spark/v1/SparkSchemaHelperTest.scala
```diff
@@ -84,13 +84,13 @@ class SparkSchemaHelperTest extends FlatSpec with ParallelTestExecution with Mat
...
     asRow(x) should
-      be(Row(None, None, None, None, Some(Map("foo" -> "row", "bar" -> "a")), None, None, None))
+      be(Row(null, null, null, null, Map("foo" -> "row", "bar" -> "a"), null, null, null))
   }
 
   it should "construct Row from optional seq type" in {
     val x = TestTypeOption(None, None, None, None, None, None, Some(Seq(None, Some(10L))), None)
     asRow(x) should
-      be(Row(None, None, None, None, None, None, Some(Seq(None, Some(10L))), None))
+      be(Row(null, null, null, null, null, null, Seq(None, Some(10L)), null))
```
mathialo, line 93: Just a curiosity question, I thought using `null`s was considered bad practice in Scala (here were also combining with a `Option`), I guess this is some limitation put on us by Spark?

wjoel: Correct! I'm a bit surprised that it worked on Spark 2.x tbh, because Spark seems to avoid Option in the public APIs, which I would guess is because they also have a Java API.

mathialo (approved):

PR #433, opened 2020-05-06
Sequences

exyi:
Adds two relations

* **sequences** - contains the metadata about sequences
    - supports insert, update and delete. Upsert is problematic, since sequences don't return the nice 409 error when ids already exist.
    - does not do partitioned read nor filter pushdown (no support in API, no support in Scala SDK).
* **sequenceRows** - the sequence rows
    - requires `sequenceId` or `sequenceExternalId` option that specifies the sequence that should be used
    - when created, it fetches the sequence metadata and the Spark schema will correspond to that
    - supports upsert, delete and read
    - does filter and projection pushdown, does not do any partitioning

PR #477, opened 2020-11-10
Increase the default number of partitions

hakontro:
Turns out we were doing single partitions everywhere by default. Was this implemented to avoid anything in particular? We could do this change in jetfire-backend instead, but is this ever good at 1 * 1 anyways?

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -9,7 +9,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: olafurpg/setup-scala@v5
+      - uses: olafurpg/setup-scala@v10
       - name: Set variables
         run: |
           IMAGE_REVISION=$(date +%Y-%m-%d)-$(git rev-parse --short HEAD)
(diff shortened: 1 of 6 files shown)


wjoel:
It was incorrectly changed in this PR, I would suggest we set it back to the previous values of 200 partitions and parallelism of 10: https://github.com/cognitedata/cdp-spark-datasource/pull/433/files#diff-08ba73acdff062613774029378c57db0cbed7ba07abe77e3e5e6da27151cbb01
`val DefaultMaxBackoffDelay: FiniteDuration = 120.seconds` as well.

I assume it was changed for local testing and accidentally slipped through to master.

wjoel:
It was incorrectly changed in this PR: https://github.com/cognitedata/cdp-spark-datasource/pull/433/files#diff-08ba73acdff062613774029378c57db0cbed7ba07abe77e3e5e6da27151cbb01
`val DefaultMaxBackoffDelay: FiniteDuration = 120.seconds` as well.

I assume it was changed for local testing and accidentally slipped through to master.

wjoel (approved):

exyi (approved):
Sorry guys, this is a mistake introduced for local testing - it was easier to see what's going on when there is a single request. The timeout also shouldn't have been decreased.

PR #478, opened 2020-11-12
Add relationships

emelvarol-cognite:
Ready for review


src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
```diff
@@ -0,0 +1,156 @@
...
+  var insertSchema: StructType = structType[RelationshipsInsertSchema]
+  var readSchema: StructType = structType[RelationshipsReadSchema]
+  var deleteSchema: StructType = structType[RelationshipsDeleteSchema]
+}
+
+final case class RelationshipsDeleteSchema(
+    externalId: String
+)
+
+final case class RelationshipsInsertSchema(
```
emelvarol-cognite, line 130: @hakontro I saw these classes are defined in all data types but I could not catch where they are being used looking at SdkV1Relation. Could you point me there?

For labels, it is defined as "labels: Option[Seq[Map[String, String]]]" considering it can be translated into Option[Seq[CogniteExternalId]]. Is it a correct approach?

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
```diff
@@ -0,0 +1,156 @@
...
+
+  override def upsert(rows: Seq[Row]): IO[Unit] =
+    throw new CdfSparkException("Upsert is not supported for relationships")
+
+  override def update(rows: Seq[Row]): IO[Unit] =
+    throw new CdfSparkException("Update is not supported for relationships")
+}
+
+object RelationshipsRelation {
+  var insertSchema: StructType = structType[RelationshipsInsertSchema]
```
emelvarol-cognite, line 121: the same for this

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
```diff
@@ -0,0 +1,156 @@
...
+    // the same RDD partition
+    streamsPerFilter.transpose
+      .map(s => s.reduce(_.merge(_)))
+  }
+
+  def relationshipsFilterFromMap(m: Map[String, String]): RelationshipsFilter =
+    RelationshipsFilter(
+      sourceExternalIds = m.get("sourceExternalIds").map(stringSeqFromWrappedArray),
+      sourceTypes = m.get("sourceTypes").map(stringSeqFromWrappedArray),
+      targetExternalIds = m.get("targetExternalIds").map(stringSeqFromWrappedArray),
```
emelvarol-cognite, line 87: As far as I understand, we need a function to evaluate strings here meaning "array('ext1', 'ext2')" or "array(\"ext1\", \"ext2\")" into Seq("ext1", "ext2") - just thinking about array function I used in transformations. 
Is there a builtin function for that?

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
```diff
@@ -0,0 +1,156 @@
...
+      targetTypes = m.get("targetTypes").map(stringSeqFromWrappedArray),
+      dataSetIds = m.get("dataSetId").map(idsFromWrappedArray(_).map(CogniteInternalId)),
+      startTime = timeRangeFromMinAndMax(m.get("minStartTime"), m.get("maxStartTime")),
+      endTime = timeRangeFromMinAndMax(m.get("minEndTime"), m.get("maxEndTime")),
+      activeAtTime = timeRangeFromMinAndMax(m.get("minActiveAtTime"), m.get("maxActiveAtTime")),
+      labels = getLabelsFilter(m.get("labels")),
```
emelvarol-cognite, line 93: The same goes for this. How can I convert string "map('containsAll', array('label1', 'label2'))" into a compatible data type? Had some troubles here.

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
hakontro: These are actually used in the update and upsert methods, we don't need them yet since we aren't implementing those :) But just leave them there, we'll do that soon!

hakontro: RIght click class name -> Find Usages will often help you find stuff like this!

hakontro: > For labels, it is defined as "labels: Option[Seq[Map[String, String]]]" considering it can be translated into Option[Seq[CogniteExternalId]]. Is it a correct approach?
Not sure where you're referring to here, can you point me to it?

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
hakontro: Also not sure about this question, but that's what the `stringSeqFromWrappedArray` is doing already so we're good(?)

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
```diff
@@ -0,0 +1,156 @@
...
+    targetType: String,
+    startTime: Option[Instant] = None,
+    endTime: Option[Instant] = None,
+    confidence: Float = 0,
+    labels: Option[Seq[Map[String, String]]] = None,
+    dataSetId: Option[Long] = None
+)
+
+final case class RelationshipsReadSchema(
+    externalId: Option[String] = None,
```
emelvarol-cognite, line 144: @hakontro Is this also used for update/upsert?

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
emelvarol-cognite: I think this question was nonsense since I did not understand this was for update/upsert.

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
hakontro: This is only there so we can expose it to Transformations actually, so that it can know what the column schema of reading a Relationship is like

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
```diff
@@ -0,0 +1,156 @@
...
+      endTime = timeRangeFromMinAndMax(m.get("minEndTime"), m.get("maxEndTime")),
+      activeAtTime = timeRangeFromMinAndMax(m.get("minActiveAtTime"), m.get("maxActiveAtTime")),
+      labels = getLabelsFilter(m.get("labels")),
+      confidence = confidenceRangeFromMinAndMax(m.get("minConfidence"), m.get("maxConfidence")),
+      createdTime = timeRangeFromMinAndMax(m.get("minCreatedTime"), m.get("maxCreatedTime")),
+      lastUpdatedTime = timeRangeFromMinAndMax(m.get("minLastUpdatedTime"), m.get("maxLastUpdatedTime"))
+    )
+
+  override def insert(rows: Seq[Row]): IO[Unit] = {
+    val relationships = rows.map(fromRow[RelationshipCreate](_))
```
emelvarol-cognite, line 100: Here I am curious how RelationshipCreate would be translated into Row. Type of labels is a class we defined, how does it work? Should I add something to SparkSchemaHelperImpl.fromRow?
```suggestion
final case class RelationshipCreate(
    externalId: String,
    sourceExternalId: String,
    sourceType: String,
    targetExternalId: String,
    targetType: String,
    startTime: Option[Instant] = None,
    endTime: Option[Instant] = None,
    confidence: Option[Double] = None,
    dataSetId: Option[Long] = None,
    labels: Option[Seq[CogniteExternalId]] = None
) extends WithRequiredExternalId```

File: src/main/scala/cognite/spark/v1/RelationshipsRelation.scala (hunk 4 above)
Comment (hakontro) [reply src/main/scala/cognite/spark/v1/RelationshipsRelation.scala:93]: Good question! The desired en state here is to populate the `RelationshipsFilter` from the filters that Spark provide: `filters: Array[Filter]`. These filters are actually Spark translating the filter part of a query such as `"SELECT * FROM _CDF.ASSETS WHERE ASSETID IS IN('10', '50') AND EXTERNALID LIKE '%WELL'`, and we probably won't be able to parse queries like this into a "containsAll" / "containsAny" type of filter.. probably only the "containsAny" type will work?

File: src/main/scala/cognite/spark/v1/RelationshipsRelation.scala (hunk 3 above)
Comment (emelvarol-cognite) [reply src/main/scala/cognite/spark/v1/RelationshipsRelation.scala:87]: No, I created that function (as a placeholder) but it is not implemented properly :D

File: src/main/scala/cognite/spark/v1/RelationshipsRelation.scala (hunk 6 above)
Comment (hakontro) [reply src/main/scala/cognite/spark/v1/RelationshipsRelation.scala:100]: Very good question, and I'm not sure.. the `fromRow` function is based on Scala magic called `macros`.. does it compile?

File: src/main/scala/cognite/spark/v1/RelationshipsRelation.scala (hunk 3 above)
Comment (hakontro) [reply src/main/scala/cognite/spark/v1/RelationshipsRelation.scala:87]: I think it should be good actually, given that `idsFromWrappedArray` works :)

File: src/main/scala/cognite/spark/v1/RelationshipsRelation.scala (hunk 6 above)
Comment (hakontro) [reply src/main/scala/cognite/spark/v1/RelationshipsRelation.scala:100]: But to answer some of your question - the compiler does a pre-run where it generates the fromRow function _based on the class that we pass it_, so there will exist a particular Scala implementation of the `fromRow[RelationshipCreate]` (so not quite the same as using generics to create abstract functions). This function will be able to take a Spark `Row` object and translate it into a `RelationshipCreate`, as long as that `Row` has the structure that we expect.

File: src/main/scala/cognite/spark/v1/RelationshipsRelation.scala (hunk 5 above)
Comment (emelvarol-cognite) [reply src/main/scala/cognite/spark/v1/RelationshipsRelation.scala:144]: Again, labels give me trouble here :D It is defined as a Option[Seq[CogniteExternalId]] in scala-sdk. Should I use the same type or keep the version here?

File: src/main/scala/cognite/spark/v1/RelationshipsRelation.scala (hunk 6 above)
Comment (emelvarol-cognite) [reply src/main/scala/cognite/spark/v1/RelationshipsRelation.scala:100]: Yes, it did not complain but I could not try creating anything here since again I could not figure what to do with labels again.

File: src/main/scala/cognite/spark/v1/RelationshipsRelation.scala (hunk 3 above)
Comment (emelvarol-cognite) [reply src/main/scala/cognite/spark/v1/RelationshipsRelation.scala:87]: Spent some time but could not do it properly. (very dumb when it comes to regex maybe :D )

File: src/main/scala/cognite/spark/v1/RelationshipsRelation.scala (hunk 4 above)
Comment (emelvarol-cognite) [reply src/main/scala/cognite/spark/v1/RelationshipsRelation.scala:93]: I think I missed what I would receive here in the first place.
For example, if the user queries this (of course if it is a possible case here):
"SELECT * FROM _CDF.RELATIONSHIPS WHERE "my_target_label" in LABELS"
what would I receive in m.get("labels")?

File: src/main/scala/cognite/spark/v1/RelationshipsRelation.scala (hunk 3 above)
Comment (emelvarol-cognite) [reply src/main/scala/cognite/spark/v1/RelationshipsRelation.scala:87]: ah, I checked that one. It works for lists of integers but not for lists of strings. In relationships, we receive a list of external ids instead of ids. So we need a different function to parse it.

File: src/main/scala/cognite/spark/v1/RelationshipsRelation.scala (hunk 4 above)
Comment (emelvarol-cognite) [reply src/main/scala/cognite/spark/v1/RelationshipsRelation.scala:93]: I think,
this would stand for containsAny
"SELECT * FROM _CDF.RELATIONSHIPS WHERE 'my_target_label' IN LABELS OR  'my_target_label2' IN LABELS" 
this would work for containsAll
"SELECT * FROM _CDF.RELATIONSHIPS WHERE 'my_target_label' IN LABELS AND  'my_target_label2' IN LABELS" 

but I need to understand how it works and parse it accordingly.

File: src/main/scala/cognite/spark/v1/PushdownUtilities.scala [hunk 7]
```diff
@@ -103,9 +105,14 @@ object PushdownUtilities {
...
 
-  def assetIdsFromWrappedArray(wrappedArray: String): Seq[Long] =
+  def idsFromWrappedArray(wrappedArray: String): Seq[Long] =
     wrappedArray.split("\\D+").filter(_.nonEmpty).map(_.toLong)
 
+  def stringSeqFromWrappedArray(wrappedArray: String): Seq[String] = {
+    val regQuotes: Regex = "\"(.*?)\"|'(.*?)'".r
+    regQuotes.findAllIn(wrappedArray).toSeq
+  }
+
```
emelvarol-cognite, line 121: @hakontro I think this should work for extracting externalIds from a WrappedArray of Strings. It grabs substrings between quotation or double quotations.

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
hakontro: So the code we have now doesn't allow for that, and honestly I don't think we need to..
`filtersAsMaps` is a `Seq[Map[String, String]]`, where each Map will be used to build a `RelationshipsFilter`. We decode `OR`s into multiple Maps here, so for example
```
LABELS = 'my_target_label' OR LABELS  = 'my_target_label2'
```
will be decoded into
```
Seq(
Map("label" -> "my_target_label"),
Map("label" -> "my_target_label2
```
these are then transformed into `RelationshipsFilter`s, but we can't implement OR from here, and so not containsAny.

The good thing is that these filters are only used to build the queries we run to CDF, and Spark actually filters that result again afterwards! So we can feel free to read too many rows, and could even skip the labels filter here entirely without getting the wrong result. It would only impact performance, and in the case or Relationships, probably not very much.

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
hakontro: Did we figure this one out? Don't remember.

emelvarol-cognite: I solved it by using RelationshipsReadSchema instead of Relationship class. I converted labels into a list of strings.

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
emelvarol-cognite: I got the idea, thanks! I noticed that it was not possible for now. I only implemented for containsAll and changed the approach a bit in general.

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
emelvarol-cognite: It would be great if you can review from start since it is changed a bit.

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
```diff
@@ -0,0 +1,191 @@
...
+      .flatTap(_ => incMetrics(itemsDeleted, relationshipIds.length))
+  }
+
+  override def upsert(rows: Seq[Row]): IO[Unit] =
+    throw new CdfSparkException("Upsert is not supported for relationships")
+
+  override def update(rows: Seq[Row]): IO[Unit] =
+    throw new CdfSparkException("Update is not supported for relationships")
+
+  def relationshipToRelationshipReadSchema(relationship: Relationship): RelationshipsReadSchema =
```
emelvarol-cognite, line 113: This is only needed for labels, if there is a better way to do this conversion without relationshipToRelationshipReadSchema function, that would be great. This is only needed for labels.

It uses the same type for all except for labels. Labels come as  ```{"labels": [{external_id: "label"}, {"external_id": "label2"}]}``` on Relationship objects. I don't know why API is designed like this since labels only have external ids. It could be simply ```{labels: ["label1", "label2"]}``` and it would make this whole work a one day thing. I wanted to be fast so I tried to implement this without understanding completely. After spending a long time and seeing that will not work , I decided to understand and found this conversion idea. This could be messy when adding assets labels field. Because there we have update/upsert.

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
```diff
@@ -0,0 +1,191 @@
...
+      endTime = relationship.endTime,
+      confidence = relationship.confidence,
+      labels = cogniteExternalIdSeqToStringSeq(relationship.labels),
+      createdTime = relationship.createdTime,
+      lastUpdatedTime = relationship.lastUpdatedTime,
+      dataSetId = relationship.dataSetId
+    )
+
+  def relationshipInsertSchemaToRelationshipCreate(
+      relationship: RelationshipsInsertSchema): RelationshipCreate =
```
emelvarol-cognite, line 130: same as above.

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
emelvarol-cognite: For now, I pass null to labels on AssetsRelation. It will be a separate task to make it available for assets.

emelvarol-cognite: df.where("labels in(array('test_label'))").show() or df.where("labels = array('test_label')").show()  is working with this workaround. Also, it is possible to provide a list of labels when creating relationships.

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
```diff
@@ -0,0 +1,191 @@
...
+import org.apache.spark.sql.types._
+import org.apache.spark.sql.{Row, SQLContext}
+
+class RelationshipsRelation(config: RelationConfig)(val sqlContext: SQLContext)
+    extends SdkV1Relation[RelationshipsReadSchema, String](config, "relationships")
+    with InsertableRelation
+    with WritableRelation {
+  import CdpConnector._
+
+  override def schema: StructType = structType[RelationshipsReadSchema]
```
emelvarol-cognite, line 31: Schema is not created for "Relationship" because of the reason mentioned below. Interface to user will be RelationshipsReadSchema.

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
```diff
@@ -367,13 +370,14 @@ class AssetsRelationTest extends FlatSpec with Matchers with ParallelTestExecuti
                 |lastUpdatedTime,
                 |0 as rootId,
                 |null as aggregates,
+                |null as labels,
```
emelvarol-cognite, line 373: labels is a nullable field and I did not add this into schemas here but it is available in scala sdk. If I don't add "null as labels" in some test queries, it gives an error. As a user in transformations, I was skipping some nullable parameters in my queries. That was working there, how do we handle this when user skips some? 

I am thinking if this will cause some compatibility issues. If so, I would prefer deleting "labels" from Assets in scala-sdk and take it as a separate PR along with spark part. I don't prefer to add it to spark-datasource in this PR since it would take time (because this was the hard part of all).

emelvarol-cognite, line 373: labels is a nullable field and I did not add this into schemas here but it is available in scala sdk with my merged PR. If I don't add "null as labels" in some test queries, it gives an error. As a user in transformations, I was skipping some nullable parameters in my queries. That was working there, do we handle this in jetfire when user skips some?

I am thinking if this will cause some compatibility issues. If so, I would prefer deleting "labels" from Assets in scala-sdk and take it as a separate PR along with spark part. I don't prefer to add it to spark-datasource in this PR since it would take time (because this was the hard part of all and I have several days to work in this sprint, I want to finish relationships primarily).

emelvarol-cognite: @hakontro reminder

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
hakontro: So this would have worked with `Relationships` if this https://github.com/cognitedata/cdp-spark-datasource/pull/476 was merged, but it's still a draft and I don't want to mix the two topics into this PR so let's do it like this for now and then change it when/if we get #476 merged.

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
hakontro: I think it should be fine, can you try to add some labels to the tests and see if they still work?

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
```diff
@@ -73,6 +74,41 @@ class AssetsRelationTest extends FlatSpec with Matchers with ParallelTestExecuti
     assert(assetsRead == 1)
   }
 
+  it should "support pushdown filters on labels" taggedAs WriteTest in {
```
emelvarol-cognite, line 77: @hakontro  This test passes if I only run AssetsRelationsTest but if I run all tests, it fails at line 91. Would you have a guess regarding the reason? 

I couldn't spot any part affecting this in the other tests, I thought maybe it was because of parallel testing. But I am sure that this sample is created and can be queried as follows according to my experiments in ammonite.

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -94,19 +99,64 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
   }
 
   override def getFromRowsAndCreate(rows: Seq[Row], doUpsert: Boolean = true): IO[Unit] = {
-    val assets = rows.map(fromRow[AssetCreate](_))
+    val assetsUpserts = rows.map(fromRow[AssetsUpsertSchema](_))
+    val assets = assetsUpserts.map(assetUpsertSchemaToAssetCreate)
```
emelvarol-cognite, line 103: to make sure we skip labels in upsert (maybe for now). labels are available for create/read.

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
emelvarol-cognite: I added support for create and read operations. I saved upsert/update for later since it is not a setter. We can add/remove labels so I need to think how we want to do it and it can come later separately.  Added a couple of changes to the tests to address these changes. I left a question there in the tests.

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
emelvarol-cognite: It worked this time, is it about default query limit? I saw a similar unpredictable behaviour in one of the timeseries tests (My changes does not affect it actually). I will try to reproduce it.

emelvarol-cognite: Error I sometimes get in TimeseriesRelationsTest:

[info] - should support upsert in savemode *** FAILED ***
[info]   429 did not equal 404 (TimeSeriesRelationTest.scala:846)

src/main/scala/cognite/spark/v1/RelationshipsRelation.scala
emelvarol-cognite: It would be great to have this change. I guess it is not ideal to implement a conversion for all types we have labels (if all types will have it at some point).

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
hakontro: This is related to eventual consistency.. sometimes the result of a change to data in CDF won't be present in what you read back if you read it back immediately.. i.e if can create an asset and then immediately try to read it you'll sometimes get a result without that asset. This is because it takes time for servers in the CDF backend to synchronize. For this reason we do retries in our tests, check out the `retryWhile` function if you haven't already.

Anyways, this sometimes fails some of our tests anyways, which is what you're seeing here.

src/test/scala/cognite/spark/v1/RelationshipsRelationTest.scala
```diff
@@ -0,0 +1,348 @@
+package cognite.spark.v1
+
+import java.time.Instant
+
+import cognite.spark.v1.SparkSchemaHelper.fromRow
+import com.cognite.sdk.scala.v1.{CogniteExternalId, RelationshipCreate, RelationshipsFilter}
```
hakontro, line 6: `RelationshipsFilter` can be removed from here.

hakontro (approved):
Awesome! :1st_place_medal: This looks really good now, and great work on the tests! I made some small comments, but they are just nitpicks anyways so feel free to merge after addressing them.

wjoel:
Awesome indeed! Given the very large number of commits here, I would suggest squashing some/most/all commits either as part of the rebase, or doing "squash and merge" for this one (even though we typically don't).

emelvarol-cognite:
> Awesome indeed! Given the very large number of commits here, I would suggest squashing some/most/all commits either as part of the rebase, or doing "squash and merge" for this one (even though we typically don't).

Thanks! I will definitely do that. I committed a lot in this very first PR (unnecessarily) :D but I will do reasonable number of commits in the next one.

emelvarol-cognite:
> Awesome! 🥇 This looks really good now, and great work on the tests! I made some small comments, but they are just nitpicks anyways so feel free to merge after addressing them.

Thanks a lot! I will make changes accordingly. I was having a look at jetfire that's why I could not start.

emelvarol-cognite:
> Awesome indeed! Given the very large number of commits here, I would suggest squashing some/most/all commits either as part of the rebase, or doing "squash and merge" for this one (even though we typically don't).

Squash and merge is not enabled in this repository and I think I can't change it. Would it be ok if you enable it just for this? 

This PR was created a long time ago and then I stopped working on this. In the meantime, I guess master changed a bit. When I return back to finish this,  rebasing became very problematic.

PR #479, opened 2020-11-24
Add support for OAuth2 Client Credentials flow

sebbert:
This PR is waiting on https://github.com/cognitedata/cognite-sdk-scala/pull/271


src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -264,21 +265,39 @@ object DefaultSource {
...
+    val apiKey = parameters.get("apiKey").map(apiKey => ApiKeyAuth(apiKey))
+    val clientCredentials = for {
+      tokenUri <- parameters.get("tokenUri").map { tokenUriString =>
+        Uri
+          .parse(tokenUriString)
+          .getOrElse(throw new CdfSparkIllegalArgumentException("Invalid URI in tokenUri parameter"))
+      }
+      clientId <- parameters.get("clientId")
+      clientSecret <- parameters.get("clientSecret")
+      scopes <- parameters.get("scopes").map(_.split(" ").toList)
```
hakontro, line 279: Didn't think about it before now, but does the new auth setup require that people _have_ to list scopes every time they use their credentials? Is this to be able to grant a subset of your own access, or can it be left blank and still work?

src/test/scala/cognite/spark/v1/OAuth2Test.scala
```diff
@@ -0,0 +1,30 @@
+package cognite.spark.v1
+
+import org.scalatest.{FlatSpec, LoneElement, Matchers, ParallelTestExecution}
```
hakontro, line 3: `LoneElement` isn't used here (I think)

src/test/scala/cognite/spark/v1/SparkTest.scala
```diff
@@ -6,15 +6,15 @@ import java.util.concurrent.Executors
...
+import com.cognite.sdk.scala.common.{ApiKeyAuth, Auth, AuthProvider}
 import org.apache.spark.sql.SparkSession
 import org.scalatest.Tag
 import org.apache.spark.datasource.MetricsSource
 
 import scala.concurrent.{ExecutionContext, TimeoutException}
 import scala.concurrent.duration._
 import scala.util.Random
-import cats.effect.{IO, Timer}
+import cats.effect.{Clock, IO, Timer}
```
hakontro, line 17: `Clock` and `Auth` are maybe no longer needed.

hakontro (commented):
Looks great so far! Is there a way for us to write tests that need to refresh the tokens?

mathialo:
> Is there a way for us to write tests that need to refresh the tokens?

For the python extractor utils [we mocked everything](https://github.com/cognitedata/python-extractor-utils/blob/master/tests/tests_unit/test_authentication.py), which means you could include a very short TTL in the mocked response.

src/main/scala/cognite/spark/v1/DefaultSource.scala
sebbert: It needs to be specified, as far as I know. See https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-permissions-and-consent#client-credentials-grant-flow-and-default

sebbert: This seems like a better explanation: https://dev.to/425show/just-what-is-the-default-scope-in-the-microsoft-identity-platform-azure-ad-2o4d

The "When must I use the /.default scope?" section in particular.

sebbert: So basically:
- We can't leave it blank
- We could automatically generate it based on the baseUrl, but:
  - It would only apply to azure
  - While the `{resource}` is a URL in our case, it does not have to be... I think

We will need to have some documentation on how to set this up for azure anyway (how to format the token endpoint URL, etc.), so I'm kinda leaning towards not doing too much nonstandard magic on behalf of the user, in exchange for the user having to do a bit more work.

src/test/scala/cognite/spark/v1/OAuth2Test.scala
sebbert: Good catch, thanks!

sebbert:
>Is there a way for us to write tests that need to refresh the tokens?
Yes, as @mathialo pointed out, we can mock the token endpoint and provide a short TTL.
But the question is how. We can't just create a local mock sttp backend, because we can't pass objects to Spark (as it may not be executing on the same JVM), so we'd have to run an actual HTTP server. Definitely possible, but we'd need to add a dependency on http4s or some other HTTP server library in our tests, and then also provide it with a valid bearer token somehow.

I think a good first step is to test token refresh in the Scala SDK, as we'd circumvent a lot of these issues.

src/main/scala/cognite/spark/v1/DefaultSource.scala
wjoel: Since Azure would be most of our users, having a default generated based on "baseUrl" but allowing it to be overridden via "scopes" seems reasonable.

hakontro (approved):
Awesome, LGTM! I liked Joel's suggestion of defaulting the scopes from baseUrl, so let's do that and then we can merge.

PR #482, opened 2021-01-11
Add labels for asset updates.

emelvarol-cognite:
**Ready for review**
Tests for Assets and AssetHierarchy are successful and I tried update/upsert with labels for both of them in spark-shell and jetfire. 

 Needs: https://github.com/cognitedata/cognite-sdk-scala/pull/289

### DIFF
--- a/src/main/scala/cognite/spark/v1/AssetsRelation.scala
+++ b/src/main/scala/cognite/spark/v1/AssetsRelation.scala
@@ -6,11 +6,10 @@ import cats.effect.IO
 import cats.implicits._
 import cognite.spark.v1.PushdownUtilities._
 import cognite.spark.v1.SparkSchemaHelper._
-import com.cognite.sdk.scala.common.{WithExternalId, WithId}
+import com.cognite.sdk.scala.common._
 import com.cognite.sdk.scala.v1.resources.Assets
 import com.cognite.sdk.scala.v1._
 import fs2.Stream
-import io.scalaland.chimney.Transformer
 import io.scalaland.chimney.dsl._
 import org.apache.spark.sql.sources.{Filter, InsertableRelation}
 import org.apache.spark.sql.types._
@@ -45,7 +44,13 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
     // Merge streams related to each partition to make sure duplicate values are read into
     // the same RDD partition
     streamsPerFilter.transpose
-      .map(s => s.reduce(_.merge(_)).map(toAssetReadSchema))
+      .map(
+        s =>
+          s.reduce(_.merge(_))
+            .map(
+              _.into[AssetsReadSchema]
+                .withFieldComputed(_.labels, u => cogniteExternalIdSeqToStringSeq(u.labels))
+                .transform))
   }
 
   private def assetsFilterFromMap(m: Map[String, String]): AssetsFilter =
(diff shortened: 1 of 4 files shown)


src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -115,6 +116,43 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+      .withFieldComputed(
+        _.labels,
+        u =>
+          u.labels match {
+            case None => None
+            case _ => Some(LabelsOnUpdate(add = Some(u.labels.get.map(CogniteExternalId))))
+        })
+      .buildTransformer
+
+  implicit val createToUpdateTransformer: Transformer[AssetCreate, AssetUpdate] =
```
emelvarol-cognite, line 131: @hakontro Getting some serialization exceptions pointed here, would you have a guess about the reason? :) LabelsOnUpdate encoder is defined within Assets class in scala sdk.

```

Caused by: java.io.NotSerializableException: cognite.spark.v1.AssetsRelation$$anon$4
Serialization stack:
        - object not serializable (class: cognite.spark.v1.AssetsRelation$$anon$4, value: cognite.spark.v1.AssetsRelation$$anon$4@66633170)
        - field (class: cognite.spark.v1.AssetsRelation, name: createToUpdateTransformer, type: interface io.scalaland.chimney.Transformer)
        - object (class cognite.spark.v1.AssetsRelation, cognite.spark.v1.AssetsRelation@154a0cba)
        - element of array (index: 2)
        - array (class [Ljava.lang.Object;, size 3)
        - field (class: java.lang.invoke.SerializedLambda, name: capturedArgs, type: class [Ljava.lang.Object;)

```

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -115,6 +116,43 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+        _.labels,
+        u =>
+          u.labels match {
+            case None => None
+            case _ => Some(u.labels.get.map(CogniteExternalId))
+          })
+      .withFieldComputed(_.name, u => u.name.get)
+      .buildTransformer
+
   def toAssetReadSchema(a: Asset): AssetsReadSchema =
```
emelvarol-cognite, line 156: I will revisit these transformer like functions to see if they can be Transformer.

src/main/scala/cognite/spark/v1/AssetsRelation.scala
emelvarol-cognite: Hmm maybe I am overriding Setter transformers?

emelvarol-cognite: I feel like this is about transforming into "Setter". I hoped it would grab Setter Transformer from dataTypes.scala when needed but I think it does not work like that?

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -6,7 +6,7 @@ import cats.effect.IO
 import cats.implicits._
 import cognite.spark.v1.PushdownUtilities._
 import cognite.spark.v1.SparkSchemaHelper._
-import com.cognite.sdk.scala.common.{WithExternalId, WithId}
+import com.cognite.sdk.scala.common._
```
emelvarol-cognite, line 9: necessary to be able to import LabelsOnUpdate transformers from scala-sdk/package.scala.

mathialo (approved):
Nice!

PR #485, opened 2021-01-29
Ignore uncaught exceptions from CDF-Spark-Datasource threads

wjoel:


src/main/scala/cognite/spark/v1/CdpConnector.scala
```diff
@@ -21,14 +22,23 @@ final case class Error[A](error: A)
...
     ExecutionContext.fromExecutor(
       Executors.newFixedThreadPool(
         Math.max(Runtime.getRuntime.availableProcessors(), 4) * 2,
-        new ThreadFactoryBuilder().setDaemon(true).setNameFormat("CDF-Spark-Datasource-%d").build()
+        new ThreadFactoryBuilder()
+          .setDaemon(true)
+          .setUncaughtExceptionHandler(new UncaughtExceptionHandler {
+            override def uncaughtException(t: Thread, e: Throwable): Unit =
+              // Log the exception, and move on.
+              logger.warn(e)("Ignoring uncaught exception")
```
mathialo, line 38: Just to be sure I'm understanding this correctly, the `(e)` means we'll get stack traces and other exception details logged, right?

mathialo (approved):

src/main/scala/cognite/spark/v1/CdpConnector.scala
wjoel: Indeed!

PR #488, opened 2021-02-04
Interrupt streams and iterators when tasks are killed/completed 

wjoel:


hakontro (approved):
Looks like a good improvement! Perhaps you can add some comments so we remember why in the future!

PR #493, opened 2021-05-06
Allow writing rows to multiple sequences by externalId

hakontro:


src/main/scala/cognite/spark/v1/SequenceRowsRelation.scala
```diff
@@ -187,10 +199,15 @@ class SequenceRowsRelation(config: RelationConfig, sequenceId: CogniteId)(val sq
...
-        .flatTap(_ => incMetrics(itemsCreated, rows.length))
+      import cats.instances.list._
+      projectedRows
+        .groupBy(_.externalId)
+        .toList
+        .traverse(
+          pair =>
+            client.sequenceRows
+              .insertByExternalId(pair._1.toString, columns, pair._2)
+              .flatTap(_ => incMetrics(itemsCreated, rows.length))) *> IO.unit
```
exyi, line 210: Nitpick

```suggestion
        .traverse {
          case (externalId, rows) =>
            client.sequenceRows
              .insertByExternalId(externalId.toString, columns, rows)
              .flatTap(_ => incMetrics(itemsCreated, rows.length)) } *> IO.unit
```

exyi (approved):
I like this :)

Technically, we could cause someone trouble by adding a required field, but if there is not that many direct Spark users it's probably not a problem.

I think we should mention this somewhere in the docs, but I don't know what is the proper place - the current examples are so vague :sweat_smile:

src/main/scala/cognite/spark/v1/SequenceRowsRelation.scala
hakontro: Awesome, that's the one I was looking for!
Weirdly, `(externalId, rows)` typechecks, but then everything is in `externalId` and the `rows` is just junk I think

emelvarol-cognite (approved):
Awesome, I don't have anything to add besides id support but it is not a big deal.

PR #496, opened 2021-05-26
Add support for auth-ticket in spark options

exyi:


hakontro (approved):
LGTM! Would love a test, but I guess that's difficult.

PR #499, opened 2021-06-09
Update to Cats Effect 3

wjoel:
This now works (hopefully), as we shade cats and such and publish fat JARs.

~~This doesn't work, unfortunately, because Spark includes cats-kernel
2.0.0-M4, which is not binary compatible with non-milestone releases
of cats-kernel 2.x, which is required by Cats Effect 3.~~

~~It also doesn't work because circe 0.14 and later depends on a version
of shapeless that is incompatible with the one included in Spark.~~

~~This PR currently shows how to build a fat JAR to get around that, but
that's not a viable solution for production use.~~


src/main/scala/cognite/spark/v1/CdpConnector.scala
```diff
@@ -75,7 +80,15 @@ object CdpConnector {
             Some(MetricsSource.getOrCreateCounter(metricsPrefix, s"requests.response.failure"))
         ))
     // this backend throttles when rate limiting from the serivce is encountered
-    val throttledBackend = new BackpressureThrottleBackend[IO, Any](metricsBackend, 800.milliseconds)
+    val makeQueueOf1 = for {
+      queue <- Queue.bounded[IO, Unit](1)
+      _ <- queue.offer(())
+    } yield queue
```
exyi, line 86: It feels a bit wrong that the queue is a parameter, but I understand that using unsafeRunSync in the constructor is also not nice...

exyi (approved):
As far as I understand it, cats is shaded so this won't affect Jetfire, right? If yes, then let's try this right away :)

src/main/scala/cognite/spark/v1/CdpConnector.scala
wjoel: Yeah, the alternative is to make this require an `IORuntime` so it can do `unsafeRunSync` in the constructor by itself, but I prefer this.
The single-element queue thing is from https://typelevel.org/cats-effect/docs/migration-guide#mvar

wjoel:
> As far as I understand it, cats is shaded so this won't affect Jetfire, right? If yes, then let's try this right away :)

I'm a bit uncertain if we also need to update Jetfire to cats-effect 3 at the same time, but I have https://github.com/cognitedata/scala-microservice/pull/214 ready to make that possible.

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
```diff
@@ -45,7 +45,7 @@ final case class NoPushdown() extends PushdownExpression
 
 object PushdownUtilities {
   def pushdownToUri(parameters: Seq[Map[String, String]], uri: Uri): Seq[Uri] =
-    parameters.map(params => uri.params(params))
+    parameters.map(params => uri.addParams(params))
```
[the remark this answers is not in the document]
wjoel: It looks like it can be removed entirely, as the tests just test "uris.length" which will always be the same as "params.length"...?

wjoel:
> FYI, catching raw `Throwable` is dangerous, you should check if it is `NonFatal`. https://www.scala-lang.org/api/2.13.3/scala/util/control/NonFatal$.html
> 
> More idiomatic and safer would be to write instead:
> 
> ```scala
> IO(operation(chunk.toVector)).flatten
> ```

Thanks, good point! It felt wrong typing `Throwable` and I should have stopped myself. Although `IO(..).flatten` might be more idiomatic, I think it's less clear for developers new to cats-effect, and we're still a growing company. Hopefully we won't have to keep the workarounds for much longer, anyway.

PR #500, opened 2021-06-17
Add ignoreNullFields option to allow writing NULL into CDF

exyi:
the option controls our behavior when NULL is specified.
By default (when ignoreNullFields = true), NULL means "do not update this field",
with ignoreNullFields = false, NULL will be actually written into CDF.

Currently only adds support into EventsRelation for demonstration.

I'd like to ask for review of this intermediate step. Do you think this is the right approach to the problem?


macro/src/main/scala/SparkSchemaHelper.scala
```diff
@@ -66,16 +67,17 @@ class SparkSchemaHelperImpl(val c: Context) {
       val constructor = structType.decl(termNames.CONSTRUCTOR).asMethod
       val params = constructor.paramLists.flatten.map((param: Symbol) => {
         val name = param.name.toString
-        val isOuterOption = param.typeSignature <:< optionType
-        val isOptionNonNullableSetter = param.typeSignature <:< optionNonNullableSetter
-
-        val isOptionSetter = param.typeSignature <:< optionSetter
-        val innerType = if (isOptionSetter || isOptionNonNullableSetter) {
-          param.typeSignature.typeArgs.head.typeArgs.head
```
hakontro, line 74: Do you know what this used to be for/how it was used?

src/test/scala/cognite/spark/v1/EventsRelationTest.scala
```diff
@@ -478,6 +476,81 @@ class EventsRelationTest extends FlatSpec with Matchers with ParallelTestExecuti
     }
   }
 
+  it should "allow udpates NULL in savemode" taggedAs WriteTest in {
```
hakontro, line 479:
```suggestion
  it should "allow NULL udpates in savemode" taggedAs WriteTest in {
```

src/main/scala/cognite/spark/v1/CdfRelation.scala
```diff
@@ -29,4 +30,13 @@ abstract class CdfRelation(config: RelationConfig, shortName: String)
         counter.inc(count)
       }
     )
+
+  implicit def fieldToSetter[T: Manifest]: Transformer[OptionalField[T], Option[Setter[T]]] = {
+    case FieldSpecified(null) => // scalastyle:ignore null
+      throw new Exception("FieldSpecified(null) observed, that's bad.")
```
hakontro, line 36:
```suggestion
      throw new Exception("FieldSpecified(null) observed, that's bad. Please reach out to Cognite support.")
```

hakontro (commented):
Looks really good so far, especially like that this implementation is opt-in for the user while the old behavior doesn't change.
Do you want to finish this up and start a new PR or keep going?

src/main/scala/cognite/spark/v1/EventsRelation.scala
```diff
@@ -133,16 +132,16 @@ object EventsRelation extends UpsertSchema {
...
-    description: Option[String] = None,
-    `type`: Option[String] = None,
-    subtype: Option[String] = None,
+    startTime: OptionalField[Instant] = FieldNotSpecified,
+    endTime: OptionalField[Instant] = FieldNotSpecified,
+    description: OptionalField[String] = FieldNotSpecified,
+    `type`: OptionalField[String] = FieldNotSpecified,
+    subtype: OptionalField[String] = FieldNotSpecified,
     metadata: Option[Map[String, String]] = None,
     assetIds: Option[Seq[Long]] = None,
```
emelvarol-cognite, line 141: Are these not nullable?

macro/src/main/scala/SparkSchemaHelper.scala
exyi: I think it was used before and it was not removed. It's definitely not used anymore, I added a check and since it's a macro it would throw compile-time.

src/main/scala/cognite/spark/v1/EventsRelation.scala
```diff
@@ -133,16 +132,16 @@ object EventsRelation extends UpsertSchema {
...
-    source: Option[String] = None,
-    externalId: Option[String] = None,
-    dataSetId: Option[Long] = None
+    source: OptionalField[String] = FieldNotSpecified,
+    externalId: Option[String] = None, // TODO: nullable externalId
```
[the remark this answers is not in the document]
exyi: Because we have to implement the WithExternalId trait. I will have to change the Scala SDK slightly to allow using the OptionalField there.

src/main/scala/cognite/spark/v1/EventsRelation.scala
exyi: these two are of type `NonNullableSetter`, they can not be updated to be null. You will want to update them with an empty collection to clear it.

emelvarol-cognite: I get it thanks 🚀

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -91,7 +91,7 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
   override def upsert(rows: Seq[Row]): IO[Unit] = {
     val assets = rows.map(fromRow[AssetsUpsertSchema](_))
     val (itemsToUpdate, itemsToUpdateOrCreate) =
-      assets.partition(r => r.id.exists(_ > 0) || (r.name.isEmpty && r.externalId.nonEmpty))
+      assets.partition(r => r.id.exists(_ > 0) || (r.name.isEmpty && r.getExternalId().nonEmpty))
```
hakontro, line 94: Any reason for the preference on `getExternalId()` over `externalId` here?

hakontro (commented):
Great stuff. Can you add an update to the Readme.md as well?

src/main/scala/cognite/spark/v1/AssetsRelation.scala
exyi: Because externalId contains the `OptionalField` while getExternalId is converted to Option

README.md
```diff
@@ -177,6 +177,13 @@ This is based on the assumption that upserts for the same `id` or `externalId`
 will have the same values. If you have a use case where this is not the case,
 please let us know.
 
+Another option available for `.save()` writes is `.option("ignoreNullFields", "false")`. By default,
+the Data Source ignores NULLs when updating -- nothing is written to CDF when there is NULL
+in the field. This is usually useful for ignoring the columns which are irrelevant for a task,
+however it makes it impossible to null a field in the CDF. When the `ignoreNullFields` option is set
+to `false`, NULLs are written to CDF (when possible). Fields which are not even specified in the rows
+are still ignored.
+
```
hakontro, line 186:
```suggestion
#### Handling NULLs and empty fields
By default, the Spark Datasource ignores `NULL`s when updating: nothing is written to CDF when there is a `NULL`
in the field. This can be controlled by setting the `ignoreNullFields` option which defaults to `"true"` when using `.save()` writes. This is usually useful for ignoring the columns which are irrelevant for a task,
however it makes it impossible to null a field in CDF. When the `ignoreNullFields` option is set
to `false`, NULLs are written to CDF (when possible). Fields which are not specified are still ignored.
```

hakontro (approved):
LGTM :rocket:

PR #501, opened 2021-06-23
Fix leaking threads by creating many sttpBackends

exyi:
Hopefully...


hakontro (commented):
Neat. What is the difference here?

emelvarol-cognite (commented):
Reminder: bump the version

exyi:
> Neat. What is the difference here?

This way we use the single sttpBackend from CdpConnector, don't create new one for each request. Creating sttpBackend also creates a threadpool which contains some threads and thus leaks them

hakontro (approved):
LGTM

PR #502, opened 2021-06-30
Fix NULL updates in upsert mode

exyi:
The problem was that UpsertSchema was converted to Create object and then
to update, which looses information. I have adjusted the genericUpsert
helpers to convert from UpsertSchema directly to Create or Update object


emelvarol-cognite (approved):
Awesome fix 🚀 Do not forget to bump the version

hakontro:
> is it ok to merge and release this?

I'm thinking: yes!

PR #503, opened 2021-07-08
Add RateLimitingBackend to limit number of concurrent requests

exyi:
It only limits the number of requests from one relation - we can not
really synchronize parallel executors in the number of requests. However,
since the retrying backend is inside the rate limiting backend, possible
retries on 503 will block the semaphore and cause reduction in the number of
parallel requests.


src/main/scala/cognite/spark/v1/Constants.scala
```diff
@@ -15,11 +15,12 @@ object Constants {
   val DefaultDataPointsLimit = 100000
   val DefaultSequencesLimit = 10000
   val DefaultPartitions = 200
+  val DefaultMaxParallelRequests = 200
   val DefaultDataPointsPartitions = 20
   val DefaultParallelismPerPartition = 10
   val DefaultInitialRetryDelay: FiniteDuration = 150.millis
   val DefaultMaxBackoffDelay: FiniteDuration = 120.seconds
-  val DefaultBaseUrl = "https://api.cognitedata.com"
+  val DefaultBaseUrl = "http://localhost:4001"
```
hakontro, line 23: This was probably a temporary solution?

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -19,6 +19,7 @@ final case class RelationConfig(
     partitions: Int,
     maxRetries: Int,
     maxRetryDelaySeconds: Int,
+    maxParallelRequests: Int,
```
hakontro, line 22: Let's document this in the Readme.

hakontro (commented):
Very cool! I'm a bit on the fence, this is making changes to the sttpBackend for all of our Relations, while AssetHierarchy is the only one that has seen problems with this AFAIK. Also, I think we've identified the particular code path that can lead to the issue, and I think it's simpler if we just make sure to set a maximum batch size in there.

src/main/scala/cognite/spark/v1/RateLimitingBackend.scala
```diff
@@ -0,0 +1,37 @@
...
+) extends SttpBackend[F, S] {
+  override def send[T](
+      request: Request[T, S]
+  ): F[Response[T]] =
+    // This will allow only the specified number of request run in parallel
+    // While it's not the smartest way to react to rate limiting, this will
+    // actually respond to backpressure when used with the RetryingBackend:
+    // The 503 rate limit response is retried, but with a delay. It will also
+    // block the semaphore for new requests, so it will reduce the rate
+    semaphore.withPermit(delegate.send(request))
```
hakontro, line 22: Can the semaphore leak `acquire`s if we get issue/pausing downstream?

exyi: It shouldn't :sweat_smile:

README.md
```diff
@@ -92,6 +92,7 @@ The common options are:
...
 | `maxRetries`  | The maximum number of retries to be made when a request fails. Default: 10                                                                                                                                                                                                                                        |                                            |
 | `maxRetryDelay` | The maximum number of seconds to wait between retrying requests. Default: 30                                                                                                                                                                                                                                        |                                            |
+| `maxParallelRequests` | The maximum number of requests send to CDF in parallel **per partition**. Reduce this number if running into rate limiting issues. Default: 200                                                                                                                                                                                                                                        |                                            |
```
hakontro, line 95: This sounds a lot like `DefaultParallelismPerPartition`, perhaps we can use that instead?

PR #504, opened 2021-07-16
Update to Scala SDK 1.5

exyi:
This is a bit complex update of many different dependencies,
most notably STTP. Since Scala SDK no longer uses Chimney, we redeclare
the transformers that we need.

I have included the metrics and rate limiting backend (by accident...), but
I don't want to revert it since it also depends on Sttp, so it's only disabled


src/main/scala/cognite/spark/v1/CdfRelation.scala
```diff
@@ -31,16 +30,31 @@ abstract class CdfRelation(config: RelationConfig, shortName: String)
...
+  // Until Scala SDK adds back support for Chimney
+  implicit def fieldToSetter[T: Manifest]: Transformer[OptionalField[T], Option[Setter[T]]] = {
+    case FieldSpecified(null) => // scalastyle:ignore null
+      throw new Exception(
+        "FieldSpecified(null) observed, that's bad. Please reach out to Cognite support.")
+    case FieldSpecified(x) => Some(SetValue(x))
+    case FieldNotSpecified => None
+    case FieldNull if config.ignoreNullFields => None
+    case FieldNull => Some(SetNull())
+  }
```
emelvarol-cognite, line 42: what is the difference here?

macro/src/main/scala/StructTypeEncoder.scala
```diff
@@ -39,10 +39,7 @@ case object FieldNull extends OptionalField[Nothing] {
 }
 
 object OptionalField {
-  implicit def fieldToOption[T: Manifest]: Transformer[OptionalField[T], Option[T]] =
-    new Transformer[OptionalField[T], Option[T]] {
-      override def transform(src: OptionalField[T]): Option[T] = src.toOption
-    }
+  implicit def fieldToOption[T: Manifest]: Transformer[OptionalField[T], Option[T]] = _.toOption
```
emelvarol-cognite, line 42: also what difference does this make 😓

src/main/scala/cognite/spark/v1/CdfSparkAuth.scala
```diff
@@ -2,30 +2,29 @@ package cognite.spark.v1
...
-import com.softwaremill.sttp.SttpBackend
-import com.softwaremill.sttp.asynchttpclient.cats.AsyncHttpClientCatsBackend
+import sttp.client3.SttpBackend
 
 sealed trait CdfSparkAuth extends Serializable {
   def provider(
       implicit cs: ContextShift[IO],
       clock: Clock[IO],
-      sttpBackend: SttpBackend[IO, Nothing]): IO[AuthProvider[IO]]
+      sttpBackend: SttpBackend[IO, Any]): IO[AuthProvider[IO]]
```
emelvarol-cognite, line 11: is this required by new sttp version?

macro/src/main/scala/StructTypeEncoder.scala
exyi: It's just shorter. Sorry for doing mess in the PR, I just wanted to celebrate that we have Scala 2.12 now :sweat_smile:

src/main/scala/cognite/spark/v1/CdfSparkAuth.scala
exyi: Yes, the new Sttp version has a different model of the second type parameter. I don't really understand what does it do, but Scala SDK also switched everything to Any from Nothing.

emelvarol-cognite (approved):
LGTM, thanks for fixing this!
I am not sure if we need to add `maxParallelRequests` to readme as you disabled it. 🤔 
And remember to bump the version

exyi:
Håkon was more successful with this :sweat_smile:

PR #507, opened 2021-08-09
Release version 1.4.27

hakontro:


build.sbt
```diff
@@ -14,7 +14,7 @@ val circeVersion: Option[(Long, Long)] => String = {
 val sttpVersion = "1.7.2"
 val Specs2Version = "4.2.0"
 val artifactory = "https://cognite.jfrog.io/cognite/"
-val cogniteSdkVersion = "1.4.6"
+val cogniteSdkVersion = "1.5.3"
```
wjoel, line 17: Hmm, I think I bumped sttp too far in some update (making some dependency incompatible with Spark), I wonder if that has been reverted?

wjoel: And/or Circe, we need to stay on circe 0.13.0.

hakontro: Aha, that makes sense. That was related to Scala3, right? Or can we revert the sdk to be spark compatible without breaking stuff?

wjoel: I found out through the Scala 3 work, which hasn't been (and shouldn't be) merged, but that version update was done in an earlier commit, I think. I'll go look for it.

wjoel: https://github.com/cognitedata/cognite-sdk-scala/pull/356

hakontro:
Turns out it was running with Java 8 that was causing this. Probably some package update that didn't work with j8 anymore?

emelvarol-cognite (approved):

PR #508, opened 2021-08-12
Run asset hierarchy on executors

exyi:
essentially use .repartition(1).foreachPartition instead of collect.
Although this will not enable parallelism it will not run on the driver.
In Jetfire this seemed to overload the driver in some cases, so
we'll move this work to one of the executors, which can be scaled easily


exyi:
Done in #510

PR #510, opened 2021-08-15
asset hierarchy: batch mutiple subtrees together 

exyi:
this fixes part of the problem that we fired at least 3 request
for each subtree, which was a lot for flat hierarchies

This patch is not complete fix of the problem as it does not batch fetching of
existing subtrees

It changes the behavior of AsserHierarchyBuilder slightly:
* assets.created metric is not even created when no asset is created. This is
   consistent with how the other metrics work.
* error message of moving asset between subtrees also includes externalId of
   the subtree root. No information is missing compared to previous version.

### DIFF
--- a/src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
+++ b/src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
@@ -6,7 +6,7 @@ import cats.implicits._
 import cats.syntax._
 import cognite.spark.v1.PushdownUtilities.stringSeqToCogniteExternalIdSeq
 import cognite.spark.v1.SparkSchemaHelper.{fromRow, structType}
-import com.cognite.sdk.scala.common.{CdpApiException, SetValue}
+import com.cognite.sdk.scala.common.{CdpApiException, SetNull, SetValue, Setter}
 import com.cognite.sdk.scala.v1.{
   Asset,
   AssetCreate,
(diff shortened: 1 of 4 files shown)


src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -55,45 +65,87 @@ final case class InvalidNodeReferenceException(nodeIds: Seq[String], referencedF
...
 
   import CdpConnector.cdpConnectorContextShift
 
+  def buildFromDf(data: DataFrame): Unit =
+    data
+      .repartition(numPartitions = 1)
+      .foreachPartition((rows: Iterator[Row]) => {
+        build(rows).unsafeRunSync()
+      })
+
```
hakontro, line 135: Neat

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -150,138 +202,183 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
...
+      cdfRoots <- fetchCdfRoots(trees.map(_.root.externalId))
+      upsertedRoots <- upsertRoots(trees.map(_.root), cdfRoots)
+
+      // then fetch all the subtrees to know if we should insert, update or delete them
+      // TODO: merge the subtree requests
+      cdfTrees <- trees.parTraverse(t => fetchCdfSubtree(t.root))
+      (toDelete, toInsert, toUpdate) = trees
+        .zip(cdfTrees)
+        .map((nodesToDeleteInsertUpdate _).tupled)
+        .unzip3
```
hakontro, line 217: Fancy

hakontro (approved):
LGTM :rocket:

hakontro:
Looks like those Relationships tests need to be fixed, they're failing for me locally as well..
Also, this replaces https://github.com/cognitedata/cdp-spark-datasource/pull/508, right?

exyi:
> Also, this replaces #508, right?

Ah, I forgot to switch to master before doing this... So yes, it replaces it by accident :sweat_smile: 

> Looks like those Relationships tests need to be fixed, they're failing for me locally as well..

They seem to depend on data in the test tenant. I will verify manually that the pushdown still works and adjust the tests

PR #511, opened 2021-08-18
Add rawEnsureParent option for creating RAW tables

exyi:


exyi:
We will enable this by default in Jetfire

PR #513, opened 2021-08-19
asset hierarchy: Batch fetching of existing subtrees

exyi:
This will make the AssetHierarchyBuilder actually work for
very flat and large trees, since the number of request will be about
(number of items)/100

### DIFF
--- a/src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
+++ b/src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
@@ -7,14 +7,18 @@ import cats.syntax._
 import cognite.spark.v1.PushdownUtilities.stringSeqToCogniteExternalIdSeq
 import cognite.spark.v1.SparkSchemaHelper.{fromRow, structType}
 import com.cognite.sdk.scala.common.{CdpApiException, SetValue}
-import com.cognite.sdk.scala.v1.{Asset, AssetCreate, AssetUpdate, AssetsFilter, CogniteExternalId, CogniteId, CogniteInternalId, LabelsOnUpdate}
-import io.circe.generic.semiauto.deriveDecoder
-import io.circe.{Decoder, Json, JsonObject}
+import com.cognite.sdk.scala.v1.{
+  Asset,
+  AssetCreate,
+  AssetUpdate,
+  AssetsFilter,
+  CogniteExternalId,
+  LabelsOnUpdate
+}
 import io.scalaland.chimney.dsl._
 import org.apache.spark.sql.types.{DataTypes, StructType}
 import org.apache.spark.sql.{DataFrame, Row, SQLContext}
 
-import scala.annotation.tailrec
 import scala.collection.mutable
 
 final case class AssetsIngestSchema(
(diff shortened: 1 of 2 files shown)


src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -174,86 +124,55 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
...
-            .retrieveByExternalIds(idsInGroup)
-            .adaptError({
-              case e: CdpApiException if e.code == 400 && e.missing.nonEmpty =>
-                val missingNodes = e.missing.get.map(j => j("externalId").get.asString.get).take(10)
-                val referencingNodes =
-                  missingNodes
-                  // always take the one with "lowest" externalId, so the errors are deterministic
-                    .map(missing => roots.filter(_.parentExternalId == missing).minBy(_.externalId))
-                    .map(_.externalId)
-                InvalidNodeReferenceException(missingNodes, referencingNodes)
```
exyi, line 198: We now checking for this error before we even do the insert, so no need for this complex error-parsing logic :)

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -174,86 +124,55 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
...
+      cdfRoots <- fetchCdfAssets(trees.map(_.root.externalId))
+      // fetch existing nodes in CDF in order to know if we need to insert or update them
+      cdfAssets <- fetchCdfAssets(trees.flatMap(_.nodes).map(_.externalId))
+      (toInsert, toUpdate, _) = nodesToInsertUpdate(trees.flatMap(_.nodes), cdfAssets)
+
       upsertedRoots <- upsertRoots(trees.map(_.root), cdfRoots)
 
-      // then fetch all the subtrees to know if we should insert, update or delete them
-      // TODO: merge the subtree requests
-      cdfTrees <- trees.parTraverse(t => fetchCdfSubtree(t.root))
```
exyi, line 213: This now also batched, so we get rid of the one request for each subtree :)

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -174,86 +124,55 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
...
-        case ex: CdpApiException if ex.duplicated.isDefined =>
-          val ids =
-            ex.duplicated.get
-              .flatMap(decodeCogniteId)
-              .collect { case CogniteExternalId(id) => id }
-          // If we've hit this case, the assets already exist in CDF, but under a different root asset.
-          // Since we didn't find them under the subtree, we assumed they did not exist, and attempted to to insert them,
-          // resulting in this error.
-          if (ids.nonEmpty) {
-            InvalidRootChangeException.createFromTreeList(ex, ids, trees, createdRoots)
```
exyi, line 251: We now validate for this before even doing the insert, so no need for error-parsing logic

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -270,16 +189,28 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
...
+      // list all subtrees of the tree root and filter those which are not in the ingested set
+      for {
+        idsToDelete <- batchedOperation[String, Long](
+          trees.map(_.root.externalId),
+          ids =>
+            client.assets.filter(AssetsFilter(assetSubtreeIds = Some(ids.map(CogniteExternalId))))
+              .compile.toVector
+              .map(_.filter(a => !a.externalId.exists(ingestedNodeSet.contains))
+                .map(_.id))
+        )
```
exyi, line 204: I only fetch the subtrees in the delete mode (which is not used in Jetfire). This means we'll fetch the nodes twice, but we can easily batch multiple subtrees together. For the other uses, we'd have to reaggregate them together ...

src/test/scala/cognite/spark/v1/AssetHierarchyBuilderTest.scala
```diff
@@ -1072,8 +1072,23 @@ class AssetHierarchyBuilderTest
     // Ingest the third level, 1100 subtrees
     ingest(key, grandkids, batchSize = 1000, metricsPrefix = Some(s"ingest.tree.grandkids.$key"))
 
-    // heh, this is not exactly the desired state. Just testing if the metric works
-    getNumberOfRequests(s"ingest.tree.grandkids.$key") shouldBe 1106
+    getNumberOfRowsCreated(s"ingest.tree.grandkids.$key", "assethierarchy") shouldBe 1100
+    // 2 for fetching root parents (for validation)
+    // 2 for fetching the roots
+    // 2 for creating the roots
+    getNumberOfRequests(s"ingest.tree.grandkids.$key") shouldBe 6
```
exyi, line 1079: 184-times perf improvement ;)

einarmo (approved):
I can't see anything obviously wrong. When making a major refactoring like this, the tests should probably be made to verify not just that fewer queries are made, but that the result of the queries remains the same as before.

exyi:
> When making a major refactoring like this, the tests should probably be made to verify not just that fewer queries are made, but that the result of the queries remains the same as before.

We already have quite a lot of tests that check the asset hierarchy behavior and I don't really know what else to test. If you'd like a test for specific case, definitely let me know.

einarmo:
> We already have quite a lot of tests that check the asset hierarchy behavior and I don't really know what else to test. If you'd like a test for specific case, definitely let me know.

I suppose what I was asking for was just making sure that you don't somehow produce duplicates or something else when encountering extreme cases like this. Your test proves that very wide trees produce the correct number of requests, but not that the results are actually correct.

If other trees cover this well enough, then never mind.

exyi:
> I suppose what I was asking for was just making sure that you don't somehow produce duplicates or something else when encountering extreme cases like this. Your test proves that very wide trees produce the correct number of requests, but not that the results are actually correct.

Ahaa, that's a fair point, I'll add few asserts for that :)

PR #514, opened 2021-08-31
Nice error: reading string -> numeric datapoints

exyi:
Previously this resulted in a circe error being produced or no
error in some cases.


einarmo (approved):
Looks good! You repeat yourself a bit, but I guess it is needed here?

exyi:
> You repeat yourself a bit, but I guess it is needed here?

It's always working with a different type, so it's not avoided easily. I want to get rid of some repetitiveness in this file using this Scala SDK patch: https://github.com/cognitedata/cognite-sdk-scala/pull/360. But I'd do that in a different PR.

PR #516, opened 2021-09-13
Support audience parameter for Aize OAuth2

emelvarol-cognite:


hakontro (approved):
LGTM :rocket:

src/test/scala/cognite/spark/v1/OAuth2Test.scala
```diff
@@ -49,8 +65,7 @@ class OAuth2Test
       assert(false) // fail if it works
     }
     catch {
-      case e: InvalidAuthentication => assert(true)
-      case _ => assert(false)
+      case e => assert(true)
```
exyi, line 68: Why not use the the `intercept [Exception]`? Or you can use sparkIntercept if it throws SparkException

exyi (approved):
Looks good.

However, removing the default value of scopes could break existing code, right? Could that be an issue? I don't know how people use the Spark data source, I always used API key because it's easier than OIDC :D  I understand that we need empty scopes when audience is specified, but there could be a condition in the default value

emelvarol-cognite:
> Looks good.
> 
> However, removing the default value of scopes could break existing code, right? Could that be an issue? I don't know how people use the Spark data source, I always used API key because it's easier than OIDC :D I understand that we need empty scopes when audience is specified, but there could be a condition in the default value

I think if they don't upgrade the version without checking the CHANGELOG, it makes no problem :P I didn't want to do it that way since there may be some people with the need of providing both at the same time. I stopped keeping track of the possibilities and left the doors open :D What do you think?

src/test/scala/cognite/spark/v1/OAuth2Test.scala
emelvarol-cognite: how do we do that exactly :P it was me being a noob about scala test utilities hehe

exyi:
> I think if they don't upgrade the version without checking the CHANGELOG, it makes no problem :P I didn't want to do it that way since there may be some people with the need of providing both at the same time. I stopped keeping track of the possibilities and left the doors open :D What do you think?

Yea, ok

src/test/scala/cognite/spark/v1/OAuth2Test.scala
emelvarol-cognite: fixed, thanks!!

exyi: Cool! Don't be afraid to ask how to use these things :)

PR #517, opened 2021-09-15
add directory to File ResourceType

TharinduDG:
**CDF-13219** add `directory` attribute to `File` resource type


hakontro (approved):
LGTM :+1:

hakontro:
@TharinduDG nvm the codecov going down a bit, it's np in this case

exyi:
>@TharinduDG nvm the codecov going down a bit, it's np in this case

I think you could just add directory to some existing tests.

TharinduDG:
> I think you could just add directory to some existing tests :P

Correct if I'm wrong, I couldn't find any tests that uses attributes of `FilesInsertSchema`, `FilesReadSchema`, or `FilesUpsertSchema`.
Therefore I didn't make any changes to the tests.

PR #518, opened 2021-09-15
asset hierarchy: don't use n^2 algorithm for ordering nodes

exyi:


einarmo (approved):
I guess when this hits a couple million it might be a little expensive. Nice catch.

hakontro (approved):
LGTM!

PR #522, opened 2021-10-05
Fix filter pushdown for datapoints.value

exyi:
Also, changed error messages from sys.error to throwing a
CdfSparkIllegalArgumentException. The problem was that
Transformations displayed it as internal error instead of
showing used the erorr message.

### DIFF
--- a/src/main/scala/cognite/spark/v1/PushdownUtilities.scala
+++ b/src/main/scala/cognite/spark/v1/PushdownUtilities.scala
@@ -118,7 +118,7 @@ object PushdownUtilities {
     val timestampLimits = filters.flatMap(getTimestampLimit(_, colName))
 
     if (timestampLimits.exists(_.value.isBefore(Instant.ofEpochMilli(0)))) {
-      sys.error("timestamp limits must exceed 1970-01-01T00:00:00Z")
+      throw new CdfSparkIllegalArgumentException("timestamp limits must exceed 1970-01-01T00:00:00Z")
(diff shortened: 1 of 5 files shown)


src/main/scala/cognite/spark/v1/PushdownUtilities.scala
```diff
@@ -149,10 +149,10 @@ object PushdownUtilities {
 
   def getTimestampLimit(filter: Filter, colName: String): Seq[Limit] =
     filter match {
-      case LessThan(colName, value) => Seq(timeStampStringToMax(value, -1))
-      case LessThanOrEqual(colName, value) => Seq(timeStampStringToMax(value, 0))
-      case GreaterThan(colName, value) => Seq(timeStampStringToMin(value, 1))
-      case GreaterThanOrEqual(colName, value) => Seq(timeStampStringToMin(value, 0))
+      case LessThan(`colName`, value) => Seq(timeStampStringToMax(value, -1))
```
emelvarol-cognite, line 152: why "`" ?

exyi: Oh, that is a Scala trap: https://stackoverflow.com/a/7078077

Before, it declared a new variable instead of checking that it's equal to colName...

emelvarol-cognite:
i think this will pass the checks if you merge master

> great

i think this will pass the checks if you merge master

PR #523, opened 2021-10-07
Fix reading datapoints with too many partitions

exyi:
Now we limit the number of ranges to 10k, otherwise we'd have to do
paging when splitting and parallelizing this large number of ranges
will probably only hurt performance.

Also, I have changed the range size to be 10 pages instead of only 1.
Paging through 10 datapoints will take ~30 seconds even when
CDF is slow, so it won't hurt performance and it will allow us to proceed
with more even ranges and with less paralelism (so hopefully less rate-limiting).


wjoel:
From the description I don't understand what this is fixing... did we hit the exception? Or does it mean that "fewer ranges + paging" is better than "more ranges + no paging"? Seems unintuitive as the latter should have less latency and higher parallelism, but maybe that's harmful?

emelvarol-cognite:
and please bump the version, i think would be nice to release and use this

PR #532, opened 2021-11-02
Add support for filter pushdown on externalIdPrefix

exyi:


emelvarol-cognite (approved):
LGTM 👯

emelvarol-cognite:
version bump

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -77,7 +78,8 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
       name = m.get("name"),
       source = m.get("source"),
       dataSetIds = m.get("dataSetId").map(idsFromWrappedArray(_).map(CogniteInternalId(_))),
-      labels = m.get("labels").map(externalIdsToContainsAny).getOrElse(None)
+      labels = m.get("labels").map(externalIdsToContainsAny).getOrElse(None),
```
emelvarol-cognite, line 81: do we need `getOrElse(None)` here? I know it was already like that.

exyi: it's the same as flatten in this case - it unwraps one layer of options

exyi: Actually, I don't know why didn't we use flatMap, i'll change it

emelvarol-cognite: hehe probably it was me not using scala functions effectively at the time.

emelvarol-cognite:
also changelog

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
```diff
@@ -77,6 +77,8 @@ object PushdownUtilities {
         PushdownFilter("max" + colName.capitalize, value.toString)
       case LessThanOrEqual(colName, value) =>
         PushdownFilter("max" + colName.capitalize, value.toString)
+      case StringStartsWith(colName, value) =>
```
vuhainguyen, line 80: I think we should add ` | StringStartsWith(_, null)` into the condition at the line 67 as well?

PR #533, opened 2021-11-02
Pushdown columns selection to RAW API

hakontro:


emelvarol-cognite (approved):
LGTM version bump pls and changelog

src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -125,7 +125,10 @@ class RawTableRelation(
     val (minLastUpdatedTime, maxLastUpdatedTime) = filtersToTimestampLimits(filters, "lastUpdatedTime")
 
     val rdd =
-      readRows(config.limitPerPartition, None, RawRowFilter(minLastUpdatedTime, maxLastUpdatedTime))
+      readRows(
+        config.limitPerPartition,
+        None,
+        RawRowFilter(minLastUpdatedTime, maxLastUpdatedTime, Some(requiredColumns)))
```
exyi, line 131: You can exclude "key" and "lastUpdatedTime".

We should also request all columns if the `requiredColumns` array is empty, since rely on it on line 122

PR #535, opened 2021-11-05
Fix future datapoints

vuhainguyen:


build.sbt
```diff
@@ -14,7 +14,7 @@ val circeVersion: Option[(Long, Long)] => String = {
 val sttpVersion = "3.3.15"
 val Specs2Version = "4.2.0"
 val artifactory = "https://cognite.jfrog.io/cognite/"
-val cogniteSdkVersion = "1.5.14"
+val cogniteSdkVersion = "1.5.16-SNAPOSHOT"
```
vuhainguyen, line 17:
```suggestion
val cogniteSdkVersion = "1.5.16"
```
To be updated once new maven release is done

src/test/scala/cognite/spark/v1/DataPointsRelationTest.scala
```diff
@@ -966,6 +937,81 @@ class DataPointsRelationTest extends FlatSpec with Matchers with ParallelTestExe
     assert(error.getMessage.contains(s"The timeseries id=$valhallStringTimeSeriesId"))
   }
 
+  it should "read and write datapoints in the future" taggedAs WriteTest in {
+    val testUnit = s"future ${shortRandomString()}"
+    val tsName = s"future${shortRandomString()}"
```
emelvarol-cognite, line 942: should we delete the timeseries at the end of the test, to avoid having a lot of random timeseries in the tenant?

src/main/scala/cognite/spark/v1/NumericDataPointsRdd.scala
```diff
@@ -249,7 +249,10 @@ final case class NumericDataPointsRdd(
         queryDatapointsById(id, start, end.max(start.plusMillis(1)), 1)
           .map(datapoints => id -> datapoints.headOption)
       }.parSequence
-      latest <- client.dataPoints.getLatestDataPoints(ids, ignoreUnknownIds = true)
+      latest <- client.dataPoints.getLatestDataPoints(
+        ids,
+        ignoreUnknownIds = true,
+        end.toEpochMilli.toString)
```
emelvarol-cognite, line 255: how does this change help with this issue? could you explain 😅  is the default value now here and do we override it?

emelvarol-cognite (approved):
niceee! please bump the version

src/main/scala/cognite/spark/v1/NumericDataPointsRdd.scala
exyi: instead of using `"now"`, which is the API default, we will use our upper bound. If user supplied a filter `timestamp < X', then we'll use X. If not, we have some hardcoded date at 2100. I hope nobody is trying to predict valve flow at year 2100 :sweat_smile:

exyi (approved):
Cool

src/test/scala/cognite/spark/v1/DataPointsRelationTest.scala
vuhainguyen: right, done here https://github.com/cognitedata/cdp-spark-datasource/pull/535/commits/982f286fbeb7e39678f5e8b31f575b6da4a1a5cf

src/main/scala/cognite/spark/v1/NumericDataPointsRdd.scala
vuhainguyen: actually I encountered an error that timestamp can not be greater than the year 2050 when trying to insert some dataPoints to test, should be fine I think 😁

emelvarol-cognite: BAN  ‼️ 

image

emelvarol-cognite: Aha yes, I didn't see that fix, thanks!

wjoel: Yes, but the git commit history nor the source code will record this, so a comment in either the commit message or the code itself about why this change is done, would have been useful.

vuhainguyen: done ;) https://github.com/cognitedata/cdp-spark-datasource/pull/536/commits/e4eedd3adbaf73cc0a8dedd73ebd09f7473e76b3

PR #536, opened 2021-11-05
Fix compilation parse Instant as TimeStamp in Spark

vuhainguyen:


src/test/scala/cognite/spark/v1/DataPointsRelationTest.scala
```diff
@@ -989,6 +989,7 @@ class DataPointsRelationTest extends FlatSpec with Matchers with ParallelTestExe
 
     points
       .toDF("externalId", "timestamp", "value")
+      .withColumn("timestamp", to_timestamp(col("timestamp")))
```
emelvarol-cognite, line 992: what is this for?

vuhainguyen: The build on master failed due to missing Encoder for java.time.Instant https://github.com/cognitedata/cdp-spark-datasource/runs/4118432327?check_suite_focus=true
So I try to force the conversion of that column to timeStamp type

emelvarol-cognite: I see sure

emelvarol-cognite (approved):

PR #537, opened 2021-11-08
Use type sql.Timestamp instead of java.Instant for spark encoder in test

vuhainguyen:


emelvarol-cognite (approved):
maybe leave a comment explaining this

PR #541, opened 2021-11-11
select * issue fix

TharinduDG:


hakontro (changes requested):
Looks good so far! I'd like us to add some logic to not use the filter when the required columns are too many, but I'm not sure what the number should be.. but something like `requiredColumns.mkString(",").length > 200` should fail the query, as we'll run into the same issue as now. Would also like to see a test that reads and writes, that would have failed before this fix.

src/test/scala/cognite/spark/v1/RawTableRelationTest.scala
```diff
@@ -11,7 +11,7 @@ import org.apache.spark.sql.types._
 import org.scalatest.{FlatSpec, LoneElement, Matchers, ParallelTestExecution}
 
 class RawTableRelationTest
-  extends FlatSpec
+    extends FlatSpec
```
TharinduDG, line 14: These are auto formatting that got applied by force.

src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -122,12 +122,18 @@ class RawTableRelation(
   override def buildScan(): RDD[Row] = buildScan(Array.empty, Array.empty)
 
   override def buildScan(requiredColumns: Array[String], filters: Array[Filter]): RDD[Row] = {
-    val (minLastUpdatedTime, maxLastUpdatedTime) = filtersToTimestampLimits(filters, "lastUpdatedTime")
+    if (requiredColumns.mkString(",").length > 200) {
+      // Temp fix until this parameter is moved into the request body
+      throw new CdfSparkIllegalArgumentException("Fetching a large number of columns is not supported")
+    }
```
hakontro, line 128: This can actually just be part of the if-statement below: If they specify a lot of columns, and exceed length of 200, we will just fetch all the data like normally, and Spark will filter the columns afterwards.. a little slower than optimal, but better than throwing exception :)

src/test/scala/cognite/spark/v1/RawTableRelationTest.scala
```diff
@@ -457,8 +482,40 @@ class RawTableRelationTest
...
         writeClient.rawTables(database).deleteById(table)
       } catch {
-        case _: CdpApiException => ()// Ignore
+        case _: CdpApiException => () // Ignore
       }
     }
   }
+
+  // This is a test for a Temp fix applied
+  it should "not allow to fetch a large number of columns which has 200 characters when combined with ','" taggedAs WriteTest in {
```
hakontro, line 491: The test is good! I suggest rewriting it to somehow check that all rows are _fetched_.. it's ok to just check that the read actually works though, since we know it'll be hitting the right codepath if it reads the data instead of getting an exception from RAW.

Would actually be even better if you 
```
val columnsAsString = columns.mkString(",")
s"select $columnsAsString from rw.tbl"
```

hakontro (changes requested):

hakontro (approved):
Yes! :D Awesome!

PR #545, opened 2021-11-15
Remove temporary support for scala 2.13

vuhainguyen:


build.sbt
```diff
@@ -1,8 +1,8 @@
 import com.typesafe.sbt.packager.docker.Cmd
 
-val scala212 = "2.12.15"
+//val scala212 = "2.12.15"
```
hakontro, line 3:
```suggestion
val scala212 = "2.12.15"
```

build.sbt
```diff
@@ -1,8 +1,8 @@
...
 val scala213 = "2.13.6"
-val supportedScalaVersions = List(scala212, scala213)
+val supportedScalaVersions = List(scala213)
```
hakontro, line 5:
```suggestion
val supportedScalaVersions = List(scala212)
```

I think you're doing the opposite of what you're trying to do here :sweat_smile:

hakontro (changes requested):
Opposite day!

build.sbt
vuhainguyen: totally 🤦

build.sbt
vuhainguyen: https://github.com/cognitedata/cdp-spark-datasource/pull/545/commits/cf86d767f4086967f6e927e65f297364a6e3e0b4

exyi (approved):
It's implemented correctly now, but why do we need it? is scala 2.13 causing some problems?

vuhainguyen:
> It's implemented correctly now, but why do we need it? is scala 2.13 causing some problems?

yes [many tests failed on 2.13 due to this](https://github.com/cognitedata/cdp-spark-datasource/runs/4209080473?check_suite_focus=true)
https://docs.scala-lang.org/overviews/core/collections-migration-213.html

exyi:
> yes many tests failed on 2.13 due to this
https://docs.scala-lang.org/overviews/core/collections-migration-213.html


ah shit :| love languages with breaking changes in minor versions... :facepalm:  But if we can still use Spark 3.2 then it's fine, I guess

wjoel:
The `ArraySeq` vs `Seq` test failures don't look like they'd be very difficult to fix. Especially as many other tests are passing just fine. Haven't looked at the count differences (as in RelationshipsRelationTest), they might be more difficult, but could be due to similarly trivial reasons.

hakontro (approved):

PR #546, opened 2021-11-16
Fix performance problem with datapoint insertion

exyi:
Main problem was that when inserting into very many time series,
we converged into making 1 request per datapoint - because
each datapoint for one series ended up in a different partition.

This lead to a surge of parallel requests -> rate limiting -> more problems.
So now we also limit the write parallism in the same way as
we do with read parallism.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 1.4.44
+
+## Fixes
+
+* Fixed a bug where inserting into many timeseries created too many small requests
+
 # 1.4.45
 
 ## Enhancements
--- a/src/main/scala/cognite/spark/v1/DataPointsRelation.scala
+++ b/src/main/scala/cognite/spark/v1/DataPointsRelation.scala
@@ -45,15 +45,19 @@ abstract class DataPointsRelationV1[A](config: RelationConfig, shortName: String
     client.dataPoints.deleteRanges(deleteRanges)
   }
 
-  override def insert(data: DataFrame, overwrite: Boolean): Unit =
-    data.foreachPartition((rows: Iterator[Row]) => {
-      val batches = rows.grouped(Constants.CreateDataPointsLimit).toVector
-      batches
-        .parTraverse_(insertSeqOfRows)
-        .unsafeRunSync()
-    })
+  override def insert(data: DataFrame, overwrite: Boolean): Unit = {
+    val partitions = Math.max(1, config.partitions / config.parallelismPerPartition)
+    val partitionCols =
+      data.schema.fieldNames.filter(f => f.equalsIgnoreCase("id") || f.equalsIgnoreCase("externalId"))
+    import org.apache.spark.sql.functions.col
+    data
+      .repartition(partitions, partitionCols.map(col): _*)
+      .foreachPartition((rows: Iterator[Row]) => {
+        insertRowIterator(rows).unsafeRunSync()
+      })
+  }
 
-  def insertSeqOfRows(rows: Seq[Row]): IO[Unit]
+  def insertRowIterator(rows: Iterator[Row]): IO[Unit]
 
   override def buildScan(): RDD[Row] = buildScan(Array.empty, Array.empty)
 
--- a/src/main/scala/cognite/spark/v1/DefaultSource.scala
+++ b/src/main/scala/cognite/spark/v1/DefaultSource.scala
@@ -165,6 +165,26 @@ class DefaultSource
       val relation = new AssetHierarchyBuilder(config)(sqlContext)
       relation.buildFromDf(data)
       relation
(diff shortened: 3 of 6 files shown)


src/main/scala/cognite/spark/v1/DataPointsRelation.scala
```diff
@@ -45,15 +45,16 @@ abstract class DataPointsRelationV1[A](config: RelationConfig, shortName: String
...
   }
 
-  override def insert(data: DataFrame, overwrite: Boolean): Unit =
-    data.foreachPartition((rows: Iterator[Row]) => {
-      val batches = rows.grouped(Constants.CreateDataPointsLimit).toVector
-      batches
-        .parTraverse_(insertSeqOfRows)
-        .unsafeRunSync()
+  override def insert(data: DataFrame, overwrite: Boolean): Unit = {
+    val partitions = config.partitions / config.parallelismPerPartition
```
hakontro, line 49: Should this be rounded up instead, to avoid possible 0?

hakontro (commented):
The approach looks good, but there's some compilation issues here so I'll review again when those are fixed.

exyi:
Oh sorry for commiting it half baked :sweat_smile: , it should be fixed now

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -165,6 +165,26 @@ class DefaultSource
...
       relation
+    } else if (resourceType == "datapoints" || resourceType == "stringdatapoints") {
+      val relation = resourceType match {
+        case "datapoints" =>
+          new NumericDataPointsRelationV1(config)(sqlContext)
+        case "stringdatapoints" =>
+          new StringDataPointsRelationV1(config)(sqlContext)
+      }
+      if (config.onConflict == OnConflict.Delete) {
+        // Datapoints support 100_000 per request when inserting, but only 10_000 when deleting
```
hakontro, line 176: Surprising.. wonder why!

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -165,6 +165,26 @@ class DefaultSource
...
+        val batchSize = config.batchSize.getOrElse(Constants.DefaultDataPointsLimit)
+        data.foreachPartition((rows: Iterator[Row]) => {
+          import CdpConnector._
```
hakontro, line 179: Some implicit for the partTraverse_ , right? Perhaps we can import the specific one.

src/main/scala/cognite/spark/v1/NumericDataPointsRelation.scala
```diff
@@ -147,18 +149,23 @@ class NumericDataPointsRelationV1(config: RelationConfig)(sqlContext: SQLContext
...
+    dataPoints.chunks
+      .flatMap(c => Stream.emits(c.toVector.groupBy(_.getCogniteId).toVector))
+      .parEvalMapUnordered(config.parallelismPerPartition) {
+        case (id, dataPoints) =>
+          client.dataPoints
+            .insert(id, dataPoints.map(dp => SdkDataPoint(dp.timestamp, dp.value)))
+            .flatTap(_ => incMetrics(itemsCreated, dataPoints.size))
+      }
+      .compile
+      .drain
```
hakontro, line 168: Identical in here and StringDataPointsRelation, I think? Can we deduplicate?

hakontro (approved):
LGTM, just some suggestions.

vuhainguyen:
would be great if you can add a test for these changes hehe

src/main/scala/cognite/spark/v1/DefaultSource.scala
exyi: good idea

src/main/scala/cognite/spark/v1/DefaultSource.scala
exyi: I guess because it's deleted by ranges, so it allows at most 10k ranges... not sure however, I guess we just have to accept the limitation

src/main/scala/cognite/spark/v1/NumericDataPointsRelation.scala
exyi: Not sure how, it's working with different types (SdkDataPoint, InsertDataPointsItem). I mean it's possible with type arguments and lambda function, but it would not be more readable, IMHO.

exyi:
> would be great if you can add a test for these changes hehe

problem is I don't know how to test this, but all code paths are already covered - so we know it works correctly, just not that sure if it's really faster.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 1.4.44
```
hakontro, line 1: Oops!

PR #547, opened 2021-11-21
Changelog: Fix my stupid mistake

exyi:
Apparently, incrementing a number is not as easy as I'd think :upside_down_face:


hakontro (approved):
Yes

PR #549, opened 2021-11-26
Use fs2.Pull for datapoints

exyi:
Surprisingly, this scala lalalalala improves performance quite
a bit. What took 102 seconds now takes just 50 (and it's
now limited by my network instead of the CPU).

The problem was that we were appending quite big Seq together, which was pretty slow.
fs2 does lot of things to avoid this, so that's why it's faster.


hakontro (approved):
Doubing the speed is always a good idea! :racehorse: LGTM!

PR #551, opened 2021-12-07
Add RAW limit based on sample size

hakontro:


src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -79,6 +80,13 @@ class RawTableRelation(
     }
   }
 
+  def calculateDynamicLimit(): Int = {
+    val rows = client.rawRows(database, table).list(Some(1000)).compile.toList.unsafeRunSync()
+    val sizeInBytes = SizeEstimator.estimate(rows)
+    val dynamicLimit = 1000000 / sizeInBytes // Avoid reading requests that are larger than 1MB
```
exyi, line 86: Why is that a problem? Does it cause the RAW service to struggle or does sttp struggle with that?

exyi: We should add a lower bound, if the size estimator returns more than `1000000`, we'd set the limit to zero.

And also, this estimates the size of 1000 rows together (AFAIK), so we should divide the sizeInBytes by 1000. That will give us average  I think it would be nicer to calculate the size of each row and then take a median (or some other percentile), that will be less succeptible to getting scared of one giant row and then reading them with limit=1

hakontro: RAW on westeurope-1 is struggling, but they're going to add this type of limit to the service instead (so we won't add this PR).

hakontro:
This limit will be implemented in RAW instead.

PR #553, opened 2021-12-08
Add basic response to rate limiting backpressure from CDF

exyi:
New sttp backend simply stops all requests
relation when 429 or 503 is encountered.
Operation is resumed without any limits after
a specified delay (I set it to 800ms for now)

Since each relation and executor has it's own instance of the backend,
not all requests are going to be stopped.

There is probably many smarter things we could do, but I'd prefer to have at least something for now.
This should be good enough to prevent the vicious cycle we have now - service is struggling -> sends 503/429 -> Jetfire retries all of them -> even more requests for the service.


hakontro:
This simply adds an 800ms delay to requests that get 429/503, right? I mean they're still retried, but after a flat 800ms delay + the exponentialbackoff timer?

exyi:
It also removes the permit for 800ms, so new request will have to wait too.

hakontro (approved):
LGTM, please version bump and release

PR #555, opened 2021-12-09
Don't crash on NULL in labels

exyi:
First I thought about raing some nicer error, but it's easier
to just ignore nulls in the array. It's also more convenient for
some use cases - users sometimes have a label column
in their raw table that is nullable  and then want to do `array(label)` in SQL


hakontro (approved):
LGTM

PR #559, opened 2021-12-21
Remove jackson mapping from RawTableRelation

exyi:
I replaced it with mapping directly into Circe's Json model.
Although a bit repetitive, it's actually less code than the magic
jackson configuration and it won't cause the problems
with conflicting dependencies.

I'd also bet this is faster than serializing the Row to Json String
and then parsing it again, although I did not measure


src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -40,17 +40,7 @@ class RawTableRelation(
 
   @transient lazy private val batchSize = config.batchSize.getOrElse(Constants.DefaultRawBatchSize)
 
-  @transient lazy val mapper: ObjectMapper = {
```
vuhainguyen, line 43: You can remove these useless import at the beginning of the file
```
import com.fasterxml.jackson.databind.{DeserializationFeature, ObjectMapper}
import com.fasterxml.jackson.module.scala.DefaultScalaModule
```

src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -220,19 +210,64 @@ object RawTableRelation {
...
-          io.circe.parser
-            .decode[Map[String, Json]](
-              mapper.writeValueAsString(row.getValuesMap[Any](nonKeyColumnNames)))
-            .right
-            .get
+          nonKeyColumnNames.map(f => f -> anyToRawJson(row.getAs[Any](f))).toMap
       ))
 
+  // scalastyle:off cyclomatic.complexity
+  /** Maps the object from Spark into circe Json object. This is the types we need to cover:
```
vuhainguyen, line 223: Where did you get the list below from? 😅

vuhainguyen (changes requested):

src/main/scala/cognite/spark/v1/RawTableRelation.scala
exyi: It's copy paste from documentation of `Row.get(...)` function

vuhainguyen (approved):

PR #567, opened 2022-01-13
Only shade `cats-kernel`

TharinduDG:
- only shade `cats-kernel` since others conflict with `jetfire-backend`


hakontro:
Good idea  - if this is for testing with jetfire-backend I think you'll be able to do that testing with assembly or +publishLocal, so you don't have to wait for a new Spark version online (and so we don't release a debug version that people might end up using, this repository is open source after all).

PR #569, opened 2022-01-17
shade typelevel.jawn to fix classNotFound issue in jetfire

TharinduDG:
- shade jawn with full package to fix classNotFound issue in jetfire test


hakontro (approved):
LGTM!

PR #570, opened 2022-01-19
Final changes cdp-spark-datasource fatJar

TharinduDG:
- remove `SNAPSHOT`
- correct naming mistake by renaming`cdp` to `cdf`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -137,7 +137,7 @@ lazy val library = (project in file("."))
       case _ => MergeStrategy.first
     },
     assemblyShadeRules := {
-      val shadePackage = "cognite.spark.v1.shaded"
+      val shadePackage = "cognite.shaded"
       Seq(
         ShadeRule.rename("cats.**" -> s"$shadePackage.cats.@1").inAll,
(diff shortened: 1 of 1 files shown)


wjoel:
I'd also do `val shadePackage = "cognite.spark.v1.shaded"` => `val shadePackage = "cognite.shaded"` and corresponding change in Jetfire: it doesn't hurt, and would let us explore https://github.com/cognitedata/jetfire-backend/pull/1280#issuecomment-1013032637 in a more natural way.

build.sbt
```diff
@@ -202,7 +202,7 @@ lazy val cdfdump = (project in file("cdf_dump"))
 
 lazy val fatJar = project.settings(
   commonSettings,
-  name := "cdp-spark-datasource-fat",
+  name := "cdf-spark-datasource-fat",
```
wjoel, line 205: If we don't publish a non-fat version, we should consider removing the `-fat` suffix, but that can be done in another PR.

wjoel (approved):
:crossed_fingers:

PR #574, opened 2022-01-26
Limit parallelism on write according to settings

exyi:
The only thing limiting number of write requests when
the dataset was large was the RateLimitingBackend's.
However, that's not very strict, and more importantly
scales with the number of Spark partitions. So when the
user had many partitions, we would allow basically
infinitely many parallel requests.

This PR limits the write parallelism in the same way as
we limit read paralelism - capped by config.partitions
(meaning CDF partitions). Also we do a repartition if the
number of Spark partitions is large, compared to CDF
insertion, repartition is fairly cheap and it avoids Jetfire
getting stuck on inserting one large partition, if the
distribution is uneven.


hakontro (approved):
Woho :partying_face:

PR #577, opened 2022-01-31
Handle empty string for Boolean type in RawJsonConverter

vuhainguyen:
CDF-14684


src/test/scala/cognite/spark/v1/RawTableRelationTest.scala
```diff
@@ -82,6 +98,12 @@ class RawTableRelationTest
     RawRow("k2", Map("num" -> Json.fromDouble(12.3).get))
   )
 
+  private val dataWithEmptyStringInBooleanField = Seq(
+    RawRow("k1", Map("bool" -> Json.fromString(""))),
```
exyi, line 102: You can also test if `Json.fromString("true")` works.

exyi (approved):
Booleans also :facepalm: Thanks for fixing it

src/test/scala/cognite/spark/v1/RawTableRelationTest.scala
vuhainguyen: using fromString("true") it matched the StringType in the pattern matching instead of BooleanType, I used `Json.fromBoolean(java.lang.Boolean.parseBoolean("true"))` instead, but technically it creates the object Json.True  so I think this is not really what you want to test 😅

PR #579, opened 2022-02-02
Chunk sequences when creating to avoid limit 10000 columns

vuhainguyen:
CDF-14684


src/main/scala/cognite/spark/v1/SequencesRelation.scala
```diff
@@ -48,9 +48,27 @@ class SequencesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+      client.sequences
+        .create(sequences)
+        .flatTap(_ => incMetrics(itemsCreated, sequences.size)) *> IO.unit
+    } else {
+      // Sequence API does not support more than 200 columns per sequence,
+      // and the maximum total columns per request is 10000
+      // so grouped by 50 ensure that each chunk can not have more than 10000 columns in total,
+      // otherwise we already reject because of the first condition
+      sequences
+        .grouped(50)
```
wjoel, line 62: Seems a bit pessimistic. Group them so that the number of columns in each group doesn't exceed the limit, something like this?
```scala
    // Sequence API does not support more than 200 columns per sequence,
    // and the maximum total columns per request is 10000
    // so group them such that each group has no more than Constants.DefaultSequencesTotalColumnsLimit columns in total,
    // otherwise we already reject because of the first condition
    val (_, groupedSequences) = sequences
      .foldLeft((0, List(Vector.empty[SequenceCreate]))) { case ((currentGroupColumnsCount, groups @ currentGroup :: otherGroups), sequence) =>
        val columnsInSequence = sequence.columns.size
        if (currentGroupColumnsCount + columnsInSequence > Constants.DefaultSequencesTotalColumnsLimit) {
          // create a new group
          (currentGroupColumnsCount, Vector(sequence) :: groups)
        } else {
          // add to current group
          // Would prefer +: instead of :+, but +: is giving me a syntax error for some reason. Doesn't really matter though.
          (currentGroupColumnsCount + columnsInSequence, (currentGroup :+ sequence) :: otherGroups)
        }
      }
    groupedSequences
      .map { sequencesToCreate =>
        client.sequences
          .create(sequencesToCreate)
          .flatMap(_ => incMetrics(itemsCreated, sequencesToCreate.size))
      }
      .sequence_
```
(not tested, but you get the idea)

src/main/scala/cognite/spark/v1/SequencesRelation.scala
```diff
@@ -48,9 +48,27 @@ class SequencesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
+    val nbColumns = sequences.flatMap(_.columns.toList).size
+    if (nbColumns < Constants.DefaultSequencesTotalColumnsLimit) {
+      client.sequences
+        .create(sequences)
+        .flatTap(_ => incMetrics(itemsCreated, sequences.size)) *> IO.unit
+    } else {
+      // Sequence API does not support more than 200 columns per sequence,
+      // and the maximum total columns per request is 10000
+      // so grouped by 50 ensure that each chunk can not have more than 10000 columns in total,
+      // otherwise we already reject because of the first condition
```
wjoel, line 60: Hm, I don't see where we do that rejection.

src/main/scala/cognite/spark/v1/SequencesRelation.scala
```diff
@@ -48,9 +48,27 @@ class SequencesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
       )
     }
-    client.sequences
-      .create(sequences)
-      .flatTap(_ => incMetrics(itemsCreated, sequences.size)) *> IO.unit
+    val nbColumns = sequences.flatMap(_.columns.toList).size
+    if (nbColumns < Constants.DefaultSequencesTotalColumnsLimit) {
+      client.sequences
+        .create(sequences)
+        .flatTap(_ => incMetrics(itemsCreated, sequences.size)) *> IO.unit
```
wjoel, line 55: No need to special case this with the suggestion below, I think.

vuhainguyen: yes 😁

src/main/scala/cognite/spark/v1/SequencesRelation.scala
vuhainguyen: The limit of 200 columns `[is specified in the API ](https://docs.cognite.com/api/v1/#operation/createSequence)`
But the limit about 10000 columns max per request is not, yesterday I talked with time-lord team and they just create a ticket for this https://cognitedata.atlassian.net/browse/CDF-14714 :)

src/main/scala/cognite/spark/v1/SequencesRelation.scala
vuhainguyen: Thanks a lot, I was thinking about using foldLeft but too lazy to come up with the final solution and using the brute grouped by 50 instead 😆. 
I changed 2 things: when creating the new group, we need to reset the currentGroupColumnsCount, so return 
 `columnsInSequence` instead `currentGroupColumnsCount` in the if clause.
You prefer `+:` than `:+` [for performance](https://docs.scala-lang.org/overviews/collections/performance-characteristics.html)? To use +: , the syntax is sequence `+:` currentGroup because `:`[ is where the Collection goes](https://stackoverflow.com/questions/37190619/what-does-mean-in-scala) (I made the update ;) )

wjoel: Great. Ah, that's why, the order changes! Makes sense. There's no performance difference, but it's nice to keep the same order.

wjoel (approved):
LGTM! ;)

PR #581, opened 2022-02-03
Raw reads: reduce allocations slightly

exyi:
Most importantly, don't remember all keys
in the hashset. It is only needed in resources which may
load the same item multiple times, but it never happens
in RAW, since there are essentially no filters.

### DIFF
--- a/src/main/scala/cognite/spark/v1/SdkV1Rdd.scala
+++ b/src/main/scala/cognite/spark/v1/SdkV1Rdd.scala
@@ -16,7 +16,8 @@ final case class SdkV1Rdd[A, I](
     config: RelationConfig,
     toRow: (A, Option[Int]) => Row,
     uniqueId: A => I,
-    getStreams: (GenericClient[IO], Option[Int], Int) => Seq[Stream[IO, A]])
+    getStreams: (GenericClient[IO], Option[Int], Int) => Seq[Stream[IO, A]],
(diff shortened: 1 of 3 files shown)


src/main/scala/cognite/spark/v1/SdkV1Rdd.scala
```diff
@@ -53,18 +54,21 @@ final case class SdkV1Rdd[A, I](
...
     val currentStreamsAsSingleStream = groupedStreams(split.index).reduce(_.merge(_))
 
-    val processedIds = new ConcurrentHashMap[I, Unit]
-    val processChunk = (chunk: Chunk[A]) => {
-      chunk.filter { i =>
-        // putIfAbsent returns null if the key did not exist, in which ase we
-        // should keep (and process) the item.
-        Option(processedIds.putIfAbsent(uniqueId(i), ())).isEmpty
-      }
+    val processChunk = if (deduplicateRows) {
```
wjoel, line 57: I think it would be good if we could try only this change first, instead of changing several things at once. Since ids for RAW are larger than for other resource types (being strings and all), it seems reasonable that this might be the issue.

exyi: Ok, I'll separate it. It just took me few attempt to find it and all included changes seem to help

PR #582, opened 2022-02-04
raw read: Don't deduplicate rows

exyi:
The HashMap consumes quite a bit of memory and it is
only needed in resources which may
load the same item multiple times, but it never happens
in RAW, since there are essentially no filters.

@wjoel : this is the part of #581 you requested as separate


wjoel (approved):
Very good, let's see if this is it.

vuhainguyen:
Nice, now even the broken test is fixed 😁  👍

PR #583, opened 2022-02-04
Feat/support sequence row

vuhainguyen:


exyi (approved):

exyi (changes requested):
wait, how is it supposed to work actually? This schema does not look right: https://github.com/cognitedata/jetfire-backend/pull/1322/files#diff-e6f3961e5dfdcf1e02e1d387f674f1fe6302542d962a519321a2a56de70516edR214

The schema is specific for each sequence, the `fromRow` loads the Spark Row fields into the sequence columns. It should work more like RAW than say datapoints

vuhainguyen:
as discussed I just closed this PR because we declare the schema [directly in jetfire](https://github.com/cognitedata/jetfire-backend/pull/1322/files#diff-467102ea0140e9241a1c70994af0428d3f2ba0107fba1a22e2ff3d4871123140R132)

PR #584, opened 2022-02-08
Migrate to Scala 2.13, improve conversion from row

exyi:
Adds support for Scala 2.13, but Scala 2.12 is still supported too.

They changed which collection it uses by default, so I needed to adjust
our fromRow macro as well. While at it,
the error message from using string in assetIds should be significantly improved


build.sbt
```diff
@@ -24,6 +24,7 @@ lazy val commonSettings = Seq(
   organizationHomepage := Some(url("https://cognite.com")),
   version := "1.4.63",
   crossScalaVersions := supportedScalaVersions,
+  scalaVersion := scala213, // default to Scala 2.13
```
wjoel, line 27: It's better to default to 2.12, as the IDE will pick up this setting, and code that is valid in 2.13 may not work on 2.12. It's better to not get recommendations from the IDE that may depend on 2.13.

macro/src/main/scala/SparkSchemaHelper.scala
```diff
@@ -51,16 +51,14 @@ class SparkSchemaHelperImpl(val c: Context) {
   def fromRow[T: c.WeakTypeTag](row: c.Expr[Row]): c.Expr[T] = {
     import c.universe._
     val optionType = typeOf[Option[_]]
-    val optionSeq = typeOf[Seq[Option[_]]]
+    val optionSeq = typeOf[Iterable[Option[_]]]
```
wjoel, line 54: I think the fix is usually to import `immutable.Seq` explicitly? https://docs.scala-lang.org/overviews/core/collections-migration-213.html

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -160,7 +160,14 @@ final case class AssetsReadSchema(
     createdTime: Instant = Instant.ofEpochMilli(0),
     lastUpdatedTime: Instant = Instant.ofEpochMilli(0),
     rootId: Option[Long] = Some(0),
-    aggregates: Option[Map[String, Long]] = None,
+    aggregates: Option[AssetsAggregatesSchema] = None,
```
wjoel, line 163: Why does this need to change? Or maybe it doesn't, but you're doing more than just adding support for Scala 2.13?

In that case, could we have a smaller PR with only the changes necessary for Scala 2.13, please?

build.sbt
exyi: Ok, good point. I defaulted to 2.13, because tests were mostly broken on that version, but you are right. btw, do you know when we can drop 2.12? Does anyone except Jetfire depend on that?

macro/src/main/scala/SparkSchemaHelper.scala
```diff
@@ -131,63 +116,84 @@ class SparkSchemaHelperImpl(val c: Context) {
...
+          } else if (t <:< mapString) {
             q"""($value match {
-             case x: scala.collection.immutable.Map[Any @unchecked, Any @unchecked] => x: scala.collection.immutable.Map[Any,Any]
+             case x: scala.collection.immutable.Map[Any @unchecked, Any @unchecked] =>
+               cognite.spark.v1.SparkSchemaHelperRuntime.checkMetadataMap(x, $row)
              case _ => $throwError
            })"""
-          } else {
+          } else if (t <:< seqAny) {
+            cats.data.NonEmptyList
```
wjoel, line 138: ? :)

src/main/scala/cognite/spark/v1/AssetsRelation.scala
wjoel: The changes to SparkSchemaHelper are quite large and invasive, maybe they can be smaller if `immutable.Seq` instead of `Iterable` helps. I haven't really tried to understand all those changes, or why they would be necessary just for 2.13.

exyi: I don't think that will help and Iterable makes it less picky which kind of `Seq` are we using (I think that it would mostly remove just the `.toVector` call)

macro/src/main/scala/SparkSchemaHelper.scala
```diff
@@ -131,63 +116,83 @@ class SparkSchemaHelperImpl(val c: Context) {
...
         q"$resExpr.asInstanceOf[${param.typeSignature}]"
       })
       q"new ${structType}(..$params)"
     }
-    c.Expr[T](fromRowRecurse(weakTypeOf[T], row))
+    val x = c.Expr[T](fromRowRecurse(weakTypeOf[T], row))
+    //println()
+    //println("---------------------------------------------------------------------------------------------------------")
+    //println(x)
+    x
```
wjoel, line 195: Remove those.

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
```diff
@@ -105,23 +105,30 @@ object PushdownUtilities {
   // PrunedFilteredScan, hence it's ok that our pushdown filter reads some data that should ideally
   // be filtered out
   // scalastyle:off
-  def getFilter(filter: Filter): PushdownExpression =
+  def getFilter(filter: Filter): PushdownExpression = {
+    def toStr(v: Any): String =
+      v match {
+        // let's hope that null-byte is invalid anywhere else
+        case i: Iterable[Any] => i.mkString("\u0000")
```
wjoel, line 112: Could do with a better explanation and perhaps an example. The null should be made into a `val` and reused in `idsFromWrappedArray` so the connection is made more clear.

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
```diff
@@ -154,15 +162,11 @@ object PushdownUtilities {
...
 
   def externalIdsSeqFromWrappedArray(wrappedArray: String): Seq[String] =
-    // We get "WrappedArray(ext1, ext2)" here, so we can remove irrelevant parts of the string
-    // and extract ext1 and ext2 from it -> return Seq('ext1', 'ext2')
-    wrappedArray.length match {
-      case 14 => Seq() // "WrappedArray()".length
-      case _ => wrappedArray.slice(13, wrappedArray.length - 1).split(',')
-    }
+    // we get a string separated by \0
+    wrappedArray.split('\u0000')
```
wjoel, line 166: Does that mean it's not a `WrappedArray` anymore, since we don't bother with removing the irrelevant parts of the string anymore? If so, should the name be changed?

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
```diff
@@ -154,15 +162,11 @@ object PushdownUtilities {
...
 
   def idsFromWrappedArray(wrappedArray: String): Seq[Long] =
-    wrappedArray.split("\\D+").filter(_.nonEmpty).map(_.toLong)
+    wrappedArray.split('\u0000').map(_.toLong)
```
wjoel, line 169: I don't know why we needed `.filter(_.nonEmpty)` before, but that also means I don't know why we can remove it now. ;)

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
```diff
@@ -229,14 +233,14 @@ object PushdownUtilities {
   def cogniteExternalIdSeqToStringSeq(
       cogniteExternalIds: Option[Seq[CogniteExternalId]]): Option[Seq[String]] =
     cogniteExternalIds match {
-      case Some(Seq()) => None
-      case _ => cogniteExternalIds.map(l => l.map(_.externalId))
+      case Some(x) if x.isEmpty => None
+      case _ => cogniteExternalIds.map(l => l.map(_.externalId).toVector)
```
wjoel, line 237: Do we need this? If yes, shouldn't it also be added to line 244?

wjoel (commented):
Very cool, looks pretty good now.

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
exyi: Ok, I'll add it back, but I also have no idea why would it be needed.

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
wjoel: Cool, but then what will this look like when running on 2.12? :thinking:

wjoel (approved):
Great work, lgtm as long as it also passes when running tests on 2.12 (see comment).

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
exyi: For sure, this works on 2.12. It used to be WrappedArray, because we just called `.toString` on anything that was passed to the pushdown filter logic. Now I explicitly use the null-byte separator, so it behaves in the same way on both Scala versions.

exyi:
The CI runs tests on 2.12 and on 2.13 when it's on the master branch (probably to speed things up in PRs). So it definitely passes on 2.12

PR #585, opened 2022-02-09
Add support id for SequenceRows

vuhainguyen:
CDF-14562


src/main/scala/cognite/spark/v1/SequenceRowsRelation.scala
```diff
@@ -164,9 +166,10 @@ class SequenceRowsRelation(config: RelationConfig, sequenceId: CogniteId)(val sq
         case _ => throw SparkSchemaHelperRuntime.badRowError(row, "rowNumber", "Long", "")
       }
 
-      val externalId = row.get(externalIdIndex) match {
-        case x: String => x
-        case _ => throw SparkSchemaHelperRuntime.badRowError(row, "externalId", "String", "")
+      val externalId = sequenceInfo.externalId match {
```
exyi, line 169: This is now taking the externalId from `sequenceInfo` instead of the current Spark Row. The `sequenceInfo` is global for all rows and we should basically only use it for read and schema inferrence

exyi (changes requested):

src/main/scala/cognite/spark/v1/SequenceRowsRelation.scala
```diff
@@ -164,19 +168,36 @@ class SequenceRowsRelation(config: RelationConfig, sequenceId: CogniteId)(val sq
...
+      val maybeId = Try(row.get(idIndex)).toOption match {
+        case Some(x: Long) => Some(x)
+        case Some(x: Int) => Some(x.toLong)
+        case _ => None
+      }
+
+      val externalId = (maybeExternalId, maybeId) match {
+        case (Some(externalId), _) => externalId
+        case (None, Some(id)) =>
+          client.sequences.retrieveById(id).unsafeRunSync().externalId match {
```
exyi, line 185: this would fetch the sequence metadata over and over again for each row, we should use the CogniteId union for the two possible cases.

vuhainguyen: RIght, I update the case class and use CogniteId instead ;)

src/main/scala/cognite/spark/v1/SequenceRowsRelation.scala
```diff
@@ -157,28 +161,39 @@ class SequenceRowsRelation(config: RelationConfig, sequenceId: CogniteId)(val sq
...
       val rowNumber = row.get(rowNumberIndex) match {
         case x: Long => x
         case x: Int => x: Long
         case _ => throw SparkSchemaHelperRuntime.badRowError(row, "rowNumber", "Long", "")
       }
 
-      val externalId = row.get(externalIdIndex) match {
-        case x: String => x
-        case _ => throw SparkSchemaHelperRuntime.badRowError(row, "externalId", "String", "")
+      val maybeExternalId = Try(row.get(externalIdIndex)).toOption match {
```
exyi, line 171: `row.get` will always succeed if the `externalIdIndex` is >= 0, I'd much prefer an if check on the index instead of throwing and catching an exception

src/main/scala/cognite/spark/v1/SequenceRowsRelation.scala
```diff
@@ -202,15 +217,24 @@ class SequenceRowsRelation(config: RelationConfig, sequenceId: CogniteId)(val sq
...
-              .flatTap(_ => incMetrics(itemsCreated, rows.length))
+            cogniteExternalId match {
+              case externalId: CogniteExternalId =>
+                client.sequenceRows
+                  .insertByExternalId(externalId.externalId, columns, rows.map(_.sequenceRow))
+                  .flatTap(_ => incMetrics(itemsCreated, rows.length))
+              case internalId: CogniteInternalId =>
+                client.sequenceRows
+                  .insertById(internalId.id, columns, rows.map(_.sequenceRow))
+                  .flatTap(_ => incMetrics(itemsCreated, rows.length))
```
exyi, line 235: you don't have to do that now, but it would be nice to migrate these insert functions to CogniteId too ;)

exyi (approved):

PR #589, opened 2022-02-18
Deduplicate when reading from row to reduce multiple same update request

vuhainguyen:
Trying to fix 429 error returned by assets api when we send multiple same update request.


exyi (approved):
Let's try it

PR #590, opened 2022-02-18
Get RAW /cursors just once on the driver

exyi:
We used to do that in the getStreams function, but that
is invoked for each Spark partition so we might get a
different set of cursors (in theory). It's also unnecesary
to do 20x the same request to the RAW api


wjoel (approved):
LGTM, I wonder if we should do this for other resource types as well? Remember to release new version and update changelog if you want to get this out quickly.

wjoel:
Although, maybe wait with deployments until Monday, both to not deploy too late on a Friday and to give ce3 alone some more time to prove itself.

exyi:
> LGTM, I wonder if we should do this for other resource types as well

Other resources use `partition=12/200`, not magic cursors like RAW, so I this it should be fine.

PR #593, opened 2022-02-25
Remove RAW deduplication table again

exyi:
We tried to add it back to fix the 429 Conflict issue, but
it didn't help, so I'm adding this back.
With the table added back, I'm unable to dump some
RAW tables on my laptop.


vuhainguyen (approved):
Please bump log and add change as well 😁

PR #596, opened 2022-03-07
Add pushdown filters on data model instances

emelvarol-cognite:


src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -89,6 +89,52 @@ class DataModelInstanceRelation(config: RelationConfig, modelExternalId: String)
...
+        if (Seq(
+            "text[]",
+            "boolean[]",
+            "numeric[]",
+            "float32[]",
+            "float64[]",
+            "int32[]",
+            "int[]",
+            "int64[]",
+            "bigint[]") contains propertyTypes(attribute)._1) {
```
exyi, line 108: maybe

```suggestion
        if (propertyTypes(attribute)._1.endsWith("[]")) {
```

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -89,6 +89,52 @@ class DataModelInstanceRelation(config: RelationConfig, modelExternalId: String)
...
+        val instancef2 = getInstanceFilter(f2)
+        Seq(DMIAndFilter(instancef1 ++ instancef2))
+      case Or(f1, f2) =>
+        val instancef1 = getInstanceFilter(f1)
+        val instancef2 = getInstanceFilter(f2)
+        Seq(DMIOrFilter(instancef1 ++ instancef2))
+      case IsNotNull(attribute) =>
+        Seq(DMIExistsFilter(Seq(modelExternalId, attribute)))
+      case Not(f) =>
+        Seq(DMINotFilter(getInstanceFilter(f).head))
```
exyi, line 133: I think that you should probably check the case when getInstanceFilter returns an empty collection (it may be an unsupported filter)

```suggestion
        getInstanceFilter(f).map(DMINotFilter(_))
```

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -89,6 +89,52 @@ class DataModelInstanceRelation(config: RelationConfig, modelExternalId: String)
     )
   }
 
+  // scalastyle:off cyclomatic.complexity
+  def getInstanceFilter(sparkFilter: Filter): Seq[DataModelInstanceFilter] =
```
exyi, line 93: Seems that we only ever return 0 or 1 element collection, so I think it would be better to use Option. If you need to return a collection, it would be better to use `DMIOrFilter` or `DMIAndFilter`

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -99,10 +145,16 @@ class DataModelInstanceRelation(config: RelationConfig, modelExternalId: String)
       requiredColumns
     }
 
-    // TODO: Implement pushdown filters
+    val filter = if (filters.isEmpty) {
+      None
+    } else {
+      val andFilters = filters.toVector.flatMap(getInstanceFilter)
+      if (andFilters.isEmpty) None else Some(DMIAndFilter(andFilters))
+    }
```
exyi, line 153:
```suggestion
    val filter = {
      val andFilters = filters.toVector.flatMap(getInstanceFilter)
      if (andFilters.isEmpty) None else Some(DMIAndFilter(andFilters))
    }
```

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -259,45 +333,46 @@ object DataModelInstanceRelation {
         case null if !nullable =>
           throw new CdfSparkException(propertyNotNullableMessage(propertyType)) // scalastyle:off null
         case null => Json.Null // scalastyle:off null
-        case x: Iterable[String] @unchecked if x.nonEmpty && x.head.isInstanceOf[String] =>
+        case x: Iterable[_] if x.isEmpty => Json.arr()
+        case x: Iterable[String] @unchecked if x.head.isInstanceOf[String] =>
           Json.fromValues(x.map(Json.fromString))
```
exyi, line 338: I think it would be nice to add a default case to all these `match` and throw an explicit error. just something like `throw Exception(s"$x is not valid text[]")`. In case we missed something here, it'd be easier to figure out what's wrong

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -174,6 +230,20 @@ object DataModelInstanceRelation {
...
+    case x: Int => jsonFromDouble(x.toDouble)
+    case x: Float => jsonFromDouble(x.toDouble)
+    case x: Long => jsonFromDouble(x.toDouble)
+    case x: java.math.BigDecimal => jsonFromDouble(x.doubleValue)
+    case x: java.math.BigInteger => jsonFromDouble(x.doubleValue)
+    case x: String => Json.fromString(x)
+    case x: Boolean => Json.fromBoolean(x)
+    case x: Array[Any] =>
+      Json.fromValues(x.map(parseValue))
+    case null => Json.Null // scalastyle:off null
```
exyi, line 244: since this is only used for processing filters, it feels tempting to just `parseJson(value.toString)` :joy:  (after checking for arrays and nulls)

exyi (commented):
:+1: good job

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
emelvarol-cognite: Returning Seq seemed convenient to me because there are some DMI filters expecting lists, and easy to just concatenate. Do you think we should still change this?

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
emelvarol-cognite: where is `parseJson` function 😠

emelvarol-cognite: show me pls

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
emelvarol-cognite: I couldn't understand this one :'(

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
emelvarol-cognite: done

emelvarol-cognite:
I adressed the comments other than https://github.com/cognitedata/cdp-spark-datasource/pull/596/files#r821076978, could you take a look? @exyi

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
exyi: I think it's `io.circe.parser.parse(x).right.get`

exyi (approved):
Perfect :)

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -89,6 +89,45 @@ class DataModelInstanceRelation(config: RelationConfig, modelExternalId: String)
...
+        Some(DMIRangeFilter(Seq(modelExternalId, attribute), gte = Some(parseValue(value))))
+      case GreaterThan(attribute, value) =>
+        Some(DMIRangeFilter(Seq(modelExternalId, attribute), gt = Some(parseValue(value))))
+      case LessThanOrEqual(attribute, value) =>
+        Some(DMIRangeFilter(Seq(modelExternalId, attribute), lte = Some(parseValue(value))))
+      case LessThan(attribute, value) =>
+        Some(DMIRangeFilter(Seq(modelExternalId, attribute), lt = Some(parseValue(value))))
+      case And(f1, f2) =>
+        (getInstanceFilter(f1) ++ getInstanceFilter(f2)).reduceLeftOption((sf1, sf2) =>
+          DMIAndFilter(Seq(sf1, sf2)))
```
emelvarol-cognite, line 115: I was writing a 4 case match case, then I thought there is probably a shortcut :D @exyi

PR #598, opened 2022-03-18
Use IO.blocking when dealing with queue in StreamIterator

vuhainguyen:
Spark is hanging because all the Threads `CDF-Spark-compute` is locked. Stack trace from thread dump in the spark UI
```
java.base@11.0.13/jdk.internal.misc.Unsafe.park(Native Method)
java.base@11.0.13/java.util.concurrent.locks.LockSupport.park(LockSupport.java:194)
java.base@11.0.13/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2081)
java.base@11.0.13/java.util.concurrent.ArrayBlockingQueue.put(ArrayBlockingQueue.java:367)
cognite.spark.v1.StreamIterator$.$anonfun$enqueueStreamResults$2(StreamIterator.scala:76)
cognite.spark.v1.StreamIterator$$$Lambda$1560/0x0000000840c22440.apply$mcV$sp(Unknown Source)
app//scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
cognite.shaded.cats.effect.IOFiber.runLoop(IOFiber.scala:283)
cognite.shaded.cats.effect.IOFiber.execR(IOFiber.scala:1327)
cognite.shaded.cats.effect.IOFiber.run(IOFiber.scala:139)
java.base@11.0.13/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
java.base@11.0.13/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
java.base@11.0.13/java.lang.Thread.run(Thread.java:829)
```
Use [IO.blocking](https://typelevel.org/cats-effect/docs/thread-model#blocking) when dealing with queue to move all the blocking task to the pool `CDF-Spark-blocking`
There are 2 minor changes I will comment on to make "the life of reviewer easier" 😁 , don't care about the reformat part

### DIFF
--- a/src/main/scala/cognite/spark/v1/CdpConnector.scala
+++ b/src/main/scala/cognite/spark/v1/CdpConnector.scala
@@ -113,7 +113,7 @@ object CdpConnector {
       retryingSttpBackend(
         config.maxRetries,
         config.maxRetryDelaySeconds,
-        config.partitions,
+        config.parallelismPerPartition,
         metricsPrefix)
     new GenericClient(
       applicationName = config.applicationName.getOrElse(Constants.SparkDatasourceVersion),
--- a/src/main/scala/cognite/spark/v1/StreamIterator.scala
+++ b/src/main/scala/cognite/spark/v1/StreamIterator.scala
@@ -2,8 +2,8 @@ package cognite.spark.v1
 
 import cats.effect.unsafe.IORuntime
 
-import java.util.concurrent.{ArrayBlockingQueue, Executors}
-import cats.effect.{Concurrent, IO}
+import java.util.concurrent.{ArrayBlockingQueue, Executors, TimeUnit}
+import cats.effect.IO
 import com.google.common.util.concurrent.ThreadFactoryBuilder
 import fs2.{Chunk, Stream}
 import org.log4s._
@@ -73,7 +73,7 @@ object StreamIterator {
     stream.chunks.parEvalMapUnordered(queueBufferSize) { chunk =>
       IO {
         val processedChunk = processChunk.map(f => f(chunk)).getOrElse(chunk)
-        queue.put(Right(processedChunk))
+        queue.offer(Right(processedChunk), 30, TimeUnit.SECONDS)
       }
     }
 
--- a/src/main/scala/cognite/spark/v1/StreamIterator.scala
+++ b/src/main/scala/cognite/spark/v1/StreamIterator.scala
@@ -36,7 +36,7 @@ object StreamIterator {
     val putOnQueueStream =
       enqueueStreamResults(stream, queue, queueBufferSize, processChunk)
         .handleErrorWith(e =>
-          Stream.eval(IO.blocking(queue.put(Left(e)))) ++ Stream.eval(IO {
+          Stream.eval(IO.blocking(queue.put(Left(e)))) ++ Stream.eval(IO.blocking {
             if (!drainPool.isShutdown) {
               drainPool.shutdownNow()
             }
--- a/src/main/scala/cognite/spark/v1/StreamIterator.scala
+++ b/src/main/scala/cognite/spark/v1/StreamIterator.scala
@@ -2,7 +2,7 @@ package cognite.spark.v1
 
 import cats.effect.unsafe.IORuntime
 
-import java.util.concurrent.{ArrayBlockingQueue, Executors, TimeUnit}
+import java.util.concurrent.{ArrayBlockingQueue, Executors}
 import cats.effect.IO
 import com.google.common.util.concurrent.ThreadFactoryBuilder
 import fs2.{Chunk, Stream}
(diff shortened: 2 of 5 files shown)


src/main/scala/cognite/spark/v1/CdpConnector.scala
```diff
@@ -113,7 +113,7 @@ object CdpConnector {
       retryingSttpBackend(
         config.maxRetries,
         config.maxRetryDelaySeconds,
-        config.partitions,
+        config.parallelismPerPartition,
```
vuhainguyen, line 116: third parameters is `maxParallelRequests` so should not be `partitions` here.
 ```scala
def retryingSttpBackend(
      maxRetries: Int,
      maxRetryDelaySeconds: Int,
      maxParallelRequests: Int = Constants.DefaultParallelismPerPartition,
      metricsPrefix: Option[String] = None
  )
```

File: src/test/scala/cognite/spark/v1/SparkTest.scala [hunk 2]
```diff
@@ -58,7 +65,8 @@ trait SparkTest {
   val spark: SparkSession = SparkSession
     .builder()
     .master("local[*]")
-    .config("spark.ui.enabled", "false")
+    .config("spark.ui.enabled", "false") //comment this and use the line below if you need to use Spark UI
+    //.config("spark.ui.port", "4041")
```
vuhainguyen, line 69: So that we can access to SparkUI when running test in local

src/main/scala/cognite/spark/v1/StreamIterator.scala
```diff
@@ -73,7 +73,7 @@ object StreamIterator {
     stream.chunks.parEvalMapUnordered(queueBufferSize) { chunk =>
       IO {
         val processedChunk = processChunk.map(f => f(chunk)).getOrElse(chunk)
-        queue.put(Right(processedChunk))
+        queue.offer(Right(processedChunk), 30, TimeUnit.SECONDS)
```
wjoel, line 76: Hm, I see. So all compute threads end up here, resulting in a deadlock, as no threads are available to do the reading from this queue?

If that's the case, I think the real fix might be to shift this to a different thread pool.

src/test/scala/cognite/spark/v1/SparkTest.scala
wjoel: 4040 is the default, this line isn't needed.

src/main/scala/cognite/spark/v1/CdpConnector.scala
wjoel: Nice catch.

src/main/scala/cognite/spark/v1/StreamIterator.scala
wjoel: I thought this would only be semantically blocking, not actually blocking the thread. :thinking:

wjoel: Oh, it's actually using `java.util.concurrent.ArrayBlockingQueue`. That's certainly thread blocking. Maybe it should be using a cats-effect `Queue` instead. Need to think about this for a bit.

wjoel: I think we're being totally stupid here, this should absolutely not be done on the cats-effect compute thread pool. :exploding_head:

wjoel: @vuhainguyen could  you try simply changing line 74 to `IO.blocking` (and no other changes) and testing that with the DMS test case you have? If this is the problem, I think it shouldn't hang any longer.

wjoel (approved):
Nice, this should make a lot of things better. :heart_eyes:

src/main/scala/cognite/spark/v1/StreamIterator.scala
hakontro: https://typelevel.org/cats-effect/docs/thread-model#blocking
Are we already seeing a CE3 win here? :cat2:

hakontro:
Nice! Please remember to change the PR title to fit the fix before merging :pray:

src/main/scala/cognite/spark/v1/StreamIterator.scala
wjoel: @hakontro yes, kind of! This is trivial to fix with CE3, in CE2 we would have had to create a thread pool manually and shifted and evaluated on and what not.

hakontro: Good news! From the article
> Notably, ContextShift and Blocker are no more.
Sounds like this is something we should look into, this isn't our first threadpool-related problem after all..

PR #600, opened 2022-03-20
custom implementation of RAW schema inferrence

exyi:
we used to rely on Spark's schema inferrence, but it had a few problems

* it is very slow with large number of columns
* we needed to launch a spark job just for schema
    inferrence, which increases latency (and depends on the
    cluster health)

this method should be significantly faster, so it should allow
us to increase the default inferrence limit in Jetfire.

### DIFF
--- a/src/main/scala/cognite/spark/v1/RawSchemaInferrer.scala
+++ b/src/main/scala/cognite/spark/v1/RawSchemaInferrer.scala
@@ -1,0 +1,104 @@
+package cognite.spark.v1
+
+import io.circe.Json
+import org.apache.spark.sql.types.{DataType, StructField, StructType}
+
+object RawSchemaInferrer {
+  // reimplementation of JsonInferSchema.infer from Spark which we used to rely on in the past
+  // our version works directly on circe Json and we can invoke it without starting a Spark job
+
+  def infer(j: Json): JsonType =
+    j.fold(
+      JsonNull,
+      _ => JsonBool,
+      num =>
+        if (num.toLong.isDefined) {
+          JsonInteger
+        } else {
+          // TODO: do we need to handle big integers?
+          JsonDouble
+      },
+      str =>
+        if (str.isEmpty) {
+          JsonNull
+        } else if (str.matches("^[-+]?\\d{1,18}$")) {
+          JsonInteger
+        } else if (str
+            .matches("^[-+]?\\d+(.\\d+)?(e[+-]?\\d+)?$") || str == "+Infinity" || str == "Infinity" || str == "-Infinity" || str == "NaN") {
+          JsonDouble
+        } else {
+          JsonString
+      },
+      array => JsonArray(array.map(infer).foldLeft(JsonNull: JsonType)(unify)),
+      obj => infer(obj.toIterable)
+    )
+
+  def infer(obj: Iterable[(String, Json)]): JsonObject =
+    JsonObject(
+      obj.map {
+        case (key, value) => key -> infer(value)
+      }.toMap
+    )
+
+  def isAssignableTo(from: JsonType, to: JsonType): Boolean =
+    (from, to) match {
+      case (JsonNull, _) => true
+      case (JsonInteger, JsonDouble) => true
+      case (_, JsonString) => true
+      case (JsonObject(x), JsonObject(y)) =>
+        x.keys.forall(k => y.contains(k) && isAssignableTo(x(k), y(k)))
+      case (JsonArray(x), JsonArray(y)) => isAssignableTo(x, y)
+      case (_, _) if from == to => true
+      case _ => false
+    }
+
+  def unify(a: JsonType, b: JsonType): JsonType =
+    (a, b) match {
+      case (_, _) if isAssignableTo(b, a) => a
+      case (_, _) if isAssignableTo(a, b) => b
+      case (x: JsonObject, y: JsonObject) => unifyObjects(x, y)
+      case (JsonArray(x), JsonArray(y)) => JsonArray(unify(x, y))
+
+      // anything can be put into a string - we just put the serialized JSON into it
+      case (_, _) => JsonString
+    }
+
+  def unifyObjects(a: JsonObject, b: JsonObject): JsonObject = {
+    val keys = a.fields.keySet.union(b.fields.keySet)
+    val fields = keys.iterator.map(k => k -> unify(a.field(k), b.field(k))).toMap
+    JsonObject(fields)
+  }
+
+  def toSparkSchema(j: JsonType): DataType = {
+    import org.apache.spark.sql.types._
+    j match {
+      case JsonNull => StringType
+      case JsonInteger => LongType
+      case JsonDouble => DoubleType
+      case JsonString => StringType
+      case JsonBool => BooleanType
+      case obj: JsonObject => toSparkSchema(obj)
+      case JsonArray(element) => ArrayType(toSparkSchema(element))
+    }
+  }
+
+  def toSparkSchema(j: JsonObject): StructType =
+    StructType(j.fields.map {
+      case (name, t) => StructField(name, toSparkSchema(t), nullable = true)
+    }.toArray)
+
+  sealed trait JsonType
+
+  /** null or empty string */
+  object JsonNull extends JsonType
+
+  /** int64 */
+  object JsonInteger extends JsonType
+  object JsonDouble extends JsonType
+  object JsonString extends JsonType
+  object JsonBool extends JsonType
+  final case class JsonObject(fields: Map[String, JsonType]) extends JsonType {
+    def field(name: String): JsonType = fields.getOrElse(name, JsonNull)
+  }
+  final case class JsonArray(element: JsonType) extends JsonType
+}
--- a/src/main/scala/cognite/spark/v1/RawSchemaInferrer.scala
+++ b/src/main/scala/cognite/spark/v1/RawSchemaInferrer.scala
@@ -1,12 +1,16 @@
 package cognite.spark.v1
 
+import cats.effect.kernel.Concurrent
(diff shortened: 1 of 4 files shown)


exyi:
Converting it to draft for now, it doesn't really seems to help

src/main/scala/cognite/spark/v1/RawSchemaInferrer.scala
```diff
@@ -0,0 +1,188 @@
...
+      i = skipPlusMinus(str, i + 1)
+      val i2 = skipInteger(str, i)
+      if (i2 == i) {
+        return false
+      }
+      i = i2
+    }
+
+    i == str.length
+  }
```
vuhainguyen, line 171: I don't get the logic here 😂 , this might need some comments to explain I think, I have a suggestion below to avoid using return
```scala
if ("NaN".equals(str)) {
  true
} else {
  var i = skipPlusMinus(str, 0)
  var whatever = true
  if (str.length == i + "Infinity".length && str.endsWith("Infinity")) {
    true
  } else {
    i = skipInteger(str, i)
    if (i < str.length && str(i) == '.') {
      i = skipInteger(str, i + 1)
    }
    if (i < str.length && (str(i) == 'e' || str(i) == 'E')) {
      i = skipPlusMinus(str, i + 1)
      val i2 = skipInteger(str, i)
      if (i2 == i) {
        whatever = false
      }
      i = i2
    }
    i == str.length && whatever
  }
}
```

src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -46,28 +46,36 @@ class RawTableRelation(
...
     }
   }
 
+  private def inferSchema(
+      limit: Int
+  ): IO[StructType] = {
+    val client = CdpConnector.clientFromConfig(config).rawRows(database, table)
+
+    val rows = client
+      .list(Some(limit))
```
wjoel, line 68: This isn't the one you want, and probably why it's not so fast, as it's not doing concurrent reads. ;) You should use `listConcurrently`. If you want to limit, just limit the resulting (combined) `fs2.Stream`.

More like we're doing here: https://github.com/cognitedata/profiler-api/blob/main/src/main/scala/com/cognite/raw_profiler/raw/RawClient.scala#L116

fwiw `Stream(streams: _*).parJoin(10)` seems to be a bit faster than the `.fold(Stream.empty)(_.merge(_))` we're doing in the Scala SDK.

src/main/scala/cognite/spark/v1/RawSchemaInferrer.scala
wjoel: I assume it's just copied from `JsonInferSchema.infer`?

src/test/scala/cognite/spark/v1/RawTableRelationTest.scala
```diff
@@ -755,4 +755,39 @@ class RawTableRelationTest
     err.getMessage shouldBe "Error while loading RAW row [key='k'] in column 'value': java.lang.NumberFormatException: For input string: \"test\""
 
   }
+
+  private def inferSchema(json: String): DataType = {
+    val circeJson = io.circe.parser.parse(json).right.get
+    RawSchemaInferrer.toSparkSchema(RawSchemaInferrer.infer(circeJson))
+  }
+
+  "RawSchemaInferrer" should "handle integers" in {
```
wjoel, line 764: Maybe we can get some additional good test cases from Spark's `JsonInferSchema`?

src/main/scala/cognite/spark/v1/RawSchemaInferrer.scala
exyi: No it isn't, I'll add a few comments. (Spark uses jackson parser which seems to do this somehow automatically).

src/main/scala/cognite/spark/v1/RawTableRelation.scala
exyi: will do that :+1:. Currently it's not an issue, since we still use the 10k (one page) limit, but it will be very important for increasing the infer schema limit.

src/main/scala/cognite/spark/v1/RawSchemaInferrer.scala
exyi: I have tried to better explain what the code does. I first implemented it using a regex, but that was pretty slow, so I just handwrote the parser.

General principle is that I scan the entire input string with the `i` index and check if all characters are allowed in this context. The parsed format is something like this: `123[.456][e-10]`, so:
* skip the numbers
* if there is decimal part, skip the `.` and skip another sequence of numbers
* if there is the exp part, skip `e` and the number
* if we made it to the end (i==str.length), it looks like a valid number

exyi:
I have done a bit more testing and now I'm quite sure that this actually improves performance a bit. It's a bit hard to measure, since I'm limited by my network speed, but from profiling both versions I can conclude that schema inference now consumes 4-times less CPU time than before (this does not mean that RAW is 4-time faster, there is other things slowing it down).

With the parallel retrieval Joel suggested, it's actually significantly faster end-to-end. Before about 40sec, now at about 25sec on the `test-emel/test.perf-problem-repro` table with 20k inference limit.

src/main/scala/cognite/spark/v1/RawSchemaInferrer.scala
vuhainguyen: Can't we just do it simple:
```scala
def isFloat(s: String) =  
  try {
    s.toFloat
    true
  } catch {
    case _ : Throwable => false
  }
```
It works even for `NaN`, `+-Infinity` and `123[.456][e-10] `

wjoel: `def isFloat(s: String) = Try(s.toFloat).isSuccess` if we want to do it that way. ;) Maybe less efficient, but a lot easier.

Maybe we should use circe?
Their decoder https://github.com/circe/circe/blob/main/modules/core/shared/src/main/scala/io/circe/Decoder.scala#L675 uses https://github.com/circe/circe/blob/main/modules/core/shared/src/main/scala/io/circe/JsonNumber.scala#L248 which uses https://github.com/circe/circe/blob/e6685317bc9e6e65539bb6194900e0e754fbcb7c/modules/numbers/shared/src/main/scala/io/circe/numbers/BiggerDecimal.scala#L257

wjoel:
@exyi CI fails on style checks because `Avoid using return`.

src/main/scala/cognite/spark/v1/RawSchemaInferrer.scala
exyi: Ok, I'll try that, but I expect to get much worse results from that. This function is called for every string value in the RAW table and most of the time it will fail - for 350 string columns * only 20k rows -> 7 mil exceptions. I don't think JVM can throw this many exceptions in few seconds

wjoel: Yeah, it might be a bad choice for performance reasons, but the circe option might not be. It just seems to me like there might be quite a few edge cases that we haven't thought about. I also looked at jsoniter-scala, and that code looks even more scary, but also due to performance optimizations.

exyi: Ok no.... I'm stupid, seems that Spark does not do this inferrence from strings, so we can just remove it :)

wjoel:
Just some quick benchmark results doing a count on a table of 31 million rows, with full schema inference:
2.0.8: 120.3 seconds
This PR: 124.8 seconds
This PR but inferSchema = false: 115.2 seconds

So I think we need to look into how this is/isn't being parallelized. I'm not sure about the assignability stuff around objects, will try to come up with some tests or counter-examples.

TharinduDG:
> Just some quick benchmark results doing a count on a table of 31 million rows, with full schema inference: 2.0.8: 120.3 seconds This PR: 124.8 seconds This PR but inferSchema = false: 115.2 seconds
> 
> So I think we need to look into how this is/isn't being parallelized. I'm not sure about the assignability stuff around objects, will try to come up with some tests or counter-examples.


AFA i understand, this is inefficient because it is more or less iterating the data twice. 
- once for parsing data to circe json tokens https://github.com/cognitedata/cdp-spark-datasource/blob/21297b56248670ce2425b93ce0b4fa57f0a19c61/src/main/scala/cognite/spark/v1/RawSchemaInferrer.scala#L45
- another to infer and merge the schema https://github.com/cognitedata/cdp-spark-datasource/blob/21297b56248670ce2425b93ce0b4fa57f0a19c61/src/main/scala/cognite/spark/v1/RawSchemaInferrer.scala#L50-L51

In spark it does in one pass
- https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/json/JsonDataSource.scala#L102-L111
- https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JsonInferSchema.scala#L127-L222

IMO, the best option would be to change `samplingRatio` in `JSONOptions` to something like 0.4 and see how it does.
Also we can try changing other config params to suit our complex requirements. 
https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JSONOptions.scala#L55

Also I looked into this: https://openproceedings.org/2017/conf/edbt/paper-62.pdf
I think the spark json schema inference timing kind of reasonable compared to this approach proposed by the paper.

Also doing this schema inferencing by ourself will isolate us from utilising future spark improvements as well.

wjoel (commented):
This is my list of "requested for review" even after I left Cognite, so... I'm trying a "comment" to see if I can get rid of it, if not, maybe someone else could please remove me from the list. ;)

PR #601, opened 2022-03-21
Add a test to investigate why subqueries aren't filtered

hakontro:


hakontro:
This was for `is_new` investigations, which has been fixed, so this is no longer relevant.

PR #602, opened 2022-03-21
Use property type for data model instances

emelvarol-cognite:


src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -74,44 +72,43 @@ class DataModelInstanceRelation(config: RelationConfig, modelExternalId: String)
       dmi: DataModelInstanceQueryResponse,
       requiredPropsArray: Seq[String]): ProjectedDataModelInstance = {
     val dmiProperties = dmi.properties.getOrElse(Map())
-    ProjectedDataModelInstance(
+    val res = ProjectedDataModelInstance(
```
hakontro, line 75: I don't think you need the `res` thing here, perhaps it's leftovers from development?

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -74,44 +72,43 @@ class DataModelInstanceRelation(config: RelationConfig, modelExternalId: String)
...
       externalId = dmi.properties
         .flatMap(_.get("externalId"))
-        .flatMap(_.asString)
+        .map(_.asInstanceOf[StringProperty].value)
         .getOrElse(
           throw new CdfSparkException("Can't read data model instance, `externalId` is missing.")),
       properties = requiredPropsArray.map { name: String =>
```
hakontro, line 81: Are these properties required to exist in the result, as null or some, or are they actually required to be populated? Seems this code accepts null entries for the required properties, which was a little puzzling to me.

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -197,81 +194,79 @@ class DataModelInstanceRelation(config: RelationConfig, modelExternalId: String)
           (index, field.name, propertyType)
       }
 
-    def parseRow(row: Row): DataModelInstance = {
+    def parseRow(row: Row): DataModelInstanceCreate = {
```
hakontro, line 197: If we curried it like `parseRow(row: Row)(schema: StructType)`, we could avoid the function in the function. `parseRow` now uses parameters that are in the outer function `fromRow`, and I think it's nicer to be able to have it all in one place!

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -197,81 +194,79 @@ class DataModelInstanceRelation(config: RelationConfig, modelExternalId: String)
...
-            row.get(index),
-            (_: Any) =>
-              throw SparkSchemaHelperRuntime
-                .badRowError(row, name, propT._1, "")
-          )
-      }.toMap
-      DataModelInstance(modelExternalId, properties = Some(propertyValues))
+      val propertyValues: Map[String, PropertyType] = properties
+        .map {
+          case (index, name, propT) =>
```
hakontro, line 204: Could some of this code be nicer if we made a case class for `propertyTypes`? Maybe there's already one we can use!
`val propertyTypes: Map[String, (String, Boolean)]`

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -197,81 +194,79 @@ class DataModelInstanceRelation(config: RelationConfig, modelExternalId: String)
...
+    case x: Array[Double] => ArrayProperty(x.toVector.map(Float64Property))
+    case x: Array[Int] => ArrayProperty(x.toVector.map(Int32Property))
+    case x: Array[Float] => ArrayProperty(x.toVector.map(Float32Property))
+    case x: Array[Long] => ArrayProperty(x.toVector.map(Int64Property))
+    case x: Array[String] => ArrayProperty(x.toVector.map(StringProperty))
+    case x: Array[Boolean] => ArrayProperty(x.toVector.map(BooleanProperty))
+    case x: Array[java.math.BigDecimal] =>
+      ArrayProperty(x.toVector.map(i => Float64Property(i.doubleValue())))
+    case x: Array[java.math.BigInteger] =>
+      ArrayProperty(x.toVector.map(i => Int64Property(i.longValue())))
```
hakontro, line 257: Let's add a final `case: _` just in case something crazy happens.

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -197,81 +194,79 @@ class DataModelInstanceRelation(config: RelationConfig, modelExternalId: String)
...
     parseRow
   }
 }
 
 object DataModelInstanceRelation {
-  private def propertyNotNullableMessage(propertyType: String) =
-    s"Property of $propertyType type is not nullable."
   private def unknownPropertyTypeMessage(a: Any) = s"Unknown property type $a."
 
   private def notValidPropertyTypeMessage(a: Any, propertyType: String) =
```
hakontro, line 232: This seems to be thrown when someone tries to write something that can't be parsed to the property type of the schema. We accept multiple types of numbers casted to Float32 for example, perhaps there's a more informative way to let the user know what they need to do!

hakontro (commented):
Very nice, had some comments.

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -291,92 +286,103 @@ object DataModelInstanceRelation {
...
-        case x: java.math.BigDecimal => jsonFromDouble(x.doubleValue)
-        case x: java.math.BigInteger => jsonFromDouble(x.doubleValue)
+      case "float32" => {
+        case x: Float => Float32Property(x)
+        case x: Double => Float32Property(x.toFloat)
+        case x: Int => Float32Property(x.toFloat)
+        case x: Long => Float32Property(x.toFloat)
+        case x: java.math.BigDecimal => Float32Property(x.floatValue())
+        case x: java.math.BigInteger => Float32Property(x.floatValue())
+        case a => throw new CdfSparkException(notValidPropertyTypeMessage(a, propertyType))
```
vuhainguyen, line 298: I think handling each case in a separate function will reduce the cyclomatic complexity and make the code easier to read hehe 😁

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
emelvarol-cognite: required means, the selected properties i believe, name is just deceiving.

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
emelvarol-cognite: yis

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
emelvarol-cognite: The beauty of this version is that the part in `fromRow` is done once and parseRow is called for each row. I'll try moving the first part out to another function.

hakontro: That _is_ nice.. but I think we get the same benefit from currying.. Do you know @wjoel ?

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -291,92 +286,103 @@ object DataModelInstanceRelation {
...
-        case x: Iterable[_] if x.isEmpty => Json.arr()
+        case x: Iterable[_] if x.isEmpty => ArrayProperty(Vector.empty)
         case x: Iterable[Boolean] @unchecked if x.head.isInstanceOf[Boolean] =>
-          Json.fromValues(x.map(Json.fromBoolean))
+          ArrayProperty(x.toVector.map(BooleanProperty))
+        case a => throw new CdfSparkException(notValidPropertyTypeMessage(a, propertyType))
+      }
+      case "int[]" | "int32[]" => {
+        case x: Iterable[_] if x.isEmpty => ArrayProperty(Vector.empty)
+        case x: Iterable[Int] @unchecked if x.head.isInstanceOf[Int] =>
```
wjoel, line 372: I think you can just do `case x: Iterable[_] if x.head.isInstanceOf[Int]` here.

The `@unchecked` is needed because due to type erasure it's not truly possible to match on `Iterable[Int]`, at runtime it's only `Iterable` with no type information stored for the element types, so the type check against `Iterable[Int]` is a lie, and that's why the compiler requires us to mark it as `@unchecked` (as it can't be checked).

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -291,92 +286,103 @@ object DataModelInstanceRelation {
...
+          ArrayProperty(x.toVector.map(Int32Property))
+        case x: Iterable[java.math.BigInteger] @unchecked if x.head.isInstanceOf[java.math.BigInteger] =>
+          ArrayProperty(x.toVector.map(i => Int32Property(i.intValue())))
```
wjoel, line 375: I'm not sure we should allow this. Maybe we have to because `BigInteger` is used often by Spark, even if the values are small? But it will not work well if the value is larger than `Integer.MaxValue`...

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
emelvarol-cognite: I thought it is just a way of grouping the arguments? 🤔

hakontro: I may not be using the term right, but essentially having multiple argument lists to the function, so that we can do
`def parseRow(row: Row)(schema: StructType): DataModelInstanceCreate`
and then use it like
`parseRow(someSchema)`, which will then return a function of the type `Row => DataModelInstanceCreate`..

Seems the actual definition is https://www.baeldung.com/scala/currying#currying

> Currying is the process of converting a function with multiple arguments into a sequence of functions that take one argument. Each function returns another function that consumes the following argument.

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
emelvarol-cognite: yes sure, we have a type from scala-sdk which is nice

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
emelvarol-cognite: I thought you wanted me to remove the inner function and make it one, sorry I misunderstood

emelvarol-cognite: we don't make use of `schema: StructType` directly in `parseRow`. It uses an array calculated in `fromRow` using the schema. Should I take that one as input?

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
wjoel: I don't know how often they end up being inferred, so I'm a bit unsure, but I guess we can remove it for now, and add it back if we find out it's a common thing?

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
wjoel: It's not about IntelliJ's type matching, it's about the JVM and type erasure.
At runtime, the type of `x` will be `Iterable`. It's not `Iterable[Int]`, `Iterable[Double]`, etc. it's compiled into only `Iterable`. Type parameters don't exist in JVM bytecode.

The Scala compiler knows this, and it knows that it can't actually check if `x` is of type `Iterable[Int]`. That's why it fails to compile. Adding `@unchecked` says "trust me, I know what I'm doing" and in our case we know what we're doing (we think) because of the `x.head.isInstanceOf[Int]` check.

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
emelvarol-cognite: I already hit class cast exception in the tests when I removed `BigDecimal` from `float32` and `BigInteger` from `int`. So I keep it if ok.

src/test/scala/cognite/spark/v1/DataModelInstancesRelationTest.scala
```diff
@@ -39,16 +39,13 @@ class DataModelInstancesRelationTest
...
 
   private def byExternalId(modelExternalId: String, externalId: String): String =
     listInstances(
       modelExternalId,
-      filter = Some(DMIEqualsFilter(Seq("instance", "externalId"), Json.fromString(externalId))))
-      .flatMap(_.properties.flatMap(_.get("externalId")).toList)
-      .flatMap(_.asString.toList)
-      .head
+      filter = Some(DMIEqualsFilter(Seq("instance", "externalId"), StringProperty(externalId))))
+      .flatMap(_.properties.flatMap(_.get("externalId")).toList).head.asInstanceOf[StringProperty].value
```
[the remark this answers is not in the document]
vuhainguyen: Suggestion:
```scala
    listInstances(
      modelExternalId,
      filter = Some(DMIEqualsFilter(Seq("instance", "externalId"), StringProperty(externalId))))
      .flatMap(_.properties.flatMap(_.get("externalId")))
      .collectFirst { case StringProperty(id) => id }
      .getOrElse("")
```
If you don't want to have getOrElse at the end then return `Option[String]` and later do this when comparing
```scala
byExternalId(primitiveExtId, randomId) shouldBe Some(randomId)
```

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -77,14 +63,12 @@ class DataModelInstanceRelation(config: RelationConfig, modelExternalId: String)
     ProjectedDataModelInstance(
       externalId = dmi.properties
         .flatMap(_.get("externalId"))
-        .flatMap(_.asString)
+        .map(_.asInstanceOf[StringProperty].value)
```
vuhainguyen, line 66: I think you can also use `collect { case StringProperty(externalId) => externalId }` here instead

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -93,25 +77,25 @@ class DataModelInstanceRelation(config: RelationConfig, modelExternalId: String)
...
+        Some(DMIRangeFilter(Seq(modelExternalId, attribute), gt = Some(parsePropertyValue(value))))
       case LessThanOrEqual(attribute, value) =>
-        Some(DMIRangeFilter(Seq(modelExternalId, attribute), lte = Some(parseValue(value))))
+        Some(DMIRangeFilter(Seq(modelExternalId, attribute), lte = Some(parsePropertyValue(value))))
       case LessThan(attribute, value) =>
-        Some(DMIRangeFilter(Seq(modelExternalId, attribute), lt = Some(parseValue(value))))
+        Some(DMIRangeFilter(Seq(modelExternalId, attribute), lt = Some(parsePropertyValue(value))))
       case And(f1, f2) =>
         (getInstanceFilter(f1) ++ getInstanceFilter(f2)).reduceLeftOption((sf1, sf2) =>
           DMIAndFilter(Seq(sf1, sf2)))
```
vuhainguyen, line 101: I have a doubt about the case Or:
```scala
case Or(f1, f2) =>
    (getInstanceFilter(f1), getInstanceFilter(f2)) match {
      case (Some(sf1), Some(sf2)) =>
        Some(DMIOrFilter(Seq(sf1, sf2)))
      case _ =>
        None
    }
```
Don't we want  to have 2 case below as well ? 
```scala
case (Some(sf1), None) =>
            Some(DMIOrFilter(Seq(sf1)))
case (None, Some(sf2)) =>
            Some(DMIOrFilter(Seq(sf2)))
```

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -179,99 +163,93 @@ class DataModelInstanceRelation(config: RelationConfig, modelExternalId: String)
...
+                Some(toPropertyType(propT.`type`)(row.get(index)))
+            })
+        }
+        .collect { case (a, Some(value)) => a -> value }
+        .toMap
+
+      DataModelInstanceCreate(modelExternalId, properties = Some(propertyValues))
     }
-    parseRow
+    parseRow(indexedPropertyList)
```
emelvarol-cognite, line 207: @hakontro does this look better, can i resolve this one?

src/test/scala/cognite/spark/v1/DataModelInstancesRelationTest.scala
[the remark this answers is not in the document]
emelvarol-cognite: I thought here if the code and the API works correctly, we would always have a head so worst case the test fail as it should. But I'll fix it 👍

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
emelvarol-cognite: I had a reason for this, let me try recalling...

emelvarol-cognite: So the reason we get `None` here, is we can't match the filter any case here. So we should not push down by only one condition we can match to something. It may be a wider filter when we `or` and we pushdown more than we should. So we let spark filter the whole thing in that case, not to miss requested data.

src/test/scala/cognite/spark/v1/DataModelInstancesRelationTest.scala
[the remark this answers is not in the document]
wjoel: It should return `Option[String]` or throw an exception in `getOrElse` with a message explaining what went wrong (if it's unexpected and we can't recover). Converting to an empty string is no good.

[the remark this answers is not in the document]
emelvarol-cognite: took @vuhainguyen 's suggestion without `getOrElse`

emelvarol-cognite:
> I found a typo, and there's uncommented stuff that needs to be cleaned up, but overall this looks really good now.

Thanks a lot for reviewing again! Can you point me to the typo pls? I'll go through the pr to find commented stuff and clean up. Also, I want to add a test for reads on new types I added. Then I'll ping you again for a review again 🙏

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -291,98 +287,169 @@ object DataModelInstanceRelation {
...
+  }
+
+  private def toTimestampProperty: Any => TimeStampProperty = {
+    case x: Instant =>
+      TimeStampProperty(OffsetDateTime.ofInstant(x, ZoneId.systemDefault()).toZonedDateTime)
+    case x: java.sql.Timestamp =>
+      // Using ZonedDateTime directly without OffSetDateTime, adds ZoneId to the end of encoded string
+      // 2019-10-02T07:00+09:00[Asia/Seoul] instead of 2019-10-02T07:00+09:00, API does not like it.
+      TimeStampProperty(OffsetDateTime.ofInstant(x.toInstant, ZoneId.systemDefault()).toZonedDateTime)
+    case a => throw new CdfSparkException(notValidPropertyTypeMessage(a, "timestamp"))
```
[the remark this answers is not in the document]
wjoel: It is nice, but we shouldn't use the system default `ZoneId`, rather convert it to UTC explicitly.
Maybe `LocalDateTime.from(ts.toLocalDateTime.atZone(ZoneOffset.UTC))` or similar. Tests would be needed to make sure we don't get it wrong for different time zones.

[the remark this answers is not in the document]
emelvarol-cognite: `TimeStampProperty` wants `ZonedDateTime` as value type. I'll see if it can be adapted to that.

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -179,99 +165,104 @@ class DataModelInstanceRelation(config: RelationConfig, modelExternalId: String)
...
+      ArrayProperty(x.toVector.map(i => Float64Property(i.doubleValue())))
+    case x: Array[java.math.BigInteger] =>
+      ArrayProperty(x.toVector.map(i => Int64Property(i.longValue())))
+    case x: LocalDate => DateProperty(x)
+    case x: java.sql.Date => DateProperty(x.toLocalDate)
+    case x: LocalDateTime => DateProperty(x.toLocalDate)
+    case x: Instant => TimeStampProperty(x.atZone(ZoneId.systemDefault()))
+    case x: java.sql.Timestamp =>
+      TimeStampProperty(x.toInstant.atZone(ZoneId.systemDefault()))
+    case x => throw new CdfSparkException(s"Cannot parse the value with udentified type: $x")
```
wjoel, line 248: Here's the typo, in "udentified". I don't know what happened to my previous comment... but my suggestion was `s"Unsupported value ${x.toString} of type ${x.getClass.getName}"`

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -179,99 +165,104 @@ class DataModelInstanceRelation(config: RelationConfig, modelExternalId: String)
...
+    case Float64Property(value) => value
+    case BooleanProperty(value) => value
+    case StringProperty(value) => value
+    case DateProperty(value) => value
+    case TimeStampProperty(value) => value
+    case DirectRelationProperty(value) => value
+    case GeographyProperty(value) => value
+    case GeometryProperty(value) => value
+    case ArrayProperty(values) => values.map(fromPropertyType)
+    case x => throw new CdfSparkException(s"Unknown property type with value $x")
```
wjoel, line 264: Similarly, `s"Unknown property type ${x.getClass.getName} with value ${x.toString}"` could be more helpful.

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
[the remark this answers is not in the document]
wjoel: Well, this PR is quite big already, so it might be good to get this merged first even with a simple-but-not-always-correct solution for timestamps first, and then add tests and fix edge cases in another (shorter) PR.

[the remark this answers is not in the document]
emelvarol-cognite: Sure I think this version can be merged 👍 I can create another PR to test type conversions and filters on timestamp etc.

emelvarol-cognite:
Can I get one last look? I believe it is clean now. I'll add tests for timestamp conversions and check filters on new types as a separate PR as it is already pretty big (as @wjoel suggested)

wjoel (approved):
LGTM, nice work!

wjoel:
For the next few PRs it would be great if we could get the test coverage up to what it was, -1.24% is significant and we should have tests that cover all types, eventually.

PR #603, opened 2022-03-24
Use CogniteId when calling api of sequenceRows

vuhainguyen:
Follow up of https://github.com/cognitedata/cognite-sdk-scala/pull/453

### DIFF
--- a/src/main/scala/cognite/spark/v1/SequenceRowsRelation.scala
+++ b/src/main/scala/cognite/spark/v1/SequenceRowsRelation.scala
@@ -89,8 +89,8 @@ class SequenceRowsRelation(config: RelationConfig, sequenceId: CogniteId)(val sq
         }
       val projectedRows =
         client.sequenceRows
-          .queryById(
-            sequenceInfo.id,
+          .query(
+            CogniteInternalId(sequenceInfo.id),
             filter.inclusiveStart,
             filter.exclusiveEnd,
             limit,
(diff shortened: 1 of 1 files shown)


src/main/scala/cognite/spark/v1/SequenceRowsRelation.scala
```diff
@@ -89,8 +89,8 @@ class SequenceRowsRelation(config: RelationConfig, sequenceId: CogniteId)(val sq
         }
       val projectedRows =
         client.sequenceRows
-          .queryById(
-            sequenceInfo.id,
+          .query(
+            CogniteInternalId(sequenceInfo.id),
```
wjoel, line 93: I think you can just use `sequenceId` from `class SequenceRowsRelation(config: RelationConfig, sequenceId: CogniteId)`?

src/main/scala/cognite/spark/v1/SequenceRowsRelation.scala
```diff
@@ -228,16 +228,9 @@ class SequenceRowsRelation(config: RelationConfig, sequenceId: CogniteId)(val sq
         .toList
         .parTraverse {
           case (cogniteExternalId, rows) =>
```
wjoel, line 230: Looks like this is incorrectly named and should be `cogniteId` (or just `id`) instead.

wjoel (approved):

src/main/scala/cognite/spark/v1/SequenceRowsRelation.scala
vuhainguyen: right, we get the `sequenceInfo` by using that `sequenceId` any way :)

src/main/scala/cognite/spark/v1/SequenceRowsRelation.scala
vuhainguyen: correct, updated :)

PR #604, opened 2022-04-05
Add missing assethierarchy.deleted metrics

exyi:


src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -78,6 +78,7 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
         .fromIterator[IO](deletes, chunkSize = batchSize)
         .chunks
         .parEvalMapUnordered(config.parallelismPerPartition) { chunk =>
+          itemsDeleted.inc(chunk.size)
```
vuhainguyen, line 81: I think it's better if we use `incMetrics` instead, it's standard like in the rest of the file and there's a guard check `if (config.collectMetrics)` ;)

```suggestion
                   client.assets
            .deleteRecursive(
              chunk.toVector.map(_.toCogniteId),
              recursive = true,
              ignoreUnknownIds = true)
            .flatTap(_ => incMetrics(itemsDeleted, chunk.size))
```

PR #607, opened 2022-05-05
Cleanup imports

TharinduDG:


src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -1,17 +1,15 @@
...
-import cats.syntax._
 import cognite.spark.v1.PushdownUtilities.stringSeqToCogniteExternalIdSeq
 import cognite.spark.v1.SparkSchemaHelper.{fromRow, structType}
-import com.cognite.sdk.scala.common.{CdpApiException, NonNullableSetter, SetValue}
+import com.cognite.sdk.scala.common.{CdpApiException, SetValue}
 import com.cognite.sdk.scala.v1.{Asset, AssetCreate, AssetUpdate, AssetsFilter, CogniteExternalId}
+import fs2.{Chunk, Stream}
 import io.scalaland.chimney.dsl._
-import org.apache.spark.sql.types.{DataTypes, StructType}
+import org.apache.spark.sql.types._
```
hakontro, line 11: Prefer the specific over `_` here (and most places as long as it doesn't get too many).

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -13,6 +11,8 @@ import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types._
 import org.apache.spark.sql.{Row, SQLContext}
 
+import java.time._
```
hakontro, line 14: So this one is ok for example!

src/main/scala/cognite/spark/v1/DataSetsRelation.scala
```diff
@@ -1,16 +1,15 @@
...
 import cognite.spark.v1.PushdownUtilities.{executeFilterOnePartition, pushdownToFilters, timeRange}
 import cognite.spark.v1.SparkSchemaHelper._
 import com.cognite.sdk.scala.common.WithId
 import com.cognite.sdk.scala.v1.resources.DataSets
 import com.cognite.sdk.scala.v1.{DataSet, DataSetCreate, DataSetFilter, DataSetUpdate, GenericClient}
 import io.scalaland.chimney.Transformer
-import org.apache.spark.sql.{Row, SQLContext}
 import org.apache.spark.sql.sources.{Filter, InsertableRelation}
-import org.apache.spark.sql.types.{DataTypes, StructType}
+import org.apache.spark.sql.types._
```
hakontro, line 11: ..but this one I disagree with (it's nice to be able to see a list of what's in use from a library from here). Nit-picky, I know :)

hakontro (approved):
Nice chore, thanks for taking the time! I'f prefer in general to revert the 
`import org.apache.spark.sql.types._`
to
`import org.apache.spark.sql.types.{DataTypes, StructType}`
otherwise good to go

PR #608, opened 2022-05-20
CDF-15674 Implement sequences upsert and support updating columns

emelvarol-cognite:


src/main/scala/cognite/spark/v1/SequencesRelation.scala
```diff
@@ -91,43 +141,84 @@ class SequencesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
   override def getFromRowsAndCreate(rows: Seq[Row], doUpsert: Boolean = true): IO[Unit] = {
     val sequences =
       rows
-        .map(fromRow[SequenceUpdateSchema](_))
+        .map(fromRow[SequenceUpsertSchema](_))
 
     implicit val toCreate =
       Transformer
-        .define[SequenceUpdateSchema, SequenceCreate]
+        .define[SequenceUpsertSchema, SequenceCreate]
```
emelvarol-cognite, line 189: I didn't like defining these implicit transformers inside functions multiple times but they are dynamically defined according to the list of columns of sequences. And moving outside does not work. Let me know if there is a nicer way.

src/main/scala/cognite/spark/v1/SequencesRelation.scala
```diff
@@ -141,25 +131,19 @@ class SequencesRelation(config: RelationConfig)(val sqlContext: SQLContext)
...
-        .withFieldComputed(
-          _.columns,
-          x => x.getSequenceColumnCreate
-        )
-        .buildTransformer
-
-    val existingSeqs: Map[Option[Long], Seq[String]] = getExistingSequenceColumnsByIds(sequences)
-    val existingSeqsWithExtId: Map[Option[String], Seq[String]] =
-      getExistingSequenceColumnsByExternalIds(sequences)
+  private def transformerUpsertToCreate() =
```
emelvarol-cognite, line 134: this didn't work for me when I tried, maybe it was sth else that I fixed :D :D thanks @vuhainguyen, probably the implicit parameters that I added to some helpers were missing and I blamed this instead...

src/main/scala/cognite/spark/v1/SequencesRelation.scala
vuhainguyen: It just need to find the implicit val, so you can define those as private def and declare the implicit val inside each bloc to retrieve the value from those private def 😁 , sorry I wanted to create a PR to your branch but push in the wrong same branch 🤦

src/main/scala/cognite/spark/v1/SequencesRelation.scala
vuhainguyen: 👯 :D

vuhainguyen: may be you can move the bloc below 
```scala
    val existingSeqs: Map[Option[Long], Seq[String]] = getExistingSequenceColumnsByIds(sequences)
    val existingSeqsWithExtId: Map[Option[String], Seq[String]] =
      getExistingSequenceColumnsByExternalIds(sequences)
```
into the transformer update to avoid duplicate code as well 😁

emelvarol-cognite: done  :p

vuhainguyen (approved):

src/main/scala/cognite/spark/v1/SequencesRelation.scala
```diff
@@ -252,25 +253,39 @@ final case class SequenceUpsertSchema(
...
+        val (modifyMap, addMap) =
+          requestedCols.partition(item => existingColumnExternalIds contains item._1)
+
+        // Column.modify do not allow valueType updates,
+        // If user tries to change the valueType, we abort.
+        existingColumns.foreach {
+          case (externalId: String, valueType: String) =>
+            val newValueType = modifyMap.get(externalId).flatMap(_.valueType)
+            if (newValueType.nonEmpty && !newValueType.contains(valueType)) {
+              throw new CdfSparkIllegalArgumentException(
```
emelvarol-cognite, line 281: @vuhainguyen dis

PR #620, opened 2022-06-23
Attempt to fix null pointer error

emelvarol-cognite:


src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
```diff
@@ -61,16 +61,16 @@ class DataModelInstanceRelation(
     if (modelType == DataModelType.EdgeType) {
       modelProps ++ Map(
         "externalId" -> DataModelPropertyDefinition(PropertyType.Text, false),
-        "type" -> DataModelPropertyDefinition(PropertyType.Text, false),
+        "nodeType" -> DataModelPropertyDefinition(PropertyType.Text, false),
```
silvavelosa, line 64: edgeType?

silvavelosa (approved):
✔️

src/main/scala/cognite/spark/v1/DataModelInstancesRelation.scala
emelvarol-cognite: true and now I am not sure if this is the fix :D

PR #621, opened 2022-06-28
Set up Scalafix

hakontro:
This isn't yet compiling, simply because we have a lot of patterns that scalafix isn't happy about. Leave it up to you to find the right settings here, I'm sure you'll want to ignore at least some of warnings.


src/main/scala/cognite/spark/v1/package.scala
```diff
@@ -32,22 +30,14 @@ package object v1 {
...
-      override def transform(src: T): Option[Setter[T]] = src match {
-        case null => Some(SetNull()) // scalastyle:ignore null
-        case value => Some(SetValue(value))
+      override def transform(src: T): Option[Setter[T]] = Option(src) match {
+        case None => Some(SetNull())
+        case Some(value) => Some(SetValue(value))
       }
     }
 
-  implicit def encodeSetter[T](implicit encodeT: Encoder[T]): Encoder[Setter[T]] =
```
wjoel, line 44: We already have those encoders in the SDK.

vuhainguyen (approved):
LGTM, but small doubt that running time of tests is 19mins, which is slower than other opened PR ([13 mins](https://github.com/cognitedata/cdp-spark-datasource/runs/7230399737?check_suite_focus=true) )or previous merged one ([16 mins](https://github.com/cognitedata/cdp-spark-datasource/runs/7025407662?check_suite_focus=true) though), may be it just depends on the resource of the CI ...

wjoel:
@vuhainguyen it looks like there's a lot of variance in the time it takes to run those tests, I have a run that finished in 15m, same as https://github.com/cognitedata/cdp-spark-datasource/pull/623

wjoel:
I want to run some manual benchmarks before merging this, though.

vuhainguyen:
> I want to run some manual benchmarks before merging this, though.

That would be great, also if you can share in the description of the PR how did you do it as well hehe 😁

wjoel:
I've tested with my `wjoel-blue` project, and it doesn't use more resources or take more time.

Using `/usr/bin/time -v` it reports similar max resident memory sizes when doing a `.count` on a very large RAW table, which takes 150-190s to complete (it varies quite a bit, the best result was 158s using this branch) and uses 3.3 GB of memory.

When doing the same `.count` on assets twice:
This PR: 14.6s, 7.6s, 2.2 GB max resident memory.
2.1.0: 15.4s, 8.0s, 2.4 GB max resident memory.

So, it's not worse. Possibly slightly better, but it might just be noise.

wjoel:
The "unused IO" rule actually crashes so we can't use that one, but everything else is a big win. Nice work, @hakontro. ;)

hakontro:
Thanks, glad it works 😃 👋

PR #624, opened 2022-07-20
[DO NOT MERGE] Add temporary test to check partition and client created

vuhainguyen:
Just a dummy test that read assets and write to events to see how many partitions and clients that we created


audunska:
Seems outdated. Closing.

PR #626, opened 2022-07-28
Add test for deleting sequence

vuhainguyen:
We don't have test for deleting sequence yet, so just add one to make sure it work


emelvarol-cognite (approved):
awsom

PR #628, opened 2022-08-17
Add originalToken for OAuth2 client credentials or sessions

vuhainguyen:
CDF-16299

Goes with https://github.com/cognitedata/cognite-sdk-scala/pull/559


src/main/scala/cognite/spark/v1/CdfSparkAuth.scala
```diff
@@ -16,17 +17,27 @@ object CdfSparkAuth {
       IO(AuthProvider(auth))
   }
 
-  final case class OAuth2ClientCredentials(credentials: OAuth2.ClientCredentials) extends CdfSparkAuth {
+  final case class OAuth2ClientCredentials(credentials: OAuth2.ClientCredentials)(
+      implicit sttpBackend: SttpBackend[IO, Any],
```
wjoel, line 21: This doesn't seem right to me, but I don't have time to come up with a better solution at the moment. At least for the `IORuntime`, I think we only ever use the one in `object CdpConnector` so we can just import that directly.

wjoel (approved):

src/main/scala/cognite/spark/v1/CdfSparkAuth.scala
vuhainguyen: I removed the IORuntime and use the import directly to make it less redundant 😁

PR #629, opened 2022-08-17
Push down raw keys

silvavelosa:
use `retrieveByKey` when raw rowns are queried with a `key` filter


src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -127,19 +161,27 @@ class RawTableRelation(
...
+  // scalastyle:off cyclomatic.complexity
   override def buildScan(requiredColumns: Array[String], filters: Array[Filter]): RDD[Row] = {
     val (minLastUpdatedTime, maxLastUpdatedTime) = filtersToTimestampLimits(filters, "lastUpdatedTime")
     val filteredRequiredColumns = getJsonColumnNames(requiredColumns)
     val filteredSchemaFields = getJsonColumnNames(schema.fieldNames)
     val lengthOfRequiredColumnsAsString = requiredColumns.mkString(",").length
 
+    val keysFilter =
+      filters.flatMap {
+        case EqualTo("key", value) => Array(value.toString())
```
wjoel, line 180: I think we should handle at least `Or` as well, and `EqualNullSafe`. Check `SequenceRowsRelation.getSeqFilter`. Possibly `And` as well, where we know that there will be zero rows matching "key = 'foo' AND key = 'bar'".

We should also have tests that cover those different cases.

src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -102,13 +130,19 @@ class RawTableRelation(
...
+        val partitionCursors =
+          CdpConnector
+            .clientFromConfig(config)
+            .rawRows(database, table)
+            .getPartitionCursors(filter, configWithLimit.partitions)
+            .unsafeRunSync()
+            .toVector
+        getStreams(filter, partitionCursors)
+      }
+      case keys => getStreamsByKeys(keys.toArray)
```
wjoel, line 144: It should be safe to assume that when filtering on keys, the number of keys will be quite limited, so we can just retrieve them all in parallel (or possibly in parallel for smaller groups, but I think our sttp backend limits parallelism to something reasonable anyway).

We can even enforce this, throwing an error if the user requests more than 100 keys or something. Note that if a user does something like `select * from my.table where keys in (select key from other.table)`, it doesn't get pushed down by Spark anyway - keys must be specified explicitly, which is why it should be feasible to assume that the number of keys will be small, and to throw an error if not.

src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -76,6 +77,32 @@ class RawTableRelation(
     cursors.map(rawClient.filterOnePartition(filter, _, limit))
   }
 
+  def pullFromKeysSeq(
+      keys: List[String],
+      get: String => IO[RawRow]
+  ): Pull[IO, RawRow, Unit] =
+    keys match {
```
wjoel, line 84: Very cool! But, I think we can do something like `Stream.emit(keys).parEvalMap(getRowByKey)`. See comment below.

src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -76,6 +77,32 @@ class RawTableRelation(
...
+      case key :: tail =>
+        Pull.eval(get(key)).flatMap { row =>
+          Pull.output(Chunk.singleton(row)) >>
+            pullFromKeysSeq(tail, get)
+        }
+    }
+
+  def getStreamsByKeys(keys: Array[String])(
+      client: GenericClient[IO],
+      limit: Option[Int],
```
wjoel, line 95: If we assume the number of keys will be relatively small, and limited to be small by us, we can safely ignore both `limit: Option[Int]` and `numPartitions`, I think. Maybe use `limit: Option[Int]` on the set of keys, somewhere, but it's more of a hint anyway so perhaps we don't even need to bother with that.

wjoel (changes requested):
Really needs tests. ;)

src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -127,19 +161,27 @@ class RawTableRelation(
...
     val (minLastUpdatedTime, maxLastUpdatedTime) = filtersToTimestampLimits(filters, "lastUpdatedTime")
     val filteredRequiredColumns = getJsonColumnNames(requiredColumns)
     val filteredSchemaFields = getJsonColumnNames(schema.fieldNames)
     val lengthOfRequiredColumnsAsString = requiredColumns.mkString(",").length
 
+    val keysFilter =
+      filters.flatMap {
+        case EqualTo("key", value) => Array(value.toString())
+        case In("key", values) => values.map(_.toString())
+        case _ => Array.empty[String]
```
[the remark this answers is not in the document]
silvavelosa: it was needed before, but no longer, no idea why 😅

src/main/scala/cognite/spark/v1/RawTableRelation.scala
silvavelosa: limit was already ignored... now `numPartitions` is used for the maximum parallelism of parEvalMap, we could put a reasonable cap (or fixed parallelism instead) of 4 - 10 if we are worried about users setting a high partition count

src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -157,7 +199,7 @@ class RawTableRelation(
     }
 
     val rdd =
-      readRows(config.limitPerPartition, None, rawRowFilter, jsonSchema)
+      readRows(config.limitPerPartition, None, rawRowFilter, keysFilter, jsonSchema)
```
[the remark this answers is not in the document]
silvavelosa: `RawRowFilter` comes from the sdk, it's the filter that is passed to the list rows endpoint.... We could make it `Either[RawRowFilter, Seq[String]]`, but since `Either` usually means error or success, I'm not sure if it would lead to confusion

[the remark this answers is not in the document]
wjoel: Ah, I see, didn't realize it was in the SDK. Maybe call it something other than "filter" if it's not a `Filter`-like class, then? `requestedKeys`? They are not "required" as it's a valid query to ask for `key in ("foo", "bar")` even though "bar" doesn't exist (as is made clear by ignoring the 404 response above).

src/main/scala/cognite/spark/v1/RawTableRelation.scala
silvavelosa: I think throwing an error might be too blocking, and I've seen at least one query with over 100 hardcoded cases (i don't remember if it was a raw query)... maybe just ignoring the `key` filters and going back to the current behavior (with last updated time and columns filters only) would be better if too many keys are provided.

src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -127,19 +161,27 @@ class RawTableRelation(
...
       (r: RawRow) => r.key,
-      getStreams(filter, partitionCursors),
+      streams,
       deduplicateRows = true // if false we might end up with 429 when trying to update assets with multiple same request
     )
   }
 
   override def buildScan(): RDD[Row] = buildScan(schema.fieldNames, Array.empty)
 
+  // scalastyle:off cyclomatic.complexity
```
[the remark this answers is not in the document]
silvavelosa: yes! 🚀

wjoel:
@silvavelosa I kind of went down a rabbit hole on this one... I think we may need to use `Option[Set[String]]` to distinguish "no keys requested" (`None`) from "no keys can match the filter" (`Some(Set.empty)`), to then be able to AND the filters together correctly.

As for the number of RDD partitions, we only need one if individual keys are requested, at least if we limit it to a small(ish) number before falling back to the previous behavior, and our CdpConnector client has its own limits on the number of concurrent requests we allow, so we can simply use `parEvalMapUnbounded`.

It ends up looking something like this: https://github.com/cognitedata/cdp-spark-datasource/pull/632, but I've probably made some mistakes. I think we should use a larger table and come up with some more difficult test cases.

wjoel (approved):
LGTM, nice! Don't forget to update CHANGELOG to let the world know about this glorious improvement.

PR #630, opened 2022-08-18
Enable sbt-tpolecat, fix compiler warnings

wjoel:


src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -40,18 +40,35 @@ final case class RelationConfig(
   def sparkPartitions: Int = Math.max(1, partitions / parallelismPerPartition)
 }
 
-object OnConflict extends Enumeration {
-  type Mode = Value
-  val Abort, Update, Upsert, Delete = Value
-
-  def withNameOpt(s: String): Option[Value] = values.find(_.toString.toLowerCase == s.toLowerCase())
+sealed trait OnConflictOption extends Serializable
```
wjoel, line 43: Scala 2.13 doesn't like exhaustiveness checking on `Enumeration`, and they're a bit awkward anyways, so using a sealed trait with objects instead.

src/main/scala/cognite/spark/v1/NumericDataPointsRdd.scala
```diff
@@ -185,9 +185,12 @@ final case class NumericDataPointsRdd(
       id: CogniteId,
       start: Instant,
       end: Instant,
-      granularities: Seq[Granularity] = granularitiesToTry): IO[Option[Seq[Range]]] =
+      granularities: List[Granularity] = granularitiesToTry): IO[Option[Seq[Range]]] =
     granularities match {
-      case granularity +: moreGranular =>
+      case Nil =>
+        throw new RuntimeException(
+          "No granularities requested for data points aggregation. Please report this to support@cognite.com")
```
vuhainguyen, line 192: I wonder if we haven't run into an issue when granularities before 😱 , anyway this would help to trace back the error easily 😁

src/main/scala/cognite/spark/v1/DefaultSource.scala
vuhainguyen: I was about to ask =))

vuhainguyen (approved):
🚀

src/main/scala/cognite/spark/v1/NumericDataPointsRdd.scala
wjoel: We're the only callers of this (private) method, so it is (or should be) impossible. Arguably we should have used a `NonEmptyList` to have the compiler verify it for us, but this is a simple fix to tell the compiler that we think we know what we're doing, and to tell users to inform us if it turns out we do not know what we're doing. :)

PR #634, opened 2022-09-13
Add support unique externalId per space for DMS

vuhainguyen:
CDF-16421


src/main/scala/cognite/spark/v1/AlphaDataModelInstancesRelation.scala
```diff
@@ -238,12 +238,20 @@ class AlphaDataModelInstanceRelation(
       if (andFilters.isEmpty) EmptyFilter else DSLAndFilter(andFilters)
     }
 
+    val instanceSpaceExternalIdFilter = filters
```
wjoel, line 241: I don't really understand what we're doing here or why. Could you add a comment explaining it? Why do we pick the first one, shouldn't it be an error to specify more than one, then?

src/main/scala/cognite/spark/v1/AlphaDataModelInstancesHelper.scala
```diff
@@ -347,5 +351,24 @@ object AlphaDataModelInstancesHelper {
...
         throw SparkSchemaHelperRuntime.badRowError(row, keyString, "String", "")
     }
+
+  def getDirectRelationIdentifierProperty(
+      row: Row,
+      keyString: String,
+      index: Int): DirectRelationIdentifier =
+    row.get(index) match {
+      case identifier: String =>
+        // Colon character cannot be used for space but it can be used for the externalId, so we should consider only
```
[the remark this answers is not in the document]
vuhainguyen: Alex said that[ they only have the restriction on the projects and spaces only,](https://cognitedata.slack.com/archives/C031G8Y19HP/p1663073024902339?thread_ts=1663071840.659449&cid=C031G8Y19HP) we can try to ask for the ban of `:` but I think as long as they allow it, we have to support it this way 😅

src/main/scala/cognite/spark/v1/AlphaDataModelInstancesHelper.scala
```diff
@@ -347,5 +351,24 @@ object AlphaDataModelInstancesHelper {
...
+        // the first ":" to split between spaceExternalId and externalId
+        identifier.split(":", 2) match {
+          case Array(spaceExternalId, externalId) =>
+            DirectRelationIdentifier(Some(spaceExternalId), externalId)
+          case _ =>
+            throw new CdfSparkException(
+              s"Can't upsert data model because `$keyString` should be a DirectRelationIdentifier of format `spaceExternalId:externalId`")
```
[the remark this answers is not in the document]
vuhainguyen: 👍 I updated the error message, and agree that `DirectRelationIdentifier` might be too technical for them, also add another check for the error message in the test

src/main/scala/cognite/spark/v1/AlphaDataModelInstancesRelation.scala
vuhainguyen: We decided during the collab session that we just go for a simple case, that is taking in account only the first `spaceExternalId` provided and use the space of the model in case nothing is provded, and ignore the rest hehe. I added a comment for it, we can throw an error later if we really have to 😁

src/main/scala/cognite/spark/v1/AlphaDataModelInstancesHelper.scala
```diff
@@ -347,5 +351,31 @@ object AlphaDataModelInstancesHelper {
...
     }
+
+  def getDirectRelationIdentifierProperty(
+      externalId: String,
+      row: Row,
+      keyString: String,
+      index: Int): DirectRelationIdentifier =
+    row.get(index) match {
+      case identifier: String =>
+        // Colon character cannot be used for space but it can be used for the externalId, so we should consider only
```
wjoel, line 362: We should have a test for that case, then. I still don't like that it's allowed, though...

src/main/scala/cognite/spark/v1/AlphaDataModelInstancesHelper.scala
```diff
@@ -347,5 +351,31 @@ object AlphaDataModelInstancesHelper {
...
+          case Array(spaceExternalId, externalId) =>
+            DirectRelationIdentifier(Some(spaceExternalId), externalId)
+          case _ =>
+            val errorHint = if (identifier.isBlank) {
+              ""
+            } else {
+              s", but was: $identifier"
+            }
+            throw new CdfSparkException(
+              s"Invalid data model instance row with external id '$externalId'. The values " +
```
wjoel, line 374: Should avoid using `+` for string concatenation, see https://nrinaudo.github.io/scala-best-practices/tricky_behaviours/string_concatenation.html

src/main/scala/cognite/spark/v1/AlphaDataModelInstancesRelation.scala
wjoel: Alright.

wjoel (approved):

src/main/scala/cognite/spark/v1/AlphaDataModelInstancesHelper.scala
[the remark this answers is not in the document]
vuhainguyen: According to Thorkil the colon is used way too often in customer systems so we can't forbid it 😅  https://cognitedata.slack.com/archives/C031G8Y19HP/p1663660773350059?thread_ts=1663656453.662199&cid=C031G8Y19HP

PR #635, opened 2022-09-19
Stringify direct relation

emelvarol-cognite:


emelvarol-cognite:
we agreed this was not necessary but good to have, I may have caused the original PR's test to fail while running this concurrently with it. I'll stop working on this for now.

PR #637, opened 2022-09-20
Fix direct relation validation

emelvarol-cognite:
This PR makes `array("my_space_external_id", null) as prop_direct_relation` effectless, meaning it writes null to that property. 
- For the other cases: The only valid direct relation value structure is `array("my_space_external_id", "my_external_id")`. The rest will fail with `CdfSparkException`.


src/main/scala/cognite/spark/v1/AlphaDataModelInstancesHelper.scala
```diff
@@ -114,34 +114,39 @@ object AlphaDataModelInstancesHelper {
...
     case x: ZonedDateTime =>
-      PropertyType.Timestamp.Property(x)
+      Some(PropertyType.Timestamp.Property(x))
     case a =>
       throw new CdfSparkException(
         notValidPropertyTypeMessage(a, PropertyType.Timestamp.code, Some("timestamp")))
   }
 
-  private def toDirectRelationProperty: Any => DataModelProperty[_] = {
+  private def toDirectRelationProperty: Any => Option[DataModelProperty[_]] = {
```
emelvarol-cognite, line 141: @vuhainguyen so this makes `array("something", null)` effectless, meaning it writes null to that property. and the cases other than `array("space", "ext_id")` will fail.

emelvarol-cognite: need to test though, but this should solve what you described.

vuhainguyen (approved):
😍 the tests. LGTM, please bump the version in the sbt as well

src/main/scala/cognite/spark/v1/AlphaDataModelInstancesHelper.scala
emelvarol-cognite: ok

PR #638, opened 2022-09-20
Rename alphadatamodels to datamodels 🚀 

silvavelosa:
CDF-16538


vuhainguyen:
Awesome 🚀 , please try to generate a lib in local and run with jetfire-backend test to see if it still works with the new renaming.
Probably you need to fix the tests as well, here for expl 👉 https://github.com/cognitedata/cdp-spark-datasource/blob/5b851cca6dd652b8f44918de5fd199d75643f5b6/src/test/scala/cognite/spark/v1/AlphaDataModelInstancesRelationTest.scala#L273

vuhainguyen (approved):
Very nice thanks Jaime 😁

PR #642, opened 2022-09-27
add support for asset subtree ids option filter

silvavelosa:
add a new option on `AssetsRelation` to filter by subtree ids


src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -126,6 +126,7 @@ class DefaultSource
       schema: StructType): BaseRelation = {
 
     val resourceType = parameters.getOrElse("type", sys.error("Resource type must be specified"))
+    val assetSubtreeIds = parameters.get("assetSubtreeIds").map(_.split(",").toList)
```
silvavelosa, line 129: not completely sure about this 🤔, can `externalId` have ',' in it?

emelvarol-cognite: hmm good question, i don't know :'( we can try creating some in transformations with comma. If so, would you consider only allowing a single root in the filtering?

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -14,8 +14,10 @@ import org.apache.spark.sql.types._
 import org.apache.spark.sql.{Row, SQLContext}
 
 import java.time.Instant
+import scala.util.Try
 
-class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
+class AssetsRelation(config: RelationConfig, subtreeIds: Option[List[String]] = None)(
```
emelvarol-cognite, line 19: my question here is: why do we take this from the options? We could add this to readSchema maybe and find in `sparkFilters`? I cannot think a nice way to reflect this in the UI... Because the options are most of the time write oriented in there 🤔

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
```diff
@@ -298,6 +298,52 @@ class AssetsRelationTest extends FlatSpec with Matchers with ParallelTestExecuti
...
   }
 
+  it should "support option filter assetSubtreeIds" taggedAs ReadTest in {
+    val df = spark.read
+      .format("cognite.spark.v1")
+      .option("apiKey", readApiKey)
+      .option("type", "assets")
+      .option("limitPerPartition", "1000")
+      .option("partitions", "1")
+      .option("assetSubtreeIds", "2161493773812721")
```
emelvarol-cognite, line 308: are the assets that assumed to be existing created in tbe beginning of test?

src/main/scala/cognite/spark/v1/DefaultSource.scala
vuhainguyen: yes it can unfortunately, I just tried to create one and re-query it, and not only , but almost any special character is accepted
image

src/main/scala/cognite/spark/v1/AssetsRelation.scala
silvavelosa: The main reason is that it's not a data column, and it would give way to ugly mixes of filters (mainly with OR)
[Here ](https://cognitedata.atlassian.net/wiki/spaces/INT/pages/3721166999/Asset+subtrees+retrieval+on+transformations )are the different alternatives considered, what you describe would be Option 1, what is being implemented corresponds to Option 3.

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
silvavelosa: those are assets in the magic `publicdata` project

src/main/scala/cognite/spark/v1/DefaultSource.scala
exyi: I don't think it matters, the user can always use the numeric InternalId to avoid this problem.

For that matter, an externalId with only digits is also valid, but you'd interpret it as internal id. Again, easy to workaround by using the actual internal id.

emelvarol-cognite: that is a really good point @silvavelosa but we need to give users a conversion function from id to externalId? as in dataset_id udf

wjoel: I don't know if there's any precedence of passing lists as options, but this is making me uncomfortable. Seems very unsafe, and like it will lead to surprising and confusing errors (when external ids contain `,`). If we have to string split, can we use something more unusual? `||` maybe? The customer in https://cognitedata.atlassian.net/wiki/spaces/INT/pages/3721166999/Asset+subtrees+retrieval+on+transformations would be happy with just one asset subtree, do we have to support multiple ones? Yes, we can, but can we save it for another day when/if a customer requests it, if they ever do?

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -44,6 +49,14 @@ class AssetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
             .transform))
   }
 
+  private val subtreeCogniteIds: Option[List[CogniteId]] =
```
wjoel, line 52: I see that we're doing this for sequence rows, but I think it's a bit terrible. I'd much rather have us be explicit as we are with "spaceExternalId" etc.

src/main/scala/cognite/spark/v1/DefaultSource.scala
wjoel: Or, for that matter, turn it into a JSON array `["\"asset1\",\"asset2\","3",3]`, and parse. Then we would even be able to tell "3" and 3 apart properly. Maybe options can have array values in Spark, but I seem to recall they're all just converted to strings, even if `.option("overwrite", true)` works...

silvavelosa: yep, that's how it went:

- integers like `123`, `456` are a single internalId
- valid json arrays of integers or strings are multiple internal/external ids
-- `["vnakls","2314","vnkalvs"]` are 3 externalIds
-- `[123, 2314, 123156]` are 3 internalIds
- anything else is a single externalId

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -472,6 +475,33 @@ object DefaultSource {
     )
   }
 
+  private[v1] def parseCogniteIds(paramName: String)(
+      parameters: Map[String, String]): Option[List[CogniteId]] = {
```
TharinduDG, line 479: Any specific reason why `paramName` & `parameters` are in 2 parameter lists? 🤔

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -472,6 +475,33 @@ object DefaultSource {
...
+          .validate(_.value.isNumber, "strict int decoder") // to distinguish 123 from "123"
+          .map(internalId => CogniteInternalId(internalId))
+          .widen,
+        Decoder.decodeString.map(externalId => CogniteExternalId(externalId)).widen
+      ).reduceLeft(_.or(_))
+    implicit val multipleIdsDecoder: Decoder[List[CogniteId]] =
+      List(
+        singleIdDecoder.map(List(_)),
+        Decoder.decodeArray[CogniteId].map(_.toList)
+      ).reduceLeft(_.or(_))
```
vuhainguyen, line 493: neat 👍

vuhainguyen, line 493: neat 👍

src/main/scala/cognite/spark/v1/DefaultSource.scala
silvavelosa: not really 🤔 .... it started as a function receiving the map and with hardcoded key, and then added the `paramName`.... but now that you mention it, probably it'll be easier to read if it receives just the string 😅

src/main/scala/cognite/spark/v1/DefaultSource.scala
TharinduDG: I don't think `.widen` is necessary since there is no covariance applicable here. 🤔

src/test/scala/cognite/spark/v1/DefaultSourceTest.scala
```diff
@@ -95,5 +97,26 @@ class DefaultSourceTest extends WordSpec with Matchers {
...
+        DefaultSource.parseCogniteIds(""" "\"123\"" """) shouldBe List(CogniteExternalId("\"123\""))
+        DefaultSource.parseCogniteIds("abc") shouldBe List(CogniteExternalId("abc"))
+        DefaultSource.parseCogniteIds("{invalid json") shouldBe List(CogniteExternalId("{invalid json"))
+        DefaultSource.parseCogniteIds("\"[123,456]\"") shouldBe List(CogniteExternalId("[123,456]"))
+      }
+      "parse arrays according to their contents" in {
+        DefaultSource.parseCogniteIds("""[123, "123"]""") shouldBe List(CogniteInternalId(123), CogniteExternalId("123"))
+      }
+      "parse other valid json as externalId" in {
+        DefaultSource.parseCogniteIds("""{"key": 123}""") shouldBe List(CogniteExternalId("""{"key": 123}"""))
```
emelvarol-cognite, line 116: i thought we wanted to extract `123` from this 🤔

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
```diff
@@ -298,6 +298,52 @@ class AssetsRelationTest extends FlatSpec with Matchers with ParallelTestExecuti
...
+    val metricsPrefix = s"pushdown.assets.name.${shortRandomString()}"
+    val df = spark.read
+      .format("cognite.spark.v1")
+      .option("apiKey", readApiKey)
+      .option("type", "assets")
+      .option("collectMetrics", "true")
+      .option("metricsPrefix", metricsPrefix)
+      .option("limitPerPartition", "1000")
+      .option("partitions", "1")
+      .option("assetSubtreeIds", """["WMT:23-YT-96105-01","WMT:23-TE-96137-02"]""")
```
emelvarol-cognite, line 337: can you also add a test like:  """[12424,"WMT:23-TE-96137-02"]""" ?

emelvarol-cognite (approved):
coool

silvavelosa:
Thanks for the reviews! 🚀

PR #643, opened 2022-10-21
Shrink partitions if configured

TharinduDG:
- reduce the partitions based on `partitionShrinkPercentage` config


vuhainguyen (commented):
LGTM, there's a compilation error and might be test to be fixed, removing SNAPSHOT if you want to publish new version and updating the CHANGELOG file to mention this change as well 😁

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -274,9 +275,13 @@ class DefaultSource
...
         if (originalNumberOfPartitions > 50 && originalNumberOfPartitions > idealNumberOfPartitions) {
-          (data.repartition(idealNumberOfPartitions), idealNumberOfPartitions)
+          val partitionsAfterShrink =
+            (idealNumberOfPartitions * (100 - config.partitionShrinkPercentage) / 100).toInt
+          (data.repartition(partitionsAfterShrink), partitionsAfterShrink)
         } else {
-          (data, originalNumberOfPartitions)
+          val partitionsAfterShrink =
+            (originalNumberOfPartitions * (100 - config.partitionShrinkPercentage) / 100).toInt
+          (data.repartition(partitionsAfterShrink), partitionsAfterShrink)
```
exyi, line 284: I think you should avoid calling repartition unless it's needed. AFAIK Spark will always do the data shuffling, even when the number of partitions remains exacrly the same.

Anyway, why do you need this option? Can't you just decrease number of `sparkPartitions`? (by increasing parallelismPerPartition, iirc)

TharinduDG:
This is no longer required.

PR #647, opened 2022-11-22
Add support for property Json in FDM

vuhainguyen:
CDF-17129

Test is added in https://github.com/cognitedata/cognite-sdk-scala/pull/577 but we don't need a new version of SDK since we keep treating Json as a String anyway.
Also doing migration to use OIDC instead of READ_API_KEY for public data.
More ref from https://hub.cognite.com/open-industrial-data-211/openid-connect-on-open-industrial-data-993

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -33,7 +33,9 @@ jobs:
       - name: Run tests
         env:
           TEST_API_KEY_WRITE: ${{ secrets.TEST_API_KEY_WRITE }}
-          TEST_API_KEY_READ: ${{ secrets.TEST_API_KEY_READ }}
+          TEST_OIDC_READ_CLIENT_ID: ${{ secrets.TEST_OIDC_READ_CLIENT_ID }}
+          TEST_OIDC_READ_CLIENT_SECRET: ${{ secrets.TEST_OIDC_READ_CLIENT_SECRET }}
+          TEST_OIDC_READ_TENANT: ${{ secrets.TEST_OIDC_READ_TENANT }}
           TEST_CLIENT_ID_BLUEFIELD: ${{ secrets.BLUEFIELD_CLIENT_ID }}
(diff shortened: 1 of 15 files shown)


src/main/scala/cognite/spark/v1/udf/CogniteUdfs.scala
```diff
@@ -66,21 +66,21 @@ object CogniteUdfs {
...
 
-    if (res.status.isDefined && res.status.get == "Completed") {
+    if (res.status == "Completed") {
       for {
         functionCallResponse <- client
           .functionCalls(functionId)
-          .retrieveResponse(res.id.getOrElse(0L))
-      } yield functionCallResponse.response.getOrElse(Json.fromJsonObject(JsonObject.empty))
+          .retrieveResponse(res.id)
+      } yield functionCallResponse.response
```
vuhainguyen, line 83: These changes is due to https://github.com/cognitedata/cognite-sdk-scala/pull/574

.github/workflows/main.yaml
```diff
@@ -33,7 +33,9 @@ jobs:
       - name: Run tests
         env:
           TEST_API_KEY_WRITE: ${{ secrets.TEST_API_KEY_WRITE }}
-          TEST_API_KEY_READ: ${{ secrets.TEST_API_KEY_READ }}
+          TEST_OIDC_READ_CLIENT_ID: ${{ secrets.TEST_OIDC_READ_CLIENT_ID }}
+          TEST_OIDC_READ_CLIENT_SECRET: ${{ secrets.TEST_OIDC_READ_CLIENT_SECRET }}
+          TEST_OIDC_READ_TENANT: ${{ secrets.TEST_OIDC_READ_TENANT }}
```
emelvarol-cognite, line 38: Isn't this the same as bluefield oidc credentials? but maybe it is good to have redundant env vars due to naming.

emelvarol-cognite (approved):
Looks good to me

PR #649, opened 2022-12-07
Reduce maxRetry on for backend that is used for auth

vuhainguyen:
Currently we use the `maxRetries` parsing from Config for getting auth, but we should give up faster in case of auth failure instead of retrying that makes the job longer and then taking over resource of the cluster.
More context in 👇 
https://cognitedata.slack.com/archives/C93QJMZE0/p1670415341226499
https://cognitedata.slack.com/archives/CAYSVGPUN/p1670416079808469


TharinduDG (approved):
👍

PR #650, opened 2022-12-09
Feat/custom type node ref

vuhainguyen:
CDF-17207


src/test/scala/cognite/spark/v1/DataModelInstancesRelationTest.scala
```diff
@@ -1115,4 +1142,93 @@ class DataModelInstancesRelationTest
...
+    ex.getMessage shouldBe "Direct relation identifier should be named_struct(\"spaceExternalId\", yourSpace, \"externalId\", yourExternalId) or " +
+      "named_struct(\"externalId\", yourExternalId) but got notExternalId as the schema."
+
+    val ex1 = sparkIntercept {
+      insertRows(
+        primitiveExtId2,
+        spark
+          .sql(s"""
+                  |select named_struct("spaceExternalId", "vu", "notExternalId", "hai") as prop_direct_relation,
+                  |'hello_my_name_is_emel' as externalId""".stripMargin)
```
hakontro, line 1213:
```suggestion
                  |'hello_my_name_is_håkon' as externalId""".stripMargin)
```

vvemel: haha :D

src/test/scala/cognite/spark/v1/DataModelInstancesRelationTest.scala
```diff
@@ -1080,9 +1080,36 @@ class DataModelInstancesRelationTest
         props.get("prop_direct_relation").map(_.value) shouldBe None
       }
     )
+
+    tryTestAndCleanUp(
+      Seq(randomId), {
+        retryWhile[Boolean](
```
TharinduDG, line 1086: I don't understand the logic behind this retry function.
If there's an issue, it should fail fast and we'll know there's an issue.
I think this retry just hide errors.

TharinduDG (approved):
LGTM except the `retryWhile` 😉

src/test/scala/cognite/spark/v1/DataModelInstancesRelationTest.scala
vuhainguyen: it helps to retry (max 20  times with backoff delay) and is used in many places of Spark DS
https://github.com/search?l=Scala&q=org%3Acognitedata+retryWhile&type=Code
When the api is not stable enough I think it's just fine to retry it anyway, we can reduce the number of retries though

PR #651, opened 2022-12-09
Chore/use separate backend auth

vuhainguyen:
Similar as https://github.com/cognitedata/cdp-spark-datasource/pull/649 but in different place


silvavelosa (approved):
LGTM ✔️ 

As a side note... I'm pretty sure we can make a small change in the sdk to avoid that unsafeRunSync

vuhainguyen:
> LGTM ✔️
> 
> As a side note... I'm pretty sure we can make a small change in the sdk to avoid that unsafeRunSync

Yep feel free to make a PR for proposal hehe 🚀

PR #658, opened 2023-01-17
Migrate APIKey to OIDC in tests

vuhainguyen:
CDF-14847


src/test/scala/cognite/spark/v1/FunctionsUDFTest.scala
```diff
@@ -1,52 +1,52 @@
...
-      .collect()
-
-    forAll(res) { row: Row =>
-      row.getString(0) shouldEqual "42"
-    }
-
-    assert(res.length == 1)
-  }
-}
+//package cognite.spark.v1
```
vuhainguyen, line 1: commented since the test is already ignore, we can make it work again in the future, but need to switch CogniteUdfs to use OIDC instead of API Key

TharinduDG (approved):
👍

PR #662, opened 2023-02-03
Multiple filter reqeusts for usedFor All

TharinduDG:
- Use multiple filter requests based on View's `usedFor`
- organising tests to avoid accidentally deleting unrelated edge dependencies
- switching to default client instead of alpha

### DIFF
--- a/src/main/scala/cognite/spark/v1/DefaultSource.scala
+++ b/src/main/scala/cognite/spark/v1/DefaultSource.scala
@@ -140,7 +140,7 @@ class DefaultSource
       throw new CdfSparkException("'instanceSpaceExternalId' should be specified"))
 
     new FlexibleDataModelsRelation(
-      config.copy(clientTag = Some("alpha")),
+      config,
       viewSpaceExternalId = viewSpaceExternalId,
       viewExternalId = viewExternalId,
       viewVersion = viewVersion,
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelsRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelsRelation.scala
@@ -389,23 +389,11 @@ class FlexibleDataModelsRelation(
       selectedColumns
     }
 
-    val instanceType = {
-      (viewDefinition.usedFor match {
-        case Usage.Node => Some(InstanceType.Node)
-        case Usage.Edge => Some(InstanceType.Edge)
-        case Usage.All => None
-      }).orElse {
-        filters.collectFirst {
-          case EqualTo("instanceType", value) =>
-            Decoder[InstanceType]
-              .decodeJson(Json.fromString(String.valueOf(value).toLowerCase(Locale.US)))
-              .toOption
-        }.flatten
-      }
-    }
+    val instanceType = getInstanceType(filters)
 
     val instanceFilter = filters.toVector.traverse(toInstanceFilter) match {
       case Right(fs) if fs.isEmpty => None
+      case Right(fs) if fs.length == 1 => fs.headOption
       case Right(fs) => Some(FilterDefinition.And(fs))
       case Left(err) => throw err
     }
@@ -531,6 +519,20 @@ class FlexibleDataModelsRelation(
       nullable
     )
 
+  private def getInstanceType(filters: Array[Filter]) =
+    (viewDefinition.usedFor match {
+      case Usage.Node => Some(InstanceType.Node)
+      case Usage.Edge => Some(InstanceType.Edge)
+      case Usage.All => None
+    }).orElse {
+      filters.collectFirst {
+        case EqualTo(attribute, value) if attribute.equalsIgnoreCase("instanceType") =>
+          Decoder[InstanceType]
+            .decodeJson(Json.fromString(String.valueOf(value).toLowerCase(Locale.US)))
+            .toOption
+      }.flatten
+    }
+
   private def toProjectedInstance(
       i: InstanceDefinition,
       selectedInstanceProps: Array[String]): ProjectedFlexibleDataModelInstance = {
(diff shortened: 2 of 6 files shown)


src/main/scala/cognite/spark/v1/FlexibleDataModelsRelation.scala
```diff
@@ -386,41 +390,20 @@ class FlexibleDataModelsRelation(
...
-      }
-    }
+    val instanceType = getInstanceType(filters)
 
     val instanceFilter = filters.toVector.traverse(toInstanceFilter) match {
       case Right(fs) if fs.isEmpty => None
+      case Right(fs) if fs.length == 1 => fs.headOption
       case Right(fs) => Some(FilterDefinition.And(fs))
       case Left(err) => throw err
     }
```
TharinduDG, line 400: No need of using `And` if there's no more than 1 filters

src/test/scala/cognite/spark/v1/FlexibleDataModelsRelationTest.scala
```diff
@@ -453,27 +529,89 @@ class FlexibleDataModelsRelationTest extends FlatSpec with Matchers with SparkTe
...
+      "doubleListProp2" -> FDMContainerPropertyTypes.Float64ListWithoutDefaultValueNullable,
+      "boolListProp1" -> FDMContainerPropertyTypes.BooleanListWithoutDefaultValueNonNullable,
+      "boolListProp2" -> FDMContainerPropertyTypes.BooleanListWithoutDefaultValueNullable,
+      "dateListProp1" -> FDMContainerPropertyTypes.DateListWithoutDefaultValueNonNullable,
+      "dateListProp2" -> FDMContainerPropertyTypes.DateListWithoutDefaultValueNullable,
+      "timestampListProp1" -> FDMContainerPropertyTypes.TimestampListWithoutDefaultValueNonNullable,
+      "timestampListProp2" -> FDMContainerPropertyTypes.TimestampListWithoutDefaultValueNullable,
+      "jsonListProp1" -> FDMContainerPropertyTypes.JsonListWithoutDefaultValueNonNullable,
+      "jsonListProp2" -> FDMContainerPropertyTypes.JsonListWithoutDefaultValueNullable,
+    )
```
TharinduDG, line 595: Testing with all possible property types to avoid surprises like fdm v2 casting issues.

src/main/scala/cognite/spark/v1/FlexibleDataModelsRelation.scala
```diff
@@ -203,7 +203,10 @@ class FlexibleDataModelsRelation(
     if (config.collectMetrics) {
       itemsRead.inc()
     }
-    new GenericRow(a.properties)
+    new GenericRow(a.properties.map {
+      case a: Array[Any] => new GenericRow(a)
+      case e => e
```
vuhainguyen, line 208: It was not clear for me why do we need this?
Can you add a comment here to explain it? If waiting for the build again is too much we can "by pass" with a comment here instead but this change is weird for me

src/main/scala/cognite/spark/v1/FlexibleDataModelsRelation.scala
```diff
@@ -537,27 +519,107 @@ class FlexibleDataModelsRelation(
...
+            includeTyping = Some(includeTyping)
+          ),
+          InstanceFilterRequest(
+            instanceType = Some(InstanceType.Edge),
+            filter = instanceFilter,
+            sort = None,
+            limit = limit,
+            cursor = None,
+            sources = Some(sources),
+            includeTyping = Some(includeTyping)
```
vuhainguyen, line 617: This is a bit boiler plate, can just create a common filterRequest and then use `filterRequest.copy(instanceType= ...) `for this  and we can see directly what make the difference instead of scrolling on those duplicate line 😄

src/test/scala/cognite/spark/v1/FlexibleDataModelsRelationTest.scala
```diff
@@ -350,16 +430,8 @@ class FlexibleDataModelsRelationTest extends FlatSpec with Matchers with SparkTe
...
                  |where
-                 |forEqualsFilter = 'str1' and
-                 |forInFilter in ('str1', 'str2', 'str3') and
-                 |forGteFilter >= 1 and
-                 |forGtFilter > 1 and
-                 |forLteFilter <= 2 and
-                 |forLtFilter < 4 and
-                 |(forOrFilter1 == 5.1 or forOrFilter2 == 6.1) and
-                 |forIsNotNullFilter is not null and
                  |forIsNullFilter is null
```
vuhainguyen, line 435: I might miss it but do we have a replacement somewhere to test other filters instead of only is null? The old test case looks better with different filter condition and values (even though test is still ignore though)

src/main/scala/cognite/spark/v1/FlexibleDataModelsRelation.scala
TharinduDG: its not bypassing.
`startNode`, `endNode` & `type` are Structs. So they are Arrays of values, which needs to be a GenericRow it self.
This function is called `toRow` and this is how it supposed to create `Row`s.
So there's nothing special here I guess. 🤔

src/test/scala/cognite/spark/v1/FlexibleDataModelsRelationTest.scala
TharinduDG: This test is ignored. 
FDM still haven't released the fix.

vuhainguyen (approved):

src/test/scala/cognite/spark/v1/FlexibleDataModelsRelationTest.scala
vuhainguyen: Yes but later please add back some tests for better filter condition rather than just isNull :)

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -140,7 +140,7 @@ class DefaultSource
       throw new CdfSparkException("'instanceSpaceExternalId' should be specified"))
 
     new FlexibleDataModelsRelation(
-      config.copy(clientTag = Some("alpha")),
+      config,
```
TharinduDG, line 143: we can switch to the default

PR #663, opened 2023-02-06
Add well data layer

sighol:
The approach we've used is similar to the one used by FDM.
In the well data layer (WDL) we have an endpoint that returns a spark-compatible struct type for all our models.
The CDP spark data source uses this endpoint to figure out the expected format for the rows it reads and writes to WDL.

The data source for WDL is implemented like this
```scala
val nptEvents = spark.read
  .format("cognite.spark.v1")
  .option("project", "jetfiretest2")
  .option("type", "welldatalayer") // main type is `welldatalayer`.
  .option("wdlDataType", "Npt") // `wdlDataType` is a required extra option.
  .load()
```


src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -121,6 +122,15 @@ class DefaultSource
     )(sqlContext)
   }
 
+  private def createWellDataLayer(
+      parameters: Map[String, String],
+      config: RelationConfig,
+      sqlContext: SQLContext
+  ): WellDataLayerRelation = {
+    val model = parameters.getOrElse("wdlDataType", sys.error("wdlDataType must be specified"))
```
vuhainguyen, line 130: prior to use `CdfSparkException("'viewSpaceExternalId' id of the 'space' must be specified")` rather than sys.error 😅

src/main/scala/cognite/spark/v1/wdl/WellDataLayerRDD.scala
```diff
@@ -0,0 +1,119 @@
...
+      case "RigOperation" => "rigoperations/list"
+      case "HoleSectionGroup" => "holesections/list"
+      case "WellTopGroup" => "welltops/list"
+      case "Npt" => "npt/list"
+      case "Nds" => "npt/list"
+      case "CasingSchematic" => "casings/list"
+      case "Trajectory" => "trajectories/list"
+      case _ => sys.error(s"Unknown model type: $modelType")
+    }
+  // scalastyle:on cyclomatic.complexity
```
vuhainguyen, line 41: Maybe use a local Map instead of this long list of pattern match to avoid cyclomatic.complexity?
Smthing like 
```scala
private val modelTypeToUrl = Map("Well" -> "wells/list",  "DepthMeasurement" -> "measurements/depth/list" , ...)
private def getReadUrlPart(modelType: String): String = {
    val key = modelType.replace("Ingestion", "")
    modelTypeToUrl.get(key).getOrElse(throw new CdfSparkException..)
}
```

src/main/scala/cognite/spark/v1/wdl/WellDataLayerRDD.scala
```diff
@@ -0,0 +1,119 @@
...
+    override def next(): Row = nextComputed match {
+      case Some((head, tail)) =>
+        nextComputed = computeNext(tail, isFirstQuery = false)
+        head
+      case None => throw new NoSuchElementException()
+    }
+  }
+
+  private def processItems(itemsWithCursor: ItemsWithCursor[JsonObject]) = {
+    val (headItem, tailItems) = itemsWithCursor.items.splitAt(1)
```
vuhainguyen, line 77: uhm why 1? 😅 Might need to wrap it with smthing or we can fall into an exception if the size <=1

src/main/scala/cognite/spark/v1/wdl/WellDataLayerRelation.scala
```diff
@@ -0,0 +1,83 @@
...
+      case "TimeMeasurement" => "measurements/time"
+      case "RigOperation" => "rigoperations"
+      case "HoleSectionGroup" => "holesections"
+      case "WellTopGroup" => "welltops"
+      case "Npt" => "npt"
+      case "Nds" => "npt"
+      case "CasingSchematic" => "casings"
+      case "Trajectory" => "trajectories"
+      case "Source" => "sources"
+      case _ => sys.error(s"Unknown model type: $modelType")
```
vuhainguyen, line 66: see comment above, may be use a local map instead of a long list of pattern matching 😁

src/test/scala/cognite/spark/v1/wdl/WDLTrajectoriesTest.scala
```diff
@@ -0,0 +1,181 @@
...
+    try {
+      testIngestionsDF.write
+        .format("cognite.spark.v1")
+        .option("project", "jetfiretest2")
+        .option("type", "welldatalayer")
+        .option("wdlDataType", "TrajectoryIngestion")
+        .useOIDCWrite
+        .save()
+    } catch {
+      case _: org.apache.spark.SparkException => ()
```
vuhainguyen, line 147: Do we expect this to throw a SparkException? If so you can just do this instead 😁 
`a[SparkException] should be thrownBy ...`

src/test/scala/cognite/spark/v1/wdl/WDLRDDCursorLimitTest.scala
```diff
@@ -0,0 +1,75 @@
...
+  private val testNptIngestionsDF = spark.read
+    .schema(testClient.getSchema("NptIngestion"))
+    .json("src/test/resources/wdl-test-ntp-ingestions.jsonl")
+
+  private val testNptsDF = spark.read
+    .schema(testClient.getSchema("Npt"))
+    .json("src/test/resources/wdl-test-ntps.jsonl")
+
+  override protected def beforeAll(): Unit = {
+    SQLConf.get.setConfString("spark.sql.legacy.respectNullabilityInTextDatasetConversion", "true")
```
vuhainguyen, line 33: Transformation runs with common Spark configs, do you need this change only for testing or also later when deploying Transformation with WDL, this can be breaking change and create issues for existed transformations 😅

src/main/scala/cognite/spark/v1/wdl/WellDataLayerRDD.scala
vvb16: Should be the same for modern compilers (I hope), but this old-school way is definitely good! :)

src/main/scala/cognite/spark/v1/wdl/WellDataLayerRDD.scala
vvb16: This way to taking one element from the head should work for any size of the items, I believe.
If source collection is empty - it will result in valid pair of two empty collections.

src/test/scala/cognite/spark/v1/wdl/WDLTrajectoriesTest.scala
vvb16: Great catch!
Leftover of debugging.

src/test/scala/cognite/spark/v1/wdl/WDLRDDCursorLimitTest.scala
vvb16: In this exact place it is only for testing.
But I believe this SQLConf is specific for Spark Session here and session is local to SparkTest parent of WDLRDDCursorLimitTest. So this config instance is specific to this test suite.

src/main/scala/cognite/spark/v1/DefaultSource.scala
vvb16: BTW, see lines 112 and 114 just above this function.
This is where we got it. ;)

src/test/scala/cognite/spark/v1/wdl/WDLRDDCursorLimitTest.scala
sighol: I've removed it now. In some cases, it added better error messages, but it isn't necessary and adds confusion imho.

vuhainguyen (approved):

PR #665, opened 2023-02-07
Remove first row schema based nullability check

TharinduDG:
- Checking nullability of `Row` fields based on the first `Row` schema seems to be not working in e2e tests
- This works fine in integrations tests

Screenshot 2023-02-07 at 19 51 04

The error `Property [Name] cannot contain null values` is not correct.
That data column is `null` free


SunilVenekarKrishnamoorthy (approved):
Removing this check seems okay 👍 
But I can't understand why did the e2e test fails :D

PR #667, opened 2023-02-08
Make instanceSpaceExternalId optional

TharinduDG:
`instanceSpaceExternalId` is only required for upserting data


vuhainguyen (approved):
Would be great if we can just merge the branch with https://github.com/cognitedata/cdp-spark-datasource/pull/664 so that we don't have to waste another build :)

PR #668, opened 2023-02-08
Parse any value to string & Handle java.sql Date, timestamps

TharinduDG:
- for string property types, accept any valid string convertible value
- handle java.sql.timestamps & java.sql.date types for timestamps and dates


vuhainguyen (approved):
LGTM, let try to remove the SNAPSHOT in the next bug fix :)

PR #672, opened 2023-02-17
CDF-17596 Support timestamps with negative epoch values

TharinduDG:
- setting the start filter to the minimum possible epoch value to support negative timestamps


SunilVenekarKrishnamoorthy (approved):
👍

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
```diff
@@ -181,7 +181,7 @@ object PushdownUtilities {
       // Note that this way of aggregating filters will not work with "Or" predicates.
       Try(timestampLimits.filter(_.isInstanceOf[Min]).max).toOption // scalafix:ok
         .map(_.value)
-        .getOrElse(Instant.ofEpochMilli(0)),
+        .getOrElse(Instant.ofEpochMilli(-2208988800000L)),
```
vuhainguyen, line 184: This constant can be imported from EpochMilli instead? 🤔

vuhainguyen (approved):
Nice 👍

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
vuhainguyen: Alright maybe add a comment here to give a hint why/how do you find that number 😄

PR #673, opened 2023-02-22
Supporting ConnectionDefinition Instances

TharinduDG:
- Split FDM relation in to `FlexibleDataModelConnectionRelation` and `FlexibleDataModelCorePropertyRelation`
- support filtering by Edge `type`, `space`, `startNode` & `endNode` and Node/Edge `space`
- For `FlexibleDataModelCorePropertyRelation`, `instanceSpace` is optional and therefore `space` will be populated from data if `instanceSpace` is not specified.
- Default to connection definition's `space` for `startNode` & `endNode` in connection instances.


src/test/scala/cognite/spark/v1/FlexibleDataModelConnectionRelationTest.scala
```diff
@@ -0,0 +1,198 @@
...
+      )
+      _ <- IO.sleep(2.seconds)
+    } yield connectionsWritten).unsafeRunSync()
+
+    val readConnectionsDf = readRows(edgeSpace = spaceExternalId, edgeExternalId = edgeTypeExtId)
+
+    readConnectionsDf.createTempView("connection_instances_table")
+
+    val selectedConnectionInstances = spark
+      .sql("select * from connection_instances_table")
```
vuhainguyen, line 79: I think we need to ingest different edge"types" so that we can have a test:
```sql
SELECT * FROM connection_instances_table WHERE type = named_struct"space", $typeSpace, "externalId", $typeExteralId) AND space = $edgeSpace
```
to make sure the filter work on different edge types and spaces 😁

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -15,58 +14,81 @@ import com.cognite.sdk.scala.v1.fdm.common.properties.PropertyType.{
...
-      instanceSpaceExternalId: String,
+      instanceSpace: Option[String],
       rows: Seq[Row],
       schema: StructType,
       propertyDefMap: Map[String, ViewPropertyDefinition],
       source: SourceReference
   ): Either[CdfSparkException, Vector[NodeOrEdgeCreate]] =
     rows.toVector.traverse { row =>
       for {
+        space <- instanceSpace.map(Right(_)).getOrElse(extractSpaceExternalId(schema, row))
```
TharinduDG, line 86: `instanceSpace` is optional for the `Relation` and therefore `space` will be populated from data.

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -15,58 +14,81 @@ import com.cognite.sdk.scala.v1.fdm.common.properties.PropertyType.{
...
+}
 import io.circe.syntax.EncoderOps
 import org.apache.spark.sql.Row
 import org.apache.spark.sql.types.StructType
 
 import java.time._
 import scala.util.{Failure, Success, Try}
 
-// scalastyle:off number.of.methods
+// scalastyle:off number.of.methods file.size.limit
```
vuhainguyen, line 37: may be this is a hint that we should split this file/object to smaller one 😄

vuhainguyen (approved):
LGTM
We should consider refactor FlexibleDataModelRelationUtils in the next PR, that object start to get too big 😄 
This contains a lot of change so I think it's better to bump the SNAP-SHOT version as well, if anything happens we can revert back the Spark Ds version

PR #674, opened 2023-02-28
Rename space to spaceExternalId when extracting DirectRelation

vuhainguyen:
In jetfire backend, the UDF relation_ref has parameter named `spaceExternalId`, and for backward compatibility and less refactoring, we should keep the parameter name as `spaceExternalId`.


TharinduDG:
This make no sense to me. In the last few meetings, we agreed to use `space` instead of `spaceExternalId`.
image

This should be handled in jetfire-backend not here.

vuhainguyen:
> This make no sense to me. In the last few meetings, we agreed to use `space` instead of `spaceExternalId`. image
> 
> Even if we want to do this, this should be handled in jetfire-backend not here.

As discussed this naming is not exposed to the user, if we want to keep this the jetfire-backend needs to have 2 versions to generate space and spaceExternalId, otherwise we have to refactor the code of v2 in Spark Data Source to use space instead of spaceExternalId this is quite a work and doesn't bring much value. Our users will use node_reference anyway, the parameter name under the hood doesn't matter for them

vuhainguyen:
|Compromising naming convention just to support v2 (which will be going away after v3) is not a worthy compromise for me.
I would trade off this for all the refactoring that need to be done in other to support both version. And the code for v2 will be removed anyway so it is a wasted of effort to refactor it just to have better naming but will be removed later.

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -225,7 +225,7 @@ object FlexibleDataModelRelationUtils {
       row: Row): Either[CdfSparkException, DirectRelationReference] =
     Try {
       val struct = row.getStruct(schema.fieldIndex(propertyName))
-      val space = Option(struct.getAs[Any]("space"))
+      val space = Option(struct.getAs[Any]("spaceExternalId"))
```
TharinduDG, line 228: I would consider handling both `space` & `spaceExternalId`.
If `space` is not defined try `spaceExternalId` as a fall back.
That make more sense than changing everything to `spaceExternalId`.
All the other places use `space` to be be consistent with the api spec.

silvavelosa (approved):
✔️ LGTM

The readable UDF name is the same for the current DMS Beta that is already in use and the new version, so we must keep it working for users already relying on it. 

Between changing 15 lines here, breaking the current functionality, or refactoring the current implementation, I think a slightly longer name isn't a big compromise.

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
vuhainguyen: I don't think it's a good idea to support both here, we can support space to keep the existing tests works for v3 with both syntax `spaceExternalId` and `space` in the named_struct, but it's useless since the UDF in jetfire-backend continues to generate `spaceExternalId` anyway. It is redundant and doesn't bring much value here. If we keep `spaceExternalId`, you can rename it to space when deprecating v2 in both Spark DS and jetfire. This is an internal naming in the code, we don't expose it to any user so we can live with the fact it's not consistent with the api spec for v3, in v2 api spec it is still spaceExternalId

PR #680, opened 2023-03-21
Use property name in data model based connections

TharinduDG:
- propertyName should be used when it comes using data models for connections
- Add `And` & `Or` filter support
- Fix default space bug


src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -869,6 +870,17 @@ object FlexibleDataModelRelationUtils {
         s"""Error parsing value of field '$propertyName' as an array of timestamps: ${e.getMessage}""".stripMargin)
     }
 
+  private def extractSpaceOrDefault(schema: StructType, row: Row, defaultSpace: Option[String]) =
+    defaultSpace.map(Right(_)).getOrElse(extractSpace(schema, row)).leftMap { e =>
+      new CdfSparkIllegalArgumentException(
```
vuhainguyen, line 875: can you add a test case to make sure we capture this Exception when the `instanceSpace` is not provided? 😁

vuhainguyen (approved):

PR #682, opened 2023-03-27
Remove session token 5 mins before it expired instead of 30s

vuhainguyen:
We saw some transformation job with many 401, and from the tracing log, it looks like we refresh the session after it expired (range from 5 - 340s), so let's set the `refreshSecondsBeforeExpiration` to 300s instead of 30s, so even though we are a bit late when refreshing the token, we still hit less 401 than the current state since the retry is backoff delay exponential, so the token would be refreshed eventually. 
Reference from: https://cognitedata.slack.com/archives/C93QJMZE0/p1679647547435819?thread_ts=1679416148.705259&cid=C93QJMZE0


TharinduDG (approved):
👍

PR #686, opened 2023-03-30
Add maxWritePartitions parameter

sighol:
This parameter can be used to force a max number of partitions used for writing to a data source.


wjoel:
I haven't looked into the details of this PR, but typically you control this by reducing the number of partitions in your Spark job through `.repartition` or `.coalesce`, not inside the data source itself. Why couldn't (shouldn't) this be done in Jetfire?

vuhainguyen:
> I haven't looked into the details of this PR, but typically you control this by reducing the number of partitions in your Spark job through `.repartition` or `.coalesce`, not inside the data source itself. Why couldn't (shouldn't) this be done in Jetfire?

I think we can call `.repartition` or `.coalesce` before `.write`  in [writeToDestination](https://github.com/cognitedata/jetfire-backend/blob/cb91c6d2801c775ff796b3b228ad75d19807cc32/driver/src/main/scala/com/cognite/jetfire/driver/TransformService.scala#L331) ? 
This piece of code was introduced from https://github.com/cognitedata/cdp-spark-datasource/commit/050b370ae2aed58229b7b153ba9516d0da31a172#diff-7bb7699495f1bf3f09c0d0885c6b97af1e430b11628b68ed4b05a413a4901ded. I think it's fine to have control the partitions in the Data Source not Jetfire, so even Spark DS users can still control the number of partitions through the options when creating the dataframe?

sighol:
Right now, all the data source-specific changes are in `DataframeOptionsExtensions.relationOptions`. Adding `maxWritePartitions` as parameter, allows us to configure all WDL-specific settings in one place.
```scala
case wdl: WellDataLayerRelation =>
  df.option("wdlDataType", wdl.wdlDataType)

  // The well data layer is struggling when there are many concurrent writes to the service.
  // To reduce the problem, set `maxWritePartitions` to a smaller number.
  Environment.forcedWDLPartitions match {
    case Some(maxPartitions) =>
      df.option("maxWritePartitions", maxPartitions.toString)
    case _ =>
  }
  df.option(
    "batchSize",
    wdl.wdlDataType match {
      case "DepthMeasurementIngestion" | "TrajectoryIngestion" => "5"
      case "NptIngestion" | "NdsIngestion" | "RigOperationIngestion" => "1000"
      case _ => "100"
    }
  )
```

related Jetfire PR: https://github.com/cognitedata/jetfire-backend/pull/1863

PR #702, opened 2023-04-19
Remove api key leftovers

TharinduDG:


src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -569,10 +567,13 @@ object DefaultSource {
...
-          GenericClient
-            .forAuthProvider[IO](Constants.SparkDatasourceVersion, authProvider, baseUrl)
-            .map(_.projectName)
+        case CdfSparkAuth.Static(a) =>
+          a.project
+            .traverse(
+              GenericClient
+                .forAuthProvider[IO](Constants.SparkDatasourceVersion, _, authProvider, baseUrl)
+                .map(_.projectName))
+            .map(_.getOrElse(""))
```
TharinduDG, line 576: This feels awkward 🤔

BenChand: Can you do something like this?

```
OptionT.fromOption[IO](a.project)
  .semiflatMap(GenericClient.forAuthProvider[IO](Constants.SparkDatasourceVersion, _, authProvider, baseUrl))
  .map(_.projectName)
  .getOrElse("")
```

dmivankov: why did it work before without `""` fallback and why do we use it now? :thinking:

wjoel: Doesn't all of `getProjectFromAuth` go away without API keys? There's no `login/status` anymore, project must be specified now, as we can't figure out the project just from auth credentials.

In that case, there's no "else" left, so it would just be `val projectName = parameters.get("project")`, I think?

dmivankov (approved):

PR #733, opened 2023-06-06
Remove unstable WDL tests

sighol:
The unstable tests reset the CDF project on every run. That means that if two tests are running concurrently, it will fail.


audunska (approved):

audunska:
Thanks!

PR #736, opened 2023-06-07
autoCreateStart/EndNodes in ConnectionRelation

audunska:
We had added autoCreate parameters in FlexibleDataModelCorePropertyRelation, but I believe we need it in FlexibleDataModelConnectionRelation too.


amorken (approved):

amorken:
shotgun debugging. yeeeehaw!

PR #737, opened 2023-06-07
Re-enable sequence tests

audunska:
Need to increase partition count in some data frames


BenChand (approved):

audunska:
> Wasn't there a problem with the sources limiting partition (aka cursor) count to 10? How can our spark workers use 200 partitions?

The 200 partitions are only for Sequence sources, which is not part of the search API that's limiting the partition count. Timelords have not complained about excessive partitioning, so I'm not worried about right now.

PR #747, opened 2023-06-19
Fixup sonatype publishall

dmivankov:
Looks like that command is only defined at top-level project. It should still be ok to call it there as it shouldn't publish, only find what's published and release that

```
Error:  Not a valid key: sonatypeReleaseAll (similar: sonatypeLogLevel, sonatypeProfileName, sonatypeBundleClean)
Error:  +fatJar/sonatypeReleaseAll
```


audunska (approved):
Worth a shot

PR #755, opened 2023-06-22
Fix: Leave unmentioned properties alone

audunska:
If we don't see a property in the row for a node or edge, don't remove it. This amounts to setting `replace = false` in the sdk call.

We still don't handle nulls completely correctly: If `ignoreNullFields = false`, we ought to remove properties that are null in the row. But let's leave that for later - the first fix is more important to get out.

Also, we need to figure out what to do in the other class, `FlexibleDataModelConnectionRelation`. But that's also for later.


dmivankov (approved):
Don't know what is the bestest thing to do, but sounds reasonable and doesn't break existing tests :grin:

PR #756, opened 2023-06-23
Renovate: Less rebasing

audunska:
Don't trigger automatic rebase of renovate pr unless there's a merge conflict.

This should stop CD from failing once we're merged because of CI overload.


dmivankov (approved):
may even consider moving that to shared config

PR #762, opened 2023-06-28
Fixup ci: run assembly step on correct target

dmivankov:


.github/workflows/main.yaml
```diff
@@ -67,9 +67,9 @@ jobs:
           fail_ci_if_error: false
       - name: Build JAR file
         run: |
-          LIB_PACKAGE="library/package"
+          LIB_PACKAGE="+library/package"
           if [ ${{github.ref }} == "refs/heads/master" ]; then
-            LIB_PACKAGE="+assembly"
+            LIB_PACKAGE="+fatJarShaded/assembly"
```
dmivankov, line 72: with the switch to fatjar only being applied to fatJar target library's assembly target became non-working, but we don't need it to build&publish non-fatjar lib or fatjar jar

.github/workflows/main.yaml
```diff
@@ -67,9 +67,9 @@ jobs:
           fail_ci_if_error: false
       - name: Build JAR file
         run: |
-          LIB_PACKAGE="library/package"
+          LIB_PACKAGE="+library/package"
```
dmivankov, line 70: `+` magic prefix is to run target for all scala versions we build for

skairunner (approved):

PR #764, opened 2023-07-03
Fix slim artifact publish: publish struct_type helper classes artifact too

dmivankov:
Splitting macro library into compile-time macro and runtime helper classes,
publishing runtime classes as a library for non-fatjar artifact.

Moving compile-time macros to `cognite.spark.compiletime.macros` package, also
extracted OptionalField to a separate file.

Verified that compile-time part isn't referenced in artifact
```console
sbt package

for x in `jar tf target/scala-2.12/cdf-spark-datasource_2.12-3.0.0-SNAPSHOT.jar | grep "\.class$" | sed -e 's/\.class$//'`; do
  echo "processing $x";
  javap -c -p -cp target/scala-2.12/cdf-spark-datasource_2.12-3.0.0-SNAPSHOT.jar "$x";
done | grep cognite.spark.compiletime
```

Not adding this a CI check as it is fairly slow and macros part isn't changed too frequently

### DIFF
--- a/macro/src/main/scala/cognite/spark/compiletime/macros/SparkSchemaHelper.scala
+++ b/macro/src/main/scala/cognite/spark/compiletime/macros/SparkSchemaHelper.scala
@@ -1,10 +1,12 @@
-package cognite.spark.v1
+package cognite.spark.compiletime.macros
 
 import org.apache.spark.sql.Row
 import org.apache.spark.sql.types.StructType
 import com.cognite.sdk.scala.common.{NonNullableSetter, Setter}
 
 import scala.reflect.macros.blackbox.Context
+
+import cognite.spark.v1.OptionalField
 
 class SparkSchemaHelperImpl(val c: Context) {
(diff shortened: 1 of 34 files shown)


macro/src/main/scala/cognite/spark/compiletime/macros/SparkSchemaHelper.scala
```diff
@@ -188,11 +190,17 @@ class SparkSchemaHelperImpl(val c: Context) {
...
+
+    q"implicitly[StructTypeEncoder[${weakTypeOf[T]}]].structType()"
+  }
 }
 
 object SparkSchemaHelper {
-  def structType[T]()(implicit encoder: StructTypeEncoder[T]): StructType =
-    encoder.structType()
+  def structType[T](): StructType =
+    macro SparkSchemaHelperImpl.structType[T]
```
dmivankov, line 203: making it macro to avoid direct SparkSchemaHelper reference

audunska (approved):
I know nothing about scala macros, but I'm sure this is fiine :smile:

dmivankov:
:thisisfine:

PR #765, opened 2023-07-07
Handle ignoreNullFields in fdm properties

audunska:
~Introduced a new sum type `RowInstancePropertyValue` to distinguish null properties from non-present properties. This could probably be generalized, but it didn't seem worth it.~
Update: `OptionalField` is the generalized sum type.


src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -514,13 +531,21 @@ object FlexibleDataModelRelationUtils {
     }
   }
 
+  sealed trait RowInstancePropertyValue
+
+  object RowInstancePropertyValue {
+    case object NotPresent extends RowInstancePropertyValue
```
dmivankov, line 537: NotPresent where :thinking: when would this value be created?

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -571,122 +600,124 @@ object FlexibleDataModelRelationUtils {
...
-            Try(InstancePropertyValue.SequenceReferenceList(skipNulls(strSeq).map(String.valueOf))).toEither
-          case t => Left(new CdfSparkException(s"Unhandled list type: ${t.toString}"))
+    Try(schema.fieldIndex(propertyName)) match {
+      case Failure(_) => Right(RowInstancePropertyValue.NotPresent)
+      case Success(i) =>
+        if (row.isNullAt(i)) {
+          if (nullable) {
+            Right(RowInstancePropertyValue.IsNull)
+          } else {
+            Left(new CdfSparkException(s"'$propertyName' cannot be null"))
```
audunska, line 612: Hmm, I wonder if this exception should be thrown only when `ignoreNullFields = false`. If `ignoreNullFields = true`, null fields should behave the same as fields not present, I think? Which should be valid.

This is different from what used to be the behavior, which assumed `replace = true`, in which case both omitting and setting to null would be invalid on a non-nullable property.

Same for the other cases below. Thoughts?

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
audunska: Not present in the spark row. Which is different from being present but null.

dmivankov: how would that happen, it's still value for one of the keys that is defined in transformation sql? Can spark sql produce different things for null and absent, and when would it happen?
Or is it more like us for some reason reading all keys from dms, then trying to apply sparksql updates to it and nopresent is stuff we want to preserve because key isn't mentioned in sparksql?
Also in case you know, how is `select *` different in that regard from `select a,b,c` (assuming for example `a,b,c` is the exhaustive list of columns in source data)?

audunska: See the new test case for some example sql. The properties do not have to be mentioned in the output sql.

The readme says that if a field is not mentioned, there should be no effect on that field. If it is mentioned but set to null, it should depend on ignireNullFields.

audunska: For example:
```sql
select
'space' as space,
'extid' as externalId,
null as property
```
(present but null) vs.
```sql
select
'space' as space,
'extid' as externalId
```
(not present)

dmivankov: Thanks, this clears it up a bit.
What about
```sql
select * from raw_something
```
would it behave same as
```sql
select a,b,c,d,e from raw_something
```
wrt nulls, where `a,b,c,d,e` is a guesstimate from schema inference?

audunska: I don't really know how spark handles `select *`, but I would guess it uses the schemas specified in the `Relation` classes. In this case,
https://github.com/cognitedata/cdp-spark-datasource/blob/7dcbd1c4cdb5e4cc6441c790e704a90beebb5fcf/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertyRelation.scala#L52

which is created by fetching the view.

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -566,148 +579,132 @@ object FlexibleDataModelRelationUtils {
 
   private val timezoneId: ZoneId = ZoneId.of("UTC")
 
+  private def lookupFieldInRow(row: Row, schema: StructType, propertyName: String, nullable: Boolean)(
+      get: => Int => Either[Throwable, InstancePropertyValue])
+    : Either[Throwable, OptionalField[InstancePropertyValue]] =
+    Try(schema.fieldIndex(propertyName)) match {
+      case Failure(_) => Right(FieldNotSpecified)
+      case Success(i) if i >= row.length => Right(FieldNotSpecified)
```
dmivankov, line 587:
```suggestion
      case Success(i) if i >= row.length || i < 0 => Right(FieldNotSpecified)
```
just in case

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -531,18 +548,14 @@ object FlexibleDataModelRelationUtils {
...
+          case _ => toInstancePropertyValueOfNonList(row, schema, propertyName, corePropDef)
         }
       case _: PropertyDefinition.ConnectionDefinition =>
-        val fieldIndex = Try(schema.fieldIndex(propertyName))
-        val nullAtIndex = fieldIndex.map(row.isNullAt).getOrElse(true)
-        if (nullAtIndex) {
-          Right(None)
-        } else {
+        lookupFieldInRow(row, schema, propertyName, true) { _ =>
           extractDirectRelation(propertyName, "Connection Reference", schema, instanceSpace, row)
```
dmivankov, line 556: not sure how to read syntax here :thinking: 
`lookupFieldInRow`'s result is ignored? (looking at `_ =>`)

dmivankov (approved):
`select *` case may need more thought, otherwise as far as I can see does as intended :+1:

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
audunska: Yes, the index that `lookupFieldInRow` provides to the continuation function is ignored. This is the rust convention that I put in unthinkingly. Is there a scala convention for patterns that ignore the argument?

dmivankov: I think that's the convention, but was unsure if it should be ignored

audunska: It's ignored because `extractDirectRelation` does its own index lookup, by the way

PR #767, opened 2023-07-12
POC for multi-output

audunska:
This is proof of concept for how we might support multiple outputs from a single transformation.

The idea is that you write sql like
```sql
select
struct(fields for asset) as a,
struct(fields for event) as e
from...
```
and then in the `type` relation option, you specify the types of those names like `a:asset,e:event`. This would be passed in a more structured json way to the jetfire-backend, but we have to flatten it to a string in spark.

Internally, we parse out the names and types, and then create a relation for each of them. Specific options to each source can be passed by prefixing the options with `${name}:`, i.e., an option named `a:assetSubTreeIds` would be passed to the asset relation.

See the new test in `AssetTests` for a proof that this works in a simple case. Unfortunately, we have to give all fields in the correct order and explicitly cast all null fields for this to typecheck in spark. Hopefully this can be alleviated in some way.

This is just a POC. Ideas / criticisms are welcome!


audunska:
Huh, that test worked locally :thinking:

PR #773, opened 2023-08-10
Add simple test about DefaultSource class

dmivankov:
DefaultSource should have default constructor for
executors to be able to create it.


audunska (approved):
Has this been a problem?

dmivankov:
Yes, hit it during local development around adding tracing. Compiles fine with extra constraints but CI tests break (luckily they do catch it but a dedicated test would be more clear)

PR #783, opened 2023-08-21
Introduce SdkV1InsertableRelation subclass

dmivankov:
OOP overrides hell consequences.
- `SdkV1Relation` was defining an `insert(data, overwrite)` method
  but didn't extend spark's `InsertableRelation`
- `SdkV1Relation` was defining a by-default-throwing `getFromRowsAndCreate`
  that is used by `insert` method above
- some subclasses were defining `getFromRowsAndCreate` and extending
  `InsertableRelation` which would pick up `insert` method in a side-channel
  manner from `SdkV1Relation` (no direct and simple way to see that that
  `insert` was actually meant for `InsertableRelation`)

Let's move insertable part to a dedicated subclass which doesn't have unsafe
fallback and does inherit-override the insert method.

This will remove `InsertableRelation` interface from some of classes,
but only from those that were throwing on using it anyway.

TODO for later:
- relation between `SdkV1InsertableRelation.insert` vs `WritableRelation.insert`
- ignored `overwrite/doUpdate` paramater in some of implementations

### DIFF
--- a/src/main/scala/cognite/spark/v1/DataSetsRelation.scala
+++ b/src/main/scala/cognite/spark/v1/DataSetsRelation.scala
@@ -7,7 +7,7 @@ import com.cognite.sdk.scala.common.WithId
 import com.cognite.sdk.scala.v1.resources.DataSets
 import com.cognite.sdk.scala.v1.{DataSet, DataSetCreate, DataSetFilter, DataSetUpdate, GenericClient}
 import io.scalaland.chimney.Transformer
-import org.apache.spark.sql.sources.{Filter, InsertableRelation}
+import org.apache.spark.sql.sources.Filter
 import org.apache.spark.sql.types.{DataTypes, StructType}
 import org.apache.spark.sql.{Row, SQLContext}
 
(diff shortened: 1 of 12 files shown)


src/main/scala/cognite/spark/v1/DataSetsRelation.scala
```diff
@@ -7,15 +7,14 @@ import com.cognite.sdk.scala.common.WithId
...
-import org.apache.spark.sql.sources.{Filter, InsertableRelation}
+import org.apache.spark.sql.sources.Filter
 import org.apache.spark.sql.types.{DataTypes, StructType}
 import org.apache.spark.sql.{Row, SQLContext}
 
 import java.time.Instant
 
 class DataSetsRelation(config: RelationConfig)(val sqlContext: SQLContext)
     extends SdkV1Relation[DataSet, String](config, "datasets")
-    with InsertableRelation
```
audunska, line 18: This is also kind of confusing: The `DataSetsRelation` _does_ support writing, but does not implement `InsertableRelation`. I guess the `DefaultSource` for writing completely bypasses the `InsertableRelation` interface and uses `WritableRelation` instead?

audunska (approved):
Definitely a helpful step

src/main/scala/cognite/spark/v1/DataSetsRelation.scala
dmivankov: Yep, I don't yet have full insight into DefaultSource vs InsertableRelation implementations

PR #785, opened 2023-08-24
Mark dmsv2 classes as deprecated in favor of dmsv3

dmivankov:


audunska (approved):
I guess we'll want to remove it from the DefaultSource at some point too

dmivankov:
> I guess we'll want to remove it from the DefaultSource at some point too

Yes, either when there's no usage from transformations, or when dmsv2 is shutdown, whichever comes first :slightly_smiling_face:

PR #797, opened 2023-09-01
CDF-19653: make EventsRelationTest less flaky

dmivankov:
Substaintial rewrite of EventsRelationTest but it should
preserve most of its meaning.
    
- explicitly create all items used by test
- introduce extra helper methods and checks
- use randomized event.source for test scope isolation
- un-ignore ignored tests
- some tests may change slightly but intent is to keep
  scenarios approximately the same
    
Extra notes:
- PushdownUtilities convert `> x` to `min(x)` aka `>= x` which
  should ideally be changed to `min(succ(x))` but for later.
  It seems that it doesn't lead to wrong results currently,
  we pushdown broader than needed condition down but afterwards
  spark still applies original exact filter
- adding SparkTest safe counter access which has fallback from
  missing metric to 0 which is useful for retry conditions in tests
  which shouldn't throw if missing metric is a valid scenario part
  and it will appear at subsequent tries.
  In some tests it's still useful to assert that metric exists
  even if it is expected to be 0, so opt-in for now.

[CDF-19653]

[CDF-19653]: https://cognitedata.atlassian.net/browse/CDF-19653?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/scala/cognite/spark/v1/DefaultSource.scala
+++ b/src/main/scala/cognite/spark/v1/DefaultSource.scala
@@ -21,7 +21,6 @@ import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.{DataFrame, Row, SQLContext, SaveMode}
 import sttp.model.Uri
-import scala.reflect.classTag
 
 import scala.reflect.classTag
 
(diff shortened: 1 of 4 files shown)


BenChand (approved):

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -21,6 +21,7 @@ import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.{DataFrame, Row, SQLContext, SaveMode}
 import sttp.model.Uri
+import scala.reflect.classTag
```
dmivankov, line 24:
```suggestion
```

PR #812, opened 2023-09-22
Introduce SparkTest.retryWhileIO method

dmivankov:
`retryWhile` is doing `.unsafeRun*` which should be avoided when possible
aka for example when the caller runs in `IO`.

Introducing new method helper here, to be used later in tests.


src/test/scala/cognite/spark/v1/SparkTest.scala
```diff
@@ -205,21 +205,26 @@ trait SparkTest {
...
       implicit prettifier: Prettifier,
-      pos: source.Position): A =
+      pos: source.Position): IO[A] =
     retryWithBackoff(
-      IO {
-        val actionValue = action
-        if (shouldRetry(actionValue)) {
+      for {
+        actionValue <- action
+        _ = IO.delay { if (shouldRetry(actionValue)) {
```
dmivankov, line 214:
```suggestion
        _ <- IO.delay { if (shouldRetry(actionValue)) {
```

mr-celo (approved):

PR #776, opened 2023-08-15
CDF-18450: add tracing support

dmivankov:
Adding tracing propagation support.

1. Introduce `type TracedIO = Kleisli[IO, Span[IO], _]` and some helper methods for lifting to it from IO, also helpers for passing parent span from Config or across partitioning calls
2. Introduce some operation spans, more detailed spans to be added later
3. Add support for passing parent tracing context via `TRACING_PARAMETER_PREFIX: String = "com.cognite.tracing.parameter."` prefix within parameters
4. Add env vars `OTEL_COLLECTOR_URL`, and optional `OTEL_SERVICE_NAME`, `OTEL_SERVICE_VERSION` that trigger span reporting
5. Bulk of lines changed is just doing IO->TracedIO and adding new span for places with unsafeRunSync

Should be safe to use without tracing as before. To use tracing
- start passing Kernel/parent context parameters with `TRACING_PARAMETER_PREFIX`
- add `OTEL_COLLECTOR_URL` env var to enable trace info reporting

[CDF-18450]

[CDF-18450]: https://cognitedata.atlassian.net/browse/CDF-18450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #836, opened 2023-10-12
Revert "CDF-18450: add tracing support"

dmivankov:
Reverts cognitedata/cdp-spark-datasource#776

There's some potential issue with this PR that fails integration tests in transformations.
Investigations are ongoing but going to take more time, so let's revert to unblock other potential changes


build.sbt
```diff
@@ -29,7 +26,7 @@ lazy val commonSettings = Seq(
   organization := "com.cognite.spark.datasource",
   organizationName := "Cognite",
   organizationHomepage := Some(url("https://cognite.com")),
-  version := "3.1." + patchVersion,
+  version := "3.0." + patchVersion,
```
dmivankov, line 29:
```suggestion
  version := "3.2." + patchVersion,
```

build.sbt
```diff
@@ -29,7 +26,7 @@ lazy val commonSettings = Seq(
   organization := "com.cognite.spark.datasource",
   organizationName := "Cognite",
   organizationHomepage := Some(url("https://cognite.com")),
-  version := "3.1." + patchVersion,
+  version := "3.2." + patchVersion,
```
dmivankov, line 29: revert would be to go back to 3.0, but we don't do parallel branches, only go forward. And the revert is equally big to the original PR to be worth a version change

skairunner (approved):

PR #842, opened 2023-10-16
Check for null items in the delete iterator

mr-celo:


src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -80,7 +80,9 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
 
   def delete(data: DataFrame): Unit =
     data.foreachPartition((rows: Iterator[Row]) => {
-      val deletes = rows.map(r => fromRow[DeleteItemByCogniteId](r))
+      val deletes = rows
+        .map(fromRow[DeleteItemByCogniteId](_))
+        .filter(_ != null)
```
dmivankov, line 85: I don't think pure exposed null is possible here, fromRow macro comes from here https://github.com/cognitedata/cdp-spark-datasource/blob/master/macro/src/main/scala/cognite/spark/compiletime/macros/SparkSchemaHelper.scala
And even it was it doesn't sound too clean to only handle it in this location & without diagnostics to indicate if or when it may happen.

VerstraeteBert (approved):

PR #853, opened 2023-10-27
Repartition batched delete

BenChand:
Refactoring delete method to match other AssetHierarchy operations, using a single partition with `batchedOperation` helper function

It's a bit suspicious that all the other operations are fine when using nullable metadata maps/structs in the query, and the only difference I can really see is that they repartition(size=1) first. Maybe there's a reason delete doesn't do this specifically?

I was wondering if the NPE could be related to some stack overflow issues with workers not having a full spark context (which only comes up when a raw RDD has been set up ??). It could be unrelated, but sparks lazy transformations seem tricky to pin down

https://stackoverflow.com/questions/38986776/nullpointerexception-in-spark-rdd-map-when-submitted-as-a-spark-job

https://stackoverflow.com/questions/32923262/spark-nullpointerexception-with-saveastextfile

https://stackoverflow.com/questions/23793117/nullpointerexception-in-scala-spark-appears-to-be-caused-be-collection-type


dmivankov:
If that's related to NPE debugging, would it be possible to first test if this helps locally or in dev cluster?

dmivankov:
Also is it an attempt to fix NPE or to get more context for when it happens.
Also note that going to java17 may provide more context to NPE due to new "helpful npes" feature

BenChand:
> If that's related to NPE debugging, would it be possible to first test if this helps locally or in dev cluster?

Testing locally might prove difficult since it only appears when there is data coming from raw - dev cluster might be easier to reproduce, but I'm not too familiar with how that works with this cdf-spark-datasource code. Can we run this partition(1) code directly on the dev spark cluster somehow? Or just put it behind a feature flag that's only enabled on the dev cluster jetfire driver

> Also is it an attempt to fix NPE or to get more context for when it happens. Also note that going to java17 may provide more context to NPE due to new "helpful npes" feature

Yeah this is more of an attempt to fix NPE. Might be good to try java17 first and see if that gives any new info

BenChand:
I've put the single partition delete behind a feature flag which can be enabled by adding an option to the dataframe in jetfire, so we can just enable it on azure-dev/greenfield

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -398,6 +398,9 @@ object DefaultSource {
     val collectMetrics = toBoolean(parameters, "collectMetrics")
     val collectTestMetrics = toBoolean(parameters, "collectTestMetrics")
 
+    val enableSinglePartitionDeleteAssetHierarchy =
+      toBoolean(parameters, "enableSinglePartitionDeleteHierarchy")
```
dmivankov, line 402:
```suggestion
      toBoolean(parameters, "enableSinglePartitionDeleteHierarchy", defaultValue = false)
```
maybe even like this, though it is the default default

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -79,23 +79,40 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
...
+          batchedOperation[DeleteItemByCogniteId, Unit](
+            rows.map(r => fromRow[DeleteItemByCogniteId](r)).toVector,
+            // The API calls throw exception when any of the ids do not exist
+            ids =>
+              client.assets
+                .deleteRecursive(ids.map(_.toCogniteId), recursive = true, ignoreUnknownIds = true)
+                .flatTap(_ => incMetrics(itemsDeleted, ids.size))
+                .map(Seq(_)),
+            batchSize = batchSize
+          ).void.unsafeRunSync()
```
dmivankov, line 95:
```suggestion
          ).void.unsafeRunSync()
```
interesting how libs author decided to use `void` name for conversion to `Unit` (parameter) type (which is not same as `Void`/`Nothing` type) :sweat_smile:

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -79,23 +79,40 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
...
-        }
-        .compile
-        .drain
-        .unsafeRunSync()
-    })
+    if (config.enableSinglePartitionDeleteAssetHierarchy) {
+      data
+        .repartition(numPartitions = 1)
+        .foreachPartition((rows: Iterator[Row]) => {
+          batchedOperation[DeleteItemByCogniteId, Unit](
```
dmivankov, line 86: A little more deviation from the other branch than just `repartition`, but on the surface looks ~same with `batchedOperation` helper. If you want to reduce things changed at a time may be worth keeping same code as the other branch (or even wrapping just the `data.` vs `data.repartition(1)` into a conditional)

dmivankov (approved):
Should be safe enough to try

PR #859, opened 2023-11-02
Asset Hierarchy - Remove single partition deletion flag

BenChand:
NPE fixed with single partition deletes so move to that

### DIFF
--- a/src/main/scala/cognite/spark/v1/DefaultSource.scala
+++ b/src/main/scala/cognite/spark/v1/DefaultSource.scala
@@ -406,6 +406,9 @@ object DefaultSource {
     val collectMetrics = toBoolean(parameters, "collectMetrics")
     val collectTestMetrics = toBoolean(parameters, "collectTestMetrics")
 
+    val userSinglePartition =
+      toBoolean(parameters, "useSinglePartition", defaultValue = false)
+
     val saveMode = parseSaveMode(parameters)
     val parallelismPerPartition = {
       toPositiveInt(parameters, "parallelismPerPartition").getOrElse(
@@ -455,7 +458,8 @@ object DefaultSource {
       deleteMissingAssets = toBoolean(parameters, "deleteMissingAssets"),
       subtrees = subtreesOption,
       ignoreNullFields = toBoolean(parameters, "ignoreNullFields", defaultValue = true),
-      rawEnsureParent = toBoolean(parameters, "rawEnsureParent", defaultValue = true)
+      rawEnsureParent = toBoolean(parameters, "rawEnsureParent", defaultValue = true),
+      useSinglePartition = userSinglePartition
     )
   }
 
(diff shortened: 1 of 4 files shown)


src/main/scala/cognite/spark/v1/RelationConfig.scala
```diff
@@ -21,8 +21,7 @@ final case class RelationConfig(
     deleteMissingAssets: Boolean,
     subtrees: AssetSubtreeOption,
     ignoreNullFields: Boolean,
-    rawEnsureParent: Boolean,
-    enableSinglePartitionDeleteAssetHierarchy: Boolean // flag to test whether single partition helps avoid NPE in asset hierarchy builder
```
dmivankov, line 25: I think we might want to keep it for potential future investigations if we see weird NPEs again, that be a way to get one of them ready

BenChand: Ah fair enough, I'll rename it to "useSinglePartition" or something

src/main/scala/cognite/spark/v1/RelationConfig.scala
```diff
@@ -22,7 +22,7 @@ final case class RelationConfig(
     subtrees: AssetSubtreeOption,
     ignoreNullFields: Boolean,
     rawEnsureParent: Boolean,
-    enableSinglePartitionDeleteAssetHierarchy: Boolean // flag to test whether single partition helps avoid NPE in asset hierarchy builder
+    useSinglePartition: Boolean // a flag for helping to test NPEs by moving the data frame to a single partition
```
dmivankov, line 25:
```suggestion
    useSinglePartitionForAssetHierarchyDeletion: Boolean // a flag for helping to test NPEs by moving the data frame to a single partition
```
or something. Maybe invert the name so that default values is false and it'd mean do repartition to 1. Also let's use repartition as default here

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -406,8 +406,8 @@ object DefaultSource {
     val collectMetrics = toBoolean(parameters, "collectMetrics")
     val collectTestMetrics = toBoolean(parameters, "collectTestMetrics")
 
-    val enableSinglePartitionDeleteAssetHierarchy =
-      toBoolean(parameters, "enableSinglePartitionDeleteHierarchy", defaultValue = false)
+    val userSinglePartition =
+      toBoolean(parameters, "useSinglePartition", defaultValue = false)
```
dmivankov, line 410:
```suggestion
      toBoolean(parameters, "useSinglePartition", defaultValue = true)
```

dmivankov, line 410:
```suggestion
      toBoolean(parameters, "useSinglePartitionForAssetHierarchyDeletes", defaultValue = true)
```
not sure about the name...

PR #890, opened 2024-01-29
CDF-20827: make FilesRelationTest less flaky

dmivankov:
Create all files (&labels) that test expects to see as part of test setup,
cleanup afterwards. Should make this test work in any project.

Changes are similar to EventRelationTest

Overall setup isn't super-pretty yet, but that's a task for another day.
For now it should fix tests in `master` and not depend on state of data in test project.

[CDF-20827]

[CDF-20827]: https://cognitedata.atlassian.net/browse/CDF-20827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test/scala/cognite/spark/v1/FilesRelationTest.scala
+++ b/src/test/scala/cognite/spark/v1/FilesRelationTest.scala
@@ -1,107 +1,174 @@
 package cognite.spark.v1
 
-import cognite.spark.v1.CdpConnector.ioRuntime
+import cats.effect.{IO, Resource}
+import com.cognite.sdk.scala.v1.{File, FileCreate, Label, LabelCreate}
 import io.scalaland.chimney.dsl._
-import org.apache.spark.sql.Row
+import org.apache.spark.sql.{DataFrame, DataFrameReader, Row}
 import org.apache.spark.sql.functions.col
 import org.scalatest.{FlatSpec, Matchers, ParallelTestExecution}
 
-import scala.util.control.NonFatal
-
 class FilesRelationTest extends FlatSpec with Matchers with ParallelTestExecution with SparkTest {
-  val sourceDf = dataFrameReaderUsingOidc
-    .option("type", "files")
-    .load()
+  val testSource = s"FilesRelationTest-${shortRandomString()}"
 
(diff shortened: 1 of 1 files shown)


src/test/scala/cognite/spark/v1/FilesRelationTest.scala
```diff
@@ -1,107 +1,172 @@
...
 import org.apache.spark.sql.functions.col
 import org.scalatest.{FlatSpec, Matchers, ParallelTestExecution}
 
-import scala.util.control.NonFatal
-
 class FilesRelationTest extends FlatSpec with Matchers with ParallelTestExecution with SparkTest {
-  val sourceDf = dataFrameReaderUsingOidc
-    .option("type", "files")
-    .load()
+  val testSource = s"FilesRelationTest-${shortRandomString()}"
```
dmivankov, line 11:
```suggestion
  lazy val testSource = s"FilesRelationTest-${shortRandomString()}"
```

src/test/scala/cognite/spark/v1/FilesRelationTest.scala
```diff
@@ -1,107 +1,174 @@
...
+    val targetView = s"destinationFile_${shortRandomString()}"
+    val metricsPrefix = s"test.${shortRandomString()}"
+    Resource.make {
+      IO.blocking {
+        val destinationDf: DataFrame = spark.read
+          .format(DefaultSource.sparkFormatString)
+          .useOIDCWrite
+          .option("type", "files")
+          .option("collectMetrics", "true")
+          .option("metricsPrefix", metricsPrefix)
```
mr-celo, line 23: Is this need for more tests than the `support updates using id and externalId` one?

dmivankov: metricsPrefix part seems to be only used there, shouldn't be getting in the way for other tests I think.
EventsRelationTest has similar thing and there it is used more often so transferred in same form just in case

mr-celo (approved):

PR #900, opened 2024-02-08
FDM sync cursors

kornelione:
Data source extension to support sync endpoint.

- Extends the existing corePropertyRelation with a new implementation which uses sync endpoint instead of filter
- At data source level we use the corePropertySyncRelation if we get a cursor as one of the paramters along with instance and view reference
- We pass down the cursor with the option, so jetfire can set it from a previous run
- We pass out the cursor as a named "metadata.cursor", so jetfire can store it in db
- The sync relation will also pass out other node and edge attributes, and prefix those with node. and edge. (lastupdated, created, deleted and version) to not conflict with other attributes.
- It will pass down filters on space, externalId, type, startNode and endNode attributes, not other filters. It will additionally  pass down filter on HasData for the view.


src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  private def syncOut(syncRequest: InstanceSyncRequest)(
+      implicit F: Async[IO]): Stream[IO, InstanceDefinition] =
+    Stream.eval {
+      syncWithCursor(syncRequest).map { sr =>
+        val items = sr.items
+        val nextCursor = sr.nextCursor
+
+        val next = (nextCursor, items.nonEmpty) match {
+          case (Some(cursor), true) => syncOut(syncRequest.copy(cursors = Some(Map("sync" -> cursor))))
```
dmivankov, line 78: is there a plan already on how would saving the cursor from last iteration would look like?

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -0,0 +1,91 @@
...
+        cursors = None,
+        select = Map("sync" -> SelectExpression(sources = sourceReference)))
+    }
+    syncRequests.distinct.map { sr =>
+      syncOut(sr).map(toProjectedInstance(_, selectedInstanceProps))
+    }
+  }
+
+  private def syncOut(syncRequest: InstanceSyncRequest)(
+      implicit F: Async[IO]): Stream[IO, InstanceDefinition] =
```
dmivankov, line 71: just a thought, maybe scala sdk could be extended to produce stream and last cursor as it might be a useful utility, also iirc there's already some paging->stream support there

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -0,0 +1,91 @@
...
+          r =>
+            SourceSelector(
+              source = r,
+              properties = selectedInstanceProps.toIndexedSeq.filter(p =>
+                p != "startNode" && p != "endNode" && p != "space" && p != "externalId" && p != "type")))
+        .toSeq
+
+      InstanceSyncRequest(
+        `with` = Map("sync" -> tableExpression),
+        cursors = None,
```
thorkildcognite, line 62: Due to how jetfire and our sparksql bindings are looking like now, we will pass the sync cursor down with the set up of the datasource (as an option()). We will also limit ourselves to one cursor at a time, as there is no way I have found that you can specify multiple with clauses from Spark SQL. 

We need to pick up this cursor and propagate it here, I guess?

dmivankov: Additional question is how to pass it out. And to double-check that cursor can be used multiple times in case job has to restart

thorkildcognite: Passing it out will be independent of the spark data source (like done for cursors today), directly to the incremental version storage. 

Yes, the same cursors can be used multiple times.

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -0,0 +1,129 @@
...
+    val selectedInstanceProps = if (selectedColumns.isEmpty) {
+      schema.fieldNames
+    } else {
+      selectedColumns
+    }
+
+    val instanceType = intendedUsage match {
+      case Usage.Edge => InstanceType.Edge
+      case Usage.Node => InstanceType.Node
+      case Usage.All => throw new CdfSparkIllegalArgumentException("Cannot sync both nodes and edges")
```
thorkildcognite, line 70: +  " at the same time."

thorkildcognite (approved):
Please consider my comments - LGTM.

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -530,6 +535,38 @@ object DefaultSource {
...
+          )(ViewReference.apply)
+        FlexibleDataModelRelationFactory.corePropertySyncRelation(
+          usageAndCursor._2,
+          config = config,
+          sqlContext = sqlContext,
+          viewCorePropConfig = ViewSyncCorePropertyConfig(
+            intendedUsage = usage,
+            viewReference = viewReference,
+            cursor = usageAndCursor._2,
+            instanceSpace = parameters.get("instanceSpace"))
```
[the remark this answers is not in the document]
kornelione: The relations for CDF need to support writes, so this relation will inherit the behviour for write from the fdm core relation.

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
```diff
@@ -274,6 +219,11 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
             case s if s.equalsIgnoreCase("space") => n.space
             case s if s.equalsIgnoreCase("spaceExternalId") => n.space
             case s if s.equalsIgnoreCase("externalId") => n.externalId
+            case s if s.equalsIgnoreCase("metadata.cursor") => cursor.getOrElse("")
```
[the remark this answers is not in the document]
kornelione: Since "." is disallowed in all property ids (and spaces and views), the thinking is that metadata. as prefix is safe. Could of course use node or edge.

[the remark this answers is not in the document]
thorkildcognite: oh, that is a good point!

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
```diff
@@ -288,6 +238,11 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
             case s if s.equalsIgnoreCase("startNode") => Array(e.startNode.space, e.startNode.externalId)
             case s if s.equalsIgnoreCase("endNode") => Array(e.endNode.space, e.endNode.externalId)
             case s if s.equalsIgnoreCase("type") => Array(e.`type`.space, e.`type`.externalId)
+            case s if s.equalsIgnoreCase("metadata.cursor") => cursor.getOrElse("")
```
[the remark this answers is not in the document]
kornelione: done

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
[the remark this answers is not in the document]
thorkildcognite: Out of the ones listed, the following are reserved:
lastUpdatedTime
deletedTime
createdTime

The ones that are not:
- version
- cursor

Since the first ones are reserved, should we not prefix those?

src/test/scala/cognite/spark/v1/AssetsRelationTest.scala
```diff
@@ -953,28 +953,36 @@ class AssetsRelationTest extends FlatSpec with Matchers with ParallelTestExecuti
         .save()
 
       val id = writeClient.assets.retrieveByExternalId(externalId).unsafeRunSync().id
+      println(externalId)
+      println(source)
+      println(id)
```
dmivankov, line 958: I have a suspicion that https://github.com/cognitedata/cdp-spark-datasource/blob/master/src/test/scala/cognite/spark/v1/AssetsRelationTest.scala#L29 load around there may interfere with tests retries where load happens once and retries won't help refresh that view. Don't have the full picture, but removing that from per-class to per-test (and with randomized name?) might be worth a shot

kornelione: reverted that change now, wonder if theses tests that copy publicdata would suffer if publicdata changes asset hierarchies..

kornelione: seems the tests worked better when disabling parallell execution of AssetRelationTest .. maybe in conflict with itself..

PR #901, opened 2024-02-08
Remove the need for requesting view in every filter request

kornelione:
- Map data types based on schema already inferred from the view
- Using sync will rely on this ability.
- With this in place, we can remove the call to includeTyping on filter.


dmivankov:
Timeseries need some, hopefully simple, update for units coming with newer sdk.
Could be done in https://github.com/cognitedata/cdp-spark-datasource/pull/894 or a separate PR

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
```diff
@@ -56,36 +56,65 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
...
-    case InstancePropertyValue.TimestampList(value) =>
-      value.map(v => java.sql.Timestamp.from(v.toInstant))
-    case InstancePropertyValue.ObjectList(value) => value.map(_.noSpaces)
-    case InstancePropertyValue.TimeSeriesReference(value) => value
-    case InstancePropertyValue.FileReference(value) => value
-    case InstancePropertyValue.SequenceReference(value) => value
-    case InstancePropertyValue.TimeSeriesReferenceList(value) => value
-    case InstancePropertyValue.FileReferenceList(value) => value
-    case InstancePropertyValue.SequenceReferenceList(value) => value
+  private def extractInstancePropertyValue(key: String, value: InstancePropertyValue): Any = {
```
dmivankov, line 61: looks like this one should be the inverse of `deriveViewPropertySchemaWithUsageSpecificAttributes`, not sure if there's an easier way to construct them both, will try to cross-check at least

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
```diff
@@ -56,36 +56,65 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
     })
   }
 
+  // scalastyle:off method.length
   // scalastyle:off cyclomatic.complexity
-  private def extractInstancePropertyValue: InstancePropertyValue => Any = {
```
dmivankov, line 60: A bit unsure what changes and why we'll no longer need typing info/view with this change.
Before it was doing FDM API values -> plain values conversion based on their type.
Now it still uses type but adds more casts (to fit spark schema better?) and looks into spark schema.
So there's no change around looking into FDM or view? On itself of course looks safe, just not quite sure how it connects to the sync endpoint work

kornelione: there are two paths in the sparc sdks for deserialising the json into 
InstancePropertyValue.Int32 or InstancePropertyValue.Int64. One path is if you fetch down the view definition inline in the request, then the deserialiser will look at the view and resolve correct type. Otherwise it will select one base on the shape of the data, and can end up with any of int, long, float or double..

dmivankov: Ah, ok, so sdk call parameter change is the goal but then we need to do some type conversions :+1:

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
kornelione: yes deriveViewPropertySchemaWithUsageSpecificAttributes derives the sparc schema from the view. Here we use it to set the "correct" type from json..

kornelione:
> Would be nice to somehow make these conversions shorter or have more tests, but don't have quick ideas atm and can be improved later

We could consider have another go at scala sdk to only use double and long (or maybe bigdecimal), then fewer things to convert from / to.
Some implicit coverage in the integration tests, probably does not cover all paths.

PR #909, opened 2024-02-27
Add parameter to enable/disable shared request throttle

dmivankov:
Retrying http client already does exponential backoff with jitter and
that should backoff and spread the load on failed requests.

BackpressureThrottlingBackend reacts to any 429/503 and throttles all
in-flight requests for a random period of time for 400..800ms.
Deficiencies:
- it is randomized but unblocks all throttled requests simultaneously
- it applies throttle to all resources even if they are unrelated

Adding an option to disable backpressure backend and rely only on
established exponential backoff algorithm.

Maybe for later we could revisit how to coordinate throttling for
requests for same resource, so when one request hits 429/503 all
new requests for that client would slow down before even trying
their luck. This needs to be randomized and improved relative
to current BackpressureThrottlingBackend, so for some other day.


src/main/scala/cognite/spark/v1/CdpConnector.scala
```diff
@@ -80,12 +80,21 @@ object CdpConnector {
...
+    new BackpressureThrottleBackend[IO, Any](sttpBackend, makeQueueOf1.unsafeRunSync(), 800.milliseconds)
+  }
+
   def retryingSttpBackend(
       maxRetries: Int,
       maxRetryDelaySeconds: Int,
       maxParallelRequests: Int = Constants.DefaultParallelismPerPartition,
       metricsPrefix: Option[String] = None,
-      initialRetryDelayMillis: Int = Constants.DefaultInitialRetryDelay.toMillis.toInt
+      useSharedThrottle: Boolean = false
```
dmivankov, line 97:
```suggestion
      initialRetryDelayMillis: Int = Constants.DefaultInitialRetryDelay.toMillis.toInt,
      useSharedThrottle: Boolean = false
```

src/main/scala/cognite/spark/v1/CdpConnector.scala
```diff
@@ -80,12 +80,22 @@ object CdpConnector {
         case _ => IO.unit
       })
 
+  def throttlingSttpBackend(sttpBackend: SttpBackend[IO, Any]): SttpBackend[IO, Any] = {
+    // this backend throttles all requests when rate limiting from the service is encountered
+    val makeQueueOf1 = for {
+      queue <- Queue.bounded[IO, Unit](1)
+      _ <- queue.offer(())
+    } yield queue
+    new BackpressureThrottleBackend[IO, Any](sttpBackend, makeQueueOf1.unsafeRunSync(), 800.milliseconds)
```
mr-celo, line 89: I know this was here before, but won't this block the thread every time more than one request is queued for this backend? Isn't there an async alternative for this?

mr-celo (approved):

src/main/scala/cognite/spark/v1/CdpConnector.scala
dmivankov: You mean the wait on queue take/offer?
https://github.com/cognitedata/cognite-sdk-scala/blob/master/src/main/scala/com/cognite/sdk/scala/sttp/BackpressureThrottleBackend.scala#L30 uses it as a mutex
On no 503/429 it is quickly acquired and released, but on 503/429 the request hitting those takes it for longer and blocks other requests by design.
This should all be within cats-effects mechanism so shouldn't block on compute threads but rather shift to io threads or do async suspend.

dmivankov:
Will hold off merging until next week

PR #913, opened 2024-03-06
Drop hasData filter if we get 408 on sync, and filter on client

kornelione:
The hasData filter on sync can lead to sync timing out with 408.
This fix tries to address this with dropping the hasData filter once we get 408, and sync out without it.
Downside is that it means we will sync a lot of nodes out that do not have data for the view, and we need to filter on client side. Upside is that if we get out data and a cursor, next time the user runs the transformation, it may be able to proceed without dropping the hasData filter.

### DIFF
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
@@ -16,6 +16,7 @@ import fs2.Stream
 import org.apache.spark.sql.SQLContext
 import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types.{DataTypes, StructField}
+import org.log4s.getLogger
 
 /**
   * Flexible Data Model Relation for syncing Nodes or Edges with properties
@@ -36,8 +37,11 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
         corePropConfig.instanceSpace)
     )(sqlContext) {
 
+  private val logger = getLogger
   // request no more data after seeing items with node/edge lastUpdatedTime >= terminationTimeStamp
   private val terminationTimeStamp = System.currentTimeMillis()
+  private val mandatoryViewFields = schema.fields.filter(f =>
+    f.name != "space" && f.name != "externalId" && f.name != "type" && f.name != "startNode" && f.name != "endNode" && !f.nullable)
 
   protected override def metadataAttributes(): Array[StructField] = {
     val nodeAttributes = Array(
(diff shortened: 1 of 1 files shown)


src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -36,8 +37,11 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
         corePropConfig.instanceSpace)
     )(sqlContext) {
 
+  private val logger = getLogger
   // request no more data after seeing items with node/edge lastUpdatedTime >= terminationTimeStamp
   private val terminationTimeStamp = System.currentTimeMillis()
+  private val mandatoryViewFields = schema.fields.filter(f =>
+    f.name != "space" && f.name != "externalId" && f.name != "type" && f.name != "startNode" && f.name != "endNode" && !f.nullable)
```
dmivankov, line 44: isn't space,externalId mandatory?
Maybe should be called `requiredViewProperties` and a comment that it doesn't include system/non-customizable

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -108,56 +137,84 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
+      selectedColumns: Array[String],
+      currentCursor: Option[String])(
+      implicit F: Async[IO]): Stream[IO, ProjectedFlexibleDataModelInstance] = {
+
+    val synced: IO[(ItemsWithCursor[InstanceDefinition], Boolean)] =
+      syncWithCursor(hasDataFilter, filters, selectedColumns, currentCursor)
+    val stream: Stream[IO, ProjectedFlexibleDataModelInstance] = Stream.eval(synced).flatMap {
+      case (ItemsWithCursor(items, nextCursor), continueWithHasData) =>
+        val projectedItems =
+          items.filter(clientsideHasDataFilter).map(toProjectedInstance(_, nextCursor, selectedColumns))
```
dmivankov, line 164: once it finds a page with data should it try to switch back to server-side filtering?

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -108,56 +137,84 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
-          }
 
-        val projected = items.map(toProjectedInstance(_, nextCursor, selectedProps))
-        fs2.Stream.emits(projected) ++ next
-      }
-    }.flatten
+  private def clientsideHasDataFilter(item: InstanceDefinition): Boolean = {
+    val values = item.properties.getOrElse(Map.empty).values.flatMap(_.values).fold(Map.empty)(_ ++ _)
+    values.nonEmpty &&
+    mandatoryViewFields.forall { field =>
```
dmivankov, line 179: A little unsure about this part, https://api-docs.cognite.com/20230101/tag/Instances/operation/syncContent#!path=with/property%20name*/0/nodes/filter/1/10/hasData&t=request and docs around don't say a lot about
- in /sync endpoint are all properties returned or only changed (seems like all)
- should it be an error if some of mandatory fields are present but some are missing?
- what happens if hasData filter isn't set in request? Would that mean getting unrelated views' changes?
- could we instead filter by view space+externalId?

Maybe also add a sanity assert or test that if server-side hasData filtering is applied then clientside filter agrees?
Also if not too hard to set up a unit test that compares clientside <> server-side filtering logic would be nice

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -108,56 +137,84 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
+
+    val synced: IO[(ItemsWithCursor[InstanceDefinition], Boolean)] =
+      syncWithCursor(hasDataFilter, filters, selectedColumns, currentCursor)
+    val stream: Stream[IO, ProjectedFlexibleDataModelInstance] = Stream.eval(synced).flatMap {
+      case (ItemsWithCursor(items, nextCursor), continueWithHasData) =>
+        val projectedItems =
+          items.filter(clientsideHasDataFilter).map(toProjectedInstance(_, nextCursor, selectedColumns))
+
+        val nextStream = nextCursor match {
+          case Some(c) if items.nonEmpty && items.last.lastUpdatedTime < terminationTimeStamp =>
```
dmivankov, line 167: could `items` still be empty due to client side filter but that won't be the end of sync stream? Maybe nonEmpty needs to be applied to unfiltered server response

dmivankov, line 167: is there a guarantee that items come in timestamp order in on response? Might be safer to pick max? Also might make sense to do it on unfiltered response

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
kornelione: right, need to node include node/edge properties here, only view properties.. 
can rename and comment

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
dmivankov, line 164: just to double-check, we're still going to pass out the very last cursor via a column? What would be the way to pick up the most recent one in case of multiple page iterations? (safe to pick any but would be nice to pick latest, for that probably need some way to mark it, would passing a "page number" along with nextCursor be an option?)

kornelione: We pass cursor on every item, also along with node/edge.lastupdatedtime. If it is necessary to create logic to pick the latest, one could try to pick the one with highest lastupdatedtime. Not guaranteed to be latest cursor, but most likely is, or page -1.

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
kornelione: items comes in timestamp orderish.. but can use max to improve the logic, should also be done on unfiltered yes..

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
kornelione: can try that .

dmivankov:
This one gets superseded but the other PR around 408?

kornelione:
dropping this for now, since it got superseded by the other PR

PR #915, opened 2024-03-18
Enable two-phase syncing to avoid 408

kornelione:
General idea is to do syncing in two phases:

Phase 1: list out items with cursor on internal node id in DMS

Phase 2: sync out incremental changes that happened since phase 1

We will initialise a cursor for phase 1 when we start it.

### DIFF
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
@@ -149,7 +149,7 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
             IO.raiseError(
               new IllegalArgumentException("/sync response must have " +
                 "nextCursor"))
-          case Some(nextCursor) => IO.pure(nextCursor, cursors.isEmpty)
+          case Some(nextCursor) => IO.pure((nextCursor, cursors.isEmpty))
         }
       }
       .redeemWith(
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
@@ -126,7 +126,7 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
   }
 
   /**
-    * Generate future items cursor.
+    * Generate future items cursor and validate cursor expiration.
     * If the cursor has expired, generate a new cursor and do a full back fill.
     */
   private def generateFutureItemsCursor(
@@ -140,7 +140,7 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
         "sync" -> generateTableExpression(instanceType, FilterDefinition.Not(MatchAll(JsonObject())))),
       select = select)
       .map { sr =>
-        (sr.nextCursor, cursor.isEmpty)
+        (sr.nextCursor, cursors.isEmpty)
       }
       .redeemWith(
         {
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
@@ -124,6 +124,8 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
         selectedColumns
       ))
   }
+  // scalastyle:on method.length
+  // scalastyle:on cyclomatic.complexity
 
   private val matchNothingFilter: FilterDefinition =
     FilterDefinition.Not(MatchAll(JsonObject()))
@@ -195,7 +197,6 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
       ItemsWithCursor(itemDefinitions, nextCursor)
     }
   }
-  // scalastyle:on cyclomatic.complexity
 
   private def syncOut(
       futureItemsCursor: Option[String],
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
@@ -141,8 +141,7 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
     fetchData(
       isBackFill = false,
       cursors = cursors,
-      `with` = Map(
-        "sync" -> generateTableExpression(instanceType, matchNothingFilter)),
+      `with` = Map("sync" -> generateTableExpression(instanceType, matchNothingFilter)),
       select = select)
       .map { sr =>
         (sr.nextCursor, cursors.isEmpty)
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
@@ -127,9 +127,7 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
 
   /**
     * Generate future items cursor.
-    * If the cursor has expired, generate a new cursor and do a full backfill.
-    * If the cursor with sync causes 408, then do a backfill, and pray that incremental sync sessions
-    * will not time out.
+    * If the cursor has expired, generate a new cursor and do a full back fill.
     */
   private def generateFutureItemsCursor(
       cursors: Option[Map[String, String]],
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
@@ -190,7 +190,7 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
       )
     }
     response.map { qr =>
-      val itemDefinitions = qr.items.getOrElse(Map.empty).get("sync").getOrElse(Vector.empty)
+      val itemDefinitions = qr.items.getOrElse(Map.empty).getOrElse("sync", Vector.empty)
       val nextCursor = qr.nextCursor.getOrElse(Map.empty).get("sync")
       ItemsWithCursor(itemDefinitions, nextCursor)
     }
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
@@ -7,7 +7,7 @@ import cognite.spark.v1.FlexibleDataModelRelationFactory.{
   ViewCorePropertyConfig,
   ViewSyncCorePropertyConfig
 }
-import com.cognite.sdk.scala.common.ItemsWithCursor
+import com.cognite.sdk.scala.common.{CdpApiException, ItemsWithCursor}
 import com.cognite.sdk.scala.v1.GenericClient
 import com.cognite.sdk.scala.v1.fdm.common.Usage
 import com.cognite.sdk.scala.v1.fdm.common.filters.FilterDefinition
@@ -18,6 +18,7 @@ import io.circe.JsonObject
 import org.apache.spark.sql.SQLContext
 import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types.{DataTypes, StructField}
+import org.log4s.getLogger
 
 /**
   * Flexible Data Model Relation for syncing Nodes or Edges with properties
@@ -37,6 +38,8 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
         corePropConfig.viewReference,
         corePropConfig.instanceSpace)
     )(sqlContext) {
+
+  private val logger = getLogger
 
   // request no more data after seeing items with node/edge lastUpdatedTime >= terminationTimeStamp
   private val terminationTimeStamp = System.currentTimeMillis()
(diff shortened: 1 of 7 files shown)


src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -111,53 +108,106 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
+        val projectedCursorId = if (futureItems.isEmpty) nextCursor else futureItems
+        val projected = items.map(toProjectedInstance(_, projectedCursorId, selectedProps))
         fs2.Stream.emits(projected) ++ next
       }
     }.flatten
 
-  private def syncWithCursor(
-      syncRequest: InstanceSyncRequest): IO[ItemsWithCursor[InstanceDefinition]] = {
-    // temporary handle expired cursor by syncing from scratch
-    val response = client.instances
```
thorkildcognite, line 146: I kind of liked this behavor, since it would never make the job fail.

thorkildcognite (approved):
Apart from the fact the query isn't working at all (which is a DMS/PG3 issue), this is the backfill change. Still think we should also add expose the option to DMS to allow for old cursors in the case where jobs are delayed, or maybe handle it with your too-old-retry.

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
kornelione: added back support for expired cursors handling

thorkildcognite (approved):
LGTM!

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -110,54 +110,111 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
 
-    val instanceSyncRequest = InstanceSyncRequest(
-      `with` = Map("sync" -> tableExpression),
-      cursors = if (cursor.nonEmpty) Some(Map("sync" -> cursor)) else None,
-      select = Map("sync" -> SelectExpression(sources = sourceReference)),
-      includeTyping = Some(true)
-    )
+  /**
+    * Generate future items cursor and validate cursor expiration.
+    * If the cursor has expired, generate a new cursor and do a full back fill.
```
dmivankov, line 130: What are "future" items? Is it different from what items will be returned when we make a request with cursors?

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -110,54 +110,111 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
+    */
+  private def generateFutureItemsCursor(
+      cursors: Option[Map[String, String]],
+      instanceType: InstanceType,
+      select: Map[String, SelectExpression]): IO[(Option[String], Boolean)] =
+    fetchData(
+      isBackFill = false,
+      cursors = cursors,
+      `with` = Map(
+        "sync" -> generateTableExpression(instanceType, FilterDefinition.Not(MatchAll(JsonObject())))),
```
dmivankov, line 140: So this condition is basically MatchNone? Is it allowed in dms api to change filtering condition between pages?

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -110,54 +110,111 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
+  /**
+    * Generate future items cursor and validate cursor expiration.
+    * If the cursor has expired, generate a new cursor and do a full back fill.
+    */
+  private def generateFutureItemsCursor(
+      cursors: Option[Map[String, String]],
+      instanceType: InstanceType,
+      select: Map[String, SelectExpression]): IO[(Option[String], Boolean)] =
+    fetchData(
+      isBackFill = false,
```
dmivankov, line 137:
```suggestion
      useSync = false,
```
Looks like it leaked here that fetchData shouldn't be thinking about backfill or not but rather should just accept whether to use sync or query endpoint

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -110,54 +110,111 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
+        "sync" -> generateTableExpression(instanceType, FilterDefinition.Not(MatchAll(JsonObject())))),
+      select = select)
+      .map { sr =>
+        (sr.nextCursor, cursors.isEmpty)
+      }
+      .redeemWith(
+        {
+          case e: CdpApiException if e.code == 400 && e.message.startsWith("Cursor has expired") =>
+            logger.warn("Cursor has expired, doing a full backfill")
+            generateFutureItemsCursor(None, instanceType, select)
```
dmivankov, line 149: With expired cursors it will call into fetchData using sync endpoint, is it expected? Then it will return `(sr.nextCursor, true)` which it seems should mean that called needs to use `query` endpoint to do a backfill starting from `sr.nextCursor`? But is cursor returned from /sync a valid cursor for /query?

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -110,54 +110,111 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
+      val itemDefinitions = qr.items.getOrElse(Map.empty).get("sync").getOrElse(Vector.empty)
+      val nextCursor = qr.nextCursor.getOrElse(Map.empty).get("sync")
+      ItemsWithCursor(itemDefinitions, nextCursor)
+    }
   }
   // scalastyle:on cyclomatic.complexity
 
-  private def syncOut(syncRequest: InstanceSyncRequest, selectedProps: Array[String])(
+  private def syncOut(
+      futureItems: Option[String],
```
dmivankov, line 198:
```suggestion
      futureItemsCursor: Option[String],
```

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -110,54 +110,111 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
+      cursors: Option[Map[String, String]],
+      `with`: Map[String, TableExpression],
+      select: Map[String, SelectExpression],
+      selectedProps: Array[String])(
       implicit F: Async[IO]): Stream[IO, ProjectedFlexibleDataModelInstance] =
     Stream.eval {
-      syncWithCursor(syncRequest).map { sr =>
+      fetchData(futureItems.nonEmpty, cursors, `with`, select).map { sr =>
```
dmivankov, line 205: How is presence of future items cursor related to backfill? generateFutureItemsCursor seems to base _2 returned boolean independently of returning the cursor?

It feels like some of confusion could be avoided if different wrapper types were used for /sync and /query cursors and if pairs of options etc were converted to case classes that represent different modes..

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
kornelione: The implementation is now that on the first sync, we do a backfill of data with cursor by id on the query endpoint. Basically list out all data. But before that, we ask the sync endpoint to create a cursor for future items from right now.

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
kornelione: yes api does not validate the filters are the same.

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
dmivankov: Is there an api to start sync from now() or now() minus some threshold?

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
kornelione: The backfill will start from empty cursor, and then continue cursoring on cursor by query. However the cursor string we return from the client will be future items cursor from sync, which will be used on next sync session..

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -110,54 +110,114 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
         val next =
           (nextCursor, items.nonEmpty && items.last.lastUpdatedTime < terminationTimeStamp) match {
             case (Some(cursor), true) =>
-              syncOut(syncRequest.copy(cursors = Some(Map("sync" -> cursor))), selectedProps)
+              syncOut(futureItemsCursor, Some(Map("sync" -> cursor)), `with`, select, selectedProps)
             case _ => fs2.Stream.empty
           }
-
-        val projected = items.map(toProjectedInstance(_, nextCursor, selectedProps))
+        val projectedCursorId = if (futureItemsCursor.isEmpty) nextCursor else futureItemsCursor
```
dmivankov, line 217:
```suggestion
        val projectedCursorId = if (!executeBackFill) nextCursor else futureItemsCursor
```

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -110,54 +110,114 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
+  private def syncOut(
+      futureItemsCursor: Option[String],
+      cursors: Option[Map[String, String]],
+      `with`: Map[String, TableExpression],
+      select: Map[String, SelectExpression],
+      selectedProps: Array[String])(
       implicit F: Async[IO]): Stream[IO, ProjectedFlexibleDataModelInstance] =
     Stream.eval {
-      syncWithCursor(syncRequest).map { sr =>
+      fetchData(futureItemsCursor.nonEmpty, cursors, `with`, select).map { sr =>
```
dmivankov, line 208:
```suggestion
      fetchData(executeBackFill, cursors, `with`, select).map { sr =>
```

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -110,54 +110,114 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
+      val itemDefinitions = qr.items.getOrElse(Map.empty).getOrElse("sync", Vector.empty)
+      val nextCursor = qr.nextCursor.getOrElse(Map.empty).get("sync")
+      ItemsWithCursor(itemDefinitions, nextCursor)
+    }
   }
-  // scalastyle:on cyclomatic.complexity
 
-  private def syncOut(syncRequest: InstanceSyncRequest, selectedProps: Array[String])(
+  private def syncOut(
+      futureItemsCursor: Option[String],
```
dmivankov, line 201:
```suggestion
      executeBackFill: Boolean,
      futureItemsCursor: Option[String],
```

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
dmivankov: But anyway, will take this mental model:
- cursors coming into the datasource and out of it are /sync cursors
- first time run == input cursors aren't set
- on first run we fetch via /query but first stash a /sync cursor that will be returned and used for the next run

dmivankov: additionally
- whenever input cursor is expired, it is also going as if it wasn't set at all aka via /query

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -110,54 +110,120 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
+      select: Map[String, SelectExpression],
+      selectedProps: Array[String])(
       implicit F: Async[IO]): Stream[IO, ProjectedFlexibleDataModelInstance] =
     Stream.eval {
-      syncWithCursor(syncRequest).map { sr =>
+      fetchData(futureItemsCursor.nonEmpty, cursors, `with`, select).map { sr =>
         val items = sr.items
         val nextCursor = sr.nextCursor
         val next =
           (nextCursor, items.nonEmpty && items.last.lastUpdatedTime < terminationTimeStamp) match {
```
dmivankov, line 218: here we have a potential problem, if we decided to do a backfill but then decide to terminate based on this condition we may miss some items, even though terminationTimeStamp is set to now() and is close to futureCursor, but

- we don't limit /query endpoint and it may fetch recently updated items here, and they may also potentially appear at items.last position, as soon as we hit it we stop the iteration and assume nothing is left up to futureCursor, while we might have missed several remaining pages

So it seems what we need to do is always sync to the end in /query case as there's no contract of resuming /query iteration across runs

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -110,54 +110,148 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
+      cursors: Option[Map[String, String]],
+      fetchData: Option[Map[String, String]] => IO[ItemsWithCursor[InstanceDefinition]],
+      projectInstance: Option[String] => InstanceDefinition => ProjectedFlexibleDataModelInstance,
+      applyTimestampTermination: Boolean
+  ): Stream[IO, ProjectedFlexibleDataModelInstance] =
+    Stream.eval {
+      fetchData(cursors).map { sr =>
+        val items = sr.items
+        val nextCursor = sr.nextCursor
+        val shouldStopEarly = items.isEmpty || (applyTimestampTermination && items.last.lastUpdatedTime >=
```
dmivankov, line 243: I think we may still have a problem here if /query may return empty page but non-empty next cursor and mean that we should page more (though somewhat likely that fdm api never returns an empty page with nextCursor present)

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
dmivankov: yeah, I think there was a bug, did some refactoring

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
dmivankov: applied a fix here

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -110,54 +118,144 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
+      useQueryEndpoint = false,
+      cursors = cursors,
+      `with` = Map("sync" -> generateTableExpression(instanceType, matchNothingFilter)),
+      select = select)
+      .flatMap { sr =>
+        (sr.nextCursor, cursors) match {
+          case (None, _) =>
+            IO.raiseError(
+              new IllegalArgumentException("/sync response must have " +
+                "nextCursor"))
```
dmivankov, line 161: https://github.com/cognitedata/cognite-sdk-scala/pull/730 to make /sync nextCursor non-required

dmivankov (approved):
After couple tweaks I think I see it doing the transparent fallback to /query based backfill and continuation with /sync next time. There could be further prettification improvements or sdk fixes, but looks ok enough atm, though would be good if you could go through open comments and code after tweaks to verify it still reflects the intended behavior

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
kornelione: ah.. for backfill we must not stop early I think.. 
We accept seeing duplicates on the next session..

dmivankov: yeah, so now I've added a flag to disable early termination, but it doesn't disable termination on seeing an empty page, not sure what's safer here

dmivankov: another version could be something like
```kotlin
val shouldStopByTimestamp = !isBacFill && items.last.lastUpdatedTime >= abc;
val shouldStopByEmptyPage = !isBackFill && items.isEmpty
val shouldStop = nextCursor.isEmpty || shouldStopByTimestamp || shouldStopByEmptyPage
```

kornelione:
> After couple tweaks I think I see it doing the transparent fallback to /query based backfill and continuation with /sync next time. There could be further prettification improvements or sdk fixes, but looks ok enough atm, though would be good if you could go through open comments and code after tweaks to verify it still reflects the intended behavior

Went over, and also run some of my large data set tests that I have locally. 

Atm. we will still get 408 from DMS on complex datamodels even if we are in the sync-phase

PR #916, opened 2024-03-25
Submit last (empty) sync cursor to callback

thorkildcognite:
Currently, we return the cursor from the /sync DMS call as part of the rows returned to the Dataframe. This cursor has a max lifespan (currently 3 days), and needs to be refreshed even if there are no rows returned.

As a workaround, for now, we therefore submit the cursor to a configured callback URL when the /sync query returns no entries.

### DIFF
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
@@ -251,18 +251,12 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
         val next =
           (nextCursor, items.isEmpty, shouldStopEarly) match {
             case (Some(cursor), true, _) =>
-              if (syncCursorSaveCallbackUrl.isDefined && jobId.isDefined && cursorName.isDefined) {
-                SyncCursorCallback
-                  .lastCursorCallback(
-                    syncCursorSaveCallbackUrl.get,
-                    cursorName.get,
-                    cursor,
-                    jobId.get
-                  )
-                  .unsafeRunSync()
-              }
+              saveLastCursor(cursor)
               fs2.Stream.empty
-            case (Some(cursor), _, false) =>
+            case (Some(cursor), false, true) =>
+              saveLastCursor(cursor)
+              fs2.Stream.empty
+            case (Some(cursor), false, false) =>
               syncOut(Some(Map("sync" -> cursor)), fetchData, projectInstance, applyTimestampTermination)
             case _ => fs2.Stream.empty
           }
@@ -272,4 +266,15 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
       }
     }.flatten
 
+  private def saveLastCursor(cursor: String) =
+    if (syncCursorSaveCallbackUrl.isDefined && jobId.isDefined && cursorName.isDefined) {
+      SyncCursorCallback
+        .lastCursorCallback(
+          syncCursorSaveCallbackUrl.get,
+          cursorName.get,
+          cursor,
+          jobId.get
+        )
+        .unsafeRunSync()
+    }
 }
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
@@ -252,12 +252,14 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
           (nextCursor, items.isEmpty, shouldStopEarly) match {
             case (Some(cursor), true, _) =>
               if (syncCursorSaveCallbackUrl.isDefined && jobId.isDefined && cursorName.isDefined) {
-                SyncCursorCallback.lastCursorCallback(
-                  syncCursorSaveCallbackUrl.get,
-                  cursorName.get,
-                  cursor,
-                  jobId.get
-                )
+                SyncCursorCallback
+                  .lastCursorCallback(
+                    syncCursorSaveCallbackUrl.get,
+                    cursorName.get,
+                    cursor,
+                    jobId.get
+                  )
+                  .unsafeRunSync()
               }
               fs2.Stream.empty
             case (Some(cursor), _, false) =>
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
@@ -232,15 +232,22 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
     }
     val projectInstance = (nextCursor: Option[String]) =>
       toProjectedInstance(_, toProjectedCursor(nextCursor), selectedProps)
+    val projectFinalCursor = (nextCursor: Option[String]) => toProjectedCursor(nextCursor)
     val dataFetcher = (cursors: Option[Map[String, String]]) =>
       fetchData(useQueryEndpoint = isBackFill, cursors, `with`, select)
-    syncOut(cursors, dataFetcher, projectInstance, applyTimestampTermination = !isBackFill)
+    syncOut(
+      cursors,
+      dataFetcher,
+      projectInstance,
+      projectFinalCursor,
+      applyTimestampTermination = !isBackFill)
   }
 
   private def syncOut(
       cursors: Option[Map[String, String]],
       fetchData: Option[Map[String, String]] => IO[ItemsWithCursor[InstanceDefinition]],
       projectInstance: Option[String] => InstanceDefinition => ProjectedFlexibleDataModelInstance,
+      projectFinalCursor: Option[String] => Option[String],
       applyTimestampTermination: Boolean
   ): Stream[IO, ProjectedFlexibleDataModelInstance] =
     Stream.eval {
@@ -250,14 +257,19 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
         val shouldStopEarly = items.nonEmpty && applyTimestampTermination && items.last.lastUpdatedTime >= terminationTimeStamp
         val next =
           (nextCursor, items.isEmpty, shouldStopEarly) match {
-            case (Some(cursor), true, _) =>
-              saveLastCursor(cursor)
+            case (Some(_), true, _) =>
+              saveLastCursor(projectFinalCursor(nextCursor))
               fs2.Stream.empty
-            case (Some(cursor), false, true) =>
-              saveLastCursor(cursor)
+            case (Some(_), false, true) =>
+              saveLastCursor(projectFinalCursor(nextCursor))
               fs2.Stream.empty
             case (Some(cursor), false, false) =>
-              syncOut(Some(Map("sync" -> cursor)), fetchData, projectInstance, applyTimestampTermination)
+              syncOut(
+                Some(Map("sync" -> cursor)),
+                fetchData,
+                projectInstance,
+                projectFinalCursor,
+                applyTimestampTermination)
             case _ => fs2.Stream.empty
           }
         val projection = projectInstance(nextCursor)
@@ -266,15 +278,12 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
       }
     }.flatten
 
-  private def saveLastCursor(cursor: String) =
-    if (syncCursorSaveCallbackUrl.isDefined && jobId.isDefined && cursorName.isDefined) {
-      SyncCursorCallback
-        .lastCursorCallback(
-          syncCursorSaveCallbackUrl.get,
-          cursorName.get,
-          cursor,
-          jobId.get
-        )
-        .unsafeRunSync()
+  private def saveLastCursor(cursorValue: Option[String]) =
+    (syncCursorSaveCallbackUrl, jobId, cursorName, cursorValue) match {
+      case (Some(syncCursorSaveCallbackUrl), Some(jobId), Some(cursorName), Some(cursorValue)) =>
+        SyncCursorCallback
+          .lastCursorCallback(syncCursorSaveCallbackUrl, cursorName, cursorValue, jobId)
+          .unsafeRunSync()
+      case _ =>
     }
 }
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
@@ -255,22 +255,18 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
         val items = sr.items
         val nextCursor = sr.nextCursor
         val shouldStopEarly = items.nonEmpty && applyTimestampTermination && items.last.lastUpdatedTime >= terminationTimeStamp
+        val shouldStop = items.isEmpty || shouldStopEarly // || nextCursor.isEmpty is handled by matching
         val next =
-          (nextCursor, items.isEmpty, shouldStopEarly) match {
-            case (Some(_), true, _) =>
-              saveLastCursor(projectFinalCursor(nextCursor))
-              fs2.Stream.empty
-            case (Some(_), false, true) =>
-              saveLastCursor(projectFinalCursor(nextCursor))
-              fs2.Stream.empty
-            case (Some(cursor), false, false) =>
+          (nextCursor, shouldStop) match {
+            case (Some(cursor), false) =>
               syncOut(
                 Some(Map("sync" -> cursor)),
                 fetchData,
                 projectInstance,
                 projectFinalCursor,
                 applyTimestampTermination)
-            case _ => fs2.Stream.empty
+            case (None, _) | (_, true) =>
+              fs2.Stream.exec(saveLastCursor(projectFinalCursor(nextCursor)))
           }
         val projection = projectInstance(nextCursor)
         val projected = items.map(projection)
@@ -278,12 +274,12 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
       }
     }.flatten
 
-  private def saveLastCursor(cursorValue: Option[String]) =
+  private def saveLastCursor(cursorValue: Option[String]): IO[Unit] =
     (syncCursorSaveCallbackUrl, jobId, cursorName, cursorValue) match {
       case (Some(syncCursorSaveCallbackUrl), Some(jobId), Some(cursorName), Some(cursorValue)) =>
         SyncCursorCallback
           .lastCursorCallback(syncCursorSaveCallbackUrl, cursorName, cursorValue, jobId)
-          .unsafeRunSync()
-      case _ =>
+          .void
+      case _ => IO.unit
     }
 }
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
@@ -38,6 +38,9 @@ final case class StreamMode(syncCursors: Map[String, String]) extends SyncMode
 private[spark] class FlexibleDataModelCorePropertySyncRelation(
     cursor: String,
     config: RelationConfig,
+    cursorName: Option[String],
+    jobId: Option[String],
+    syncCursorSaveCallbackUrl: Option[String],
     corePropConfig: ViewSyncCorePropertyConfig)(sqlContext: SQLContext)
     extends FlexibleDataModelCorePropertyRelation(
       config,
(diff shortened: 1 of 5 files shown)


thorkildcognite:
There is no test for this in this repository now, but a test covers it in jetfire. Will try to add a self-contained test here too, but had issues with auth on OID.

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
```diff
@@ -0,0 +1,88 @@
...
+      .mapResponse {
+        // We do not fail the whole transformation on a non-submitted cursor, as it is
+        // not critical for this job. Multiple failed runs so that the customer exceeds
+        // its max age, will lead to a fallback to restart.
+        case Left(_) => IncrementalCursorResponse(cursorName, cursorValue, updated = false)
+        case Right(value) => value
+      }
+      .send(sttpBackend)
+      .unsafeRunSync()
+      .body
```
dmivankov, line 86:
```suggestion
      .map(_.body)
```
It'd be nice to keep effect types for our internal things and only resort to unsafe escape/execute path where we have to like being inside a spark callback

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  implicit val incrementalCursorCodec: Codec[IncrementalCursor] = deriveCodec[IncrementalCursor]
+  implicit val incrementalCursorResponseCodec: Codec[IncrementalCursorResponse] =
+    deriveCodec[IncrementalCursorResponse]
+
+  def lastCursorCallback(
+      callbackUrl: String,
+      cursorName: String,
+      cursorValue: String,
+      jobId: String): IncrementalCursorResponse = {
```
dmivankov, line 64:
```suggestion
      jobId: String): IO[IncrementalCursorResponse] = {
```

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
```diff
@@ -0,0 +1,88 @@
...
+import java.util.concurrent.Executors
+import scala.concurrent.ExecutionContext
+
+case class IncrementalCursor(name: String, value: String)
+case class IncrementalCursorResponse(name: String, value: String, updated: Boolean)
+
+object SyncCursorCallback {
+  @transient private val logger = getLogger
+  @transient private lazy val sttpBackend: SttpBackend[IO, Any] =
+    new GzipBackend[IO, Any](AsyncHttpClientCatsBackend.usingClient(SttpClientBackendFactory.create()))
```
dmivankov, line 26: will need to check, but maybe we could reuse a common one or move this one to the others (in CdpConnector?)
additionally this default one doesn't have retries iirc

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -244,11 +247,20 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
         val next =
-          (nextCursor, shouldStopEarly) match {
-            case (Some(cursor), false) =>
+          (nextCursor, items.isEmpty, shouldStopEarly) match {
+            case (Some(cursor), true, _) =>
+              if (syncCursorSaveCallbackUrl.isDefined && jobId.isDefined && cursorName.isDefined) {
+                SyncCursorCallback.lastCursorCallback(
+                  syncCursorSaveCallbackUrl.get,
+                  cursorName.get,
+                  cursor,
```
dmivankov, line 258: here it could be a /query cursor in backfill case, so I guess the short ways to fix is to along with
```scala
// pagingCursor -> item -> itemWithSyncCursor (either futureSyncCursor or == pagingCursor)
projectInstance: Option[String] => InstanceDefinition => ProjectedFlexibleDataModelInstance,
```
parameter add
```scala
// pagingCursorOnFinalEmptyPage -> callbackSyncCursor (sync/not backfill -> last sync cursor and it also won't be empty, backfill -> futureSyncCursor)
pickFinalCursor: Option[String] => String
```

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -244,11 +247,20 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
         val nextCursor = sr.nextCursor
-        val shouldStopEarly = items.isEmpty || (applyTimestampTermination && items.last.lastUpdatedTime >=
-          terminationTimeStamp)
+        val shouldStopEarly = items.nonEmpty && applyTimestampTermination && items.last.lastUpdatedTime >= terminationTimeStamp
         val next =
-          (nextCursor, shouldStopEarly) match {
-            case (Some(cursor), false) =>
+          (nextCursor, items.isEmpty, shouldStopEarly) match {
+            case (Some(cursor), true, _) =>
+              if (syncCursorSaveCallbackUrl.isDefined && jobId.isDefined && cursorName.isDefined) {
```
dmivankov, line 254: here it won't be called for backfill case as last page cursor will be empty, we probably want to pass the futureSyncCursor here

dmivankov: biting the bullet and splitting/wrapping the cursor types between /sync and /query might've helped to catch that stuff

dmivankov, line 254: for little more fancy unpacking of several options might also do
```scala
(syncCursorCallbackUrl, jobId, cursorName) match {
  case (Some(syncCursorCallbackUrl), Some(jobId), Some(cursorName)) =>
}
```
or better yet create a small wrapper class and constructor/factory method for it
```scala
case class SyncCursorCallbackSettings(callbackUrl: String, jobId: String, cursorName: String)
```
about so

dmivankov: note that /query endpoint may currently return `nextCursor:{}` https://github.com/cognitedata/service-contracts/pull/2468#issuecomment-2017401876 but we'll go and try extract `val nextCursor = qr.nextCursor.getOrElse(Map.empty).get("sync")` "sync" entry and that will be empty on last /query page

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
thorkildcognite: Switched to return the wrapped IO and moved the call outside.

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
thorkildcognite: On purpose not to have retries actually. Keep it as simple as it can be, as the downside of not committing the cursor is limited.

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
thorkildcognite: We don't want to persist the backfill cursor, only those coming from /sync. The backfill ends up giving you that at the end.

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -244,11 +247,22 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
       fetchData(cursors).map { sr =>
         val items = sr.items
         val nextCursor = sr.nextCursor
-        val shouldStopEarly = items.isEmpty || (applyTimestampTermination && items.last.lastUpdatedTime >=
-          terminationTimeStamp)
+        val shouldStopEarly = items.nonEmpty && applyTimestampTermination && items.last.lastUpdatedTime >= terminationTimeStamp
         val next =
-          (nextCursor, shouldStopEarly) match {
-            case (Some(cursor), false) =>
+          (nextCursor, items.isEmpty, shouldStopEarly) match {
```
kornelione, line 252: I think we miss a case for (Some(cursor), false, true) ?

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
dmivankov: It doesn't, backfill cursor gets injected on per-item level while cursors remain the same matching the iteration method
https://github.com/cognitedata/cdp-spark-datasource/blob/master/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala#L230
https://github.com/cognitedata/cdp-spark-datasource/blob/master/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala#L193

dmivankov: So we need to pick up futureSync cursor like for items when in isBackfill mode

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
dmivankov: any client should have at least some retries, there's no benefit in losing the call when there's a cheap chance to not lose it (and the call should be idempotent)

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
dmivankov: Yeah, for early /sync termination we go to `_` case that also returns empty stream like the last page but callbackUrl call is currently missing there :100:

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -244,11 +247,22 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
+            case (Some(cursor), true, _) =>
+              if (syncCursorSaveCallbackUrl.isDefined && jobId.isDefined && cursorName.isDefined) {
+                SyncCursorCallback
+                  .lastCursorCallback(
+                    syncCursorSaveCallbackUrl.get,
+                    cursorName.get,
+                    cursor,
+                    jobId.get
+                  )
+                  .unsafeRunSync()
```
dmivankov, line 262: There's also `Stream.eval()` to go from IO to Stream[IO, _]

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
```diff
@@ -0,0 +1,87 @@
...
+          .setNameFormat("Cursor-Callback-%d")
+          .build()
+      )
+    )
+
+  @transient private lazy val (blocking, _) =
+    IORuntime.createDefaultBlockingExecutionContext("Cursor-Callback-blocking")
+  @transient private lazy val (scheduler, _) =
+    IORuntime.createDefaultScheduler("Cursor-Callback-scheduler")
+  @transient lazy implicit val ioRuntime: IORuntime = IORuntime(
```
dmivankov, line 48: I think this will no longer be used, if the thread shift is really needed there's https://typelevel.org/cats-effect/docs/thread-model#context-shift

dmivankov: or maybe it will be used to create the sttp client :thinking:

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
thorkildcognite: The downside with the match is that it has to be complete, so you need a "case => " at the end. If not you will end up with scala.MatchError . This is why some don't use match when you have a single case. In some cases, such as transform in Spark, there is an implicit default case, so there it isn't a problem, but for normal code, this is not the case.

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
thorkildcognite: You are right! This removed a lot of code too now.

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
thorkildcognite: So, this was actually on purpose, but might still be wrong? When we hit the shouldStopEarly, we will *not* save the items out. When this happens, we also do not want to save out the cursor, as we want the next run to actually use the previous cursor to get to this again.

Else we would lose the last page, wouldn't we?

thorkildcognite: Ah, this takes care of this by only sending the sync cursor, not the query cursor, so then we're fine. It should save the last cursor, although it will have persisted that already if there was any data. If there was no data to backfill, then it would not work, and it would always trigger backfill.

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
dmivankov: yep, it'd need a default case but upside is not calling partial `.get` functions (ideally there should be a linter to check if .get is guarded by emptiness check as scala compiler won't do that)

There's also stuff like `PartialFunction` but probably won't save much. Or quite specialized `Apply[Option].map3`
```scala
              Apply[Option].map3(syncCursorSaveCallbackUrl, jobId, cursorName){
                case (syncCursorSaveCallbackUrl, jobId, cursorName) => {
                  
                }
              }
```
also a bit too complex

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
thorkildcognite: And since "nextCursor" is the cursor of the page we actually emit right now (project later on), this is exactly the next page we want to continue on.

So, yes, adding the last variation is necessary. It also fixes the always-trigger-backfill issue.

dmivankov: With `shouldStopEarly` we try to save current page items if any.

I think we need to clarify when is the callback expected to fire
- always fire at the end of stream just for good measure
- only fire if the stream is empty
- some other combination of conditions

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
thorkildcognite: Indeed, we need to pick up the projected cursor, not the nextCursor.

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
thorkildcognite: I've redone this slight now.
The thing that got me is that this is "next" cursor for the page we are on, but even if we stop early, we still persist the page we are on. Thus the nextCursor is always the next cursor.

We should:
- always persist when we have no more entries -- also when there was none at all (this is the problem that started this)
- always persist the initial sync-cursor when doing backfill (not the query cursor one). This will only be needed if we have had 0 results, as if we have had some, then the cursor would already be stored, but if we haven't, then we would never get out of backfill mode. 

Realistic this bcomes "always fires" except when we don't have a cursor. Now, we should always have a cursor, but ... well.. avoiding the match error.

kornelione (approved):

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
dmivankov: Adding couple suggestions to https://github.com/cognitedata/cdp-spark-datasource/pull/917
I think /query case was still missing the save step in some cases, or rather almost all cases as it was only saving if nextCursor existed on last page but it won't be there.

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
```diff
@@ -0,0 +1,51 @@
...
+      .contentType(MediaType.ApplicationJson)
+      .acceptEncoding(MediaType.ApplicationJson.toString())
+      .body(IncrementalCursor(cursorName, cursorValue))
+      .post(uri)
+      .response(asJson[IncrementalCursorResponse])
+      .mapResponse {
+        // We do not fail the whole transformation on a non-submitted cursor, as it is
+        // not critical for this job. Multiple failed runs so that the customer exceeds
+        // its max age, will lead to a fallback to restart.
+        case Left(_) => IncrementalCursorResponse(cursorName, cursorValue, updated = false)
```
dmivankov, line 45: It'd be nice to avoid fake values and handle errors on the caller side, especially if we're going to handle retries.
Also would be good to log errors

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
thorkildcognite: I've now switched to using the retrying backend to have retries.

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
thorkildcognite: Fixed by introducing the projectFinalCursor.

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
thorkildcognite: I think it covered all cases in the last version, but your suggestion was much more readable and less complex (fewer branches). Pulled in and I think we're good on this now.

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
```diff
@@ -0,0 +1,51 @@
...
+import sttp.client3.asynchttpclient.cats.AsyncHttpClientCatsBackend
+import sttp.client3.circe.{asJson, circeBodySerializer}
+import sttp.model.{MediaType, Uri}
+
+case class IncrementalCursor(name: String, value: String)
+case class IncrementalCursorResponse(name: String, value: String, updated: Boolean)
+
+object SyncCursorCallback {
+  @transient private lazy val sttpBackend: SttpBackend[IO, Any] =
+    new GzipBackend[IO, Any](AsyncHttpClientCatsBackend.usingClient(SttpClientBackendFactory.create()))
```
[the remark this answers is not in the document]
thorkildcognite: Yeah, that went out the window when I switched backend. Adding prefix control in my last commit now.

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
```diff
@@ -0,0 +1,59 @@
...
+  implicit val incrementalCursorResponseCodec: Codec[IncrementalCursorResponse] =
+    deriveCodec[IncrementalCursorResponse]
+
+  def lastCursorCallback(
+      callbackUrl: String,
+      cursorName: String,
+      cursorValue: String,
+      jobId: String): IO[IncrementalCursorResponse] = {
+    val uri = Uri.parse(s"$callbackUrl/$jobId") match {
+      case Left(value) => throw new IllegalArgumentException(s"Failed to parse URI '$value'")
```
dmivankov, line 39: here we should decide what to do with uri parse errors
- burn -> `IO.raiseError(exception here)` or `IO.fromEither(uriparse.left.map(string to exception))` and then `uriio.flatMap(uri => basicRequest...)` or `for {uri <- urlio; response <- basicrequest... } yield ...`
- lift to `IO[Either[someerrortype, IncrementalCursorResponse]]` to present it as if it was a request/response error

dmivankov: another option is to parse/filter the callbackUri early and have it as URI type already

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
```diff
@@ -0,0 +1,60 @@
...
+
+case class IncrementalCursor(name: String, value: String)
+case class IncrementalCursorResponse(name: String, value: String, updated: Boolean)
+
+object SyncCursorCallback {
+  @transient private lazy val sttpBackend: SttpBackend[IO, Any] =
+    new GzipBackend[IO, Any](
+      AsyncHttpClientCatsBackend.usingClient(SttpClientBackendFactory.create("Last-Cursor-Submitter")))
+
+  @transient private val retryingBackend = new RetryingBackend[IO, Any](
```
dmivankov, line 22:
```suggestion
  @transient private lazy val retryingBackend = new RetryingBackend[IO, Any](
```
otherwise it'd force the other lazy field anyway

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
thorkildcognite: We log the cursor update, but I agree it was maybe a bit too restrictive to not log anything. I've added it now.

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    basicRequest
+      .followRedirects(false)
+      .contentType(MediaType.ApplicationJson)
+      .acceptEncoding(MediaType.ApplicationJson.toString())
+      .body(IncrementalCursor(cursorName, cursorValue))
+      .post(uri)
+      .response(asJson[IncrementalCursorResponse])
+      .mapResponse {
```
dmivankov, line 53: a little unsure of the ordering or layering between `.mapResponse` and `RetryingBackend.send` so yeah, should be simpler to ignore errors on caller level

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
thorkildcognite: The URL is coming from the dataframe setup, though, so I would say this isn't really that much needed.

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
thorkildcognite: Indeed. Merged.

src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
dmivankov: If there's a error on putting that from dataframe setup side, it should be made visible somehow

PR #918, opened 2024-03-26
Enable syncing/cursors with is_new when resolving view through data model

kornelione:


thorkildcognite (approved):
LGTM - had to review it in a local Intellij after getting a bit confused by the renamed/moved createCorePropertyRelationForDataModel. Sometimes diff-es can be quite confusing.

Confirmed this version works with jetfire-backend to resolve the issue with datamodel not being supported.

Let's merge this, bump the dep in the jetfire-backend and push it out asap!

PR #919, opened 2024-03-27
Avoid ParallelTestExecution for sequences

thorkildcognite:
apparently it is making conflicts for itself, leading to the test taking 25+ minutes (!).

### DIFF
--- a/src/test/scala/cognite/spark/v1/SequenceRowsRelationTest.scala
+++ b/src/test/scala/cognite/spark/v1/SequenceRowsRelationTest.scala
@@ -4,9 +4,9 @@ import cognite.spark.v1.CdpConnector.ioRuntime
 import com.cognite.sdk.scala.v1.{Sequence, SequenceColumnCreate}
 import io.scalaland.chimney.dsl._
 import org.apache.spark.sql.{DataFrame, Row}
-import org.scalatest.{FlatSpec, Matchers, ParallelTestExecution}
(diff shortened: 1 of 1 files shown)


src/test/scala/cognite/spark/v1/SequenceRowsRelationTest.scala
```diff
@@ -4,9 +4,9 @@ import cognite.spark.v1.CdpConnector.ioRuntime
 import com.cognite.sdk.scala.v1.{Sequence, SequenceColumnCreate}
 import io.scalaland.chimney.dsl._
 import org.apache.spark.sql.{DataFrame, Row}
-import org.scalatest.{FlatSpec, Matchers, ParallelTestExecution}
+import org.scalatest.{FlatSpec, Matchers}
 
-class SequenceRowsRelationTest extends FlatSpec with Matchers with ParallelTestExecution with SparkTest {
+class SequenceRowsRelationTest extends FlatSpec with Matchers with SparkTest {
```
dmivankov, line 9: the issue it has is probably coming from hardcoded field constants and/or not trying to focus on a subset of data, randomizing it (also trick is to make sure random seed isn't copied to all parallel executors, don't remember exactly how to do it the easiest way) may help keeping parallellism, but short-term let's just hammer it :+1:

dmivankov (approved):

kornelione (approved):

PR #920, opened 2024-03-27
Resolve connection before sync

kornelione:


thorkildcognite (approved):
I don't think this is this specific issue, but it is a valid point that it may match the wrong one first. 

But, I think it also means that we do not support sync on cdf_data_models-with-connection-type relation.

Jacob-Eliat-Eliat:
Extra context: I figured out, I think what you though the problem was here:

 * You though that the connection could be fetched and in that case it wouldn't go through sync endpoint.
 * You though that it selected the wrong one
 
In reality, connection relations are only used for destination relation. There is not tvf where you can set the edge's type in the tvf. If you want to select edges by type, you need to do

"select * from cdf_edges() where _type = ..." in which case it uses `corePropertyRelation` instead of `connectionRelation`

More precisely, in the plan's unapply, we do

```
      case UnresolvedTableValuedFunction(FunctionIdentifier("cdf_edges", _), params, _)
          if instancesRelation(InstancesRelationType.edges).isDefinedAt(params) =>
        Some(instancesRelation(InstancesRelationType.edges)(params))
```
but it is not defined with parameters "edgeType..." so if you try to pass that it won't work. There is thus no way to fetch instances by edge type and to get a connectionRelation, therefore it doesn't need to be handled by sync.

PR #924, opened 2024-04-09
Celo/cdf 21249/attempt number metrics

mr-celo:
### Add attempt number to metric names

By adding the attempt number to the metrics it will be possible to get
the highest attempt for each stage and task and allowing ignoring
metrics coming from failed/retried tasks. A good example would be the
effective number of rows added.

Though, the metrics don't allow labels, and instead uses dot-delimited
fields such as job id.
By keeping always a consistent number of fields, these can be extracted
while aggregating the metrics. Therefore, it is the responsibility of
the MetricSource to enforce this consistency.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -46,7 +46,13 @@ lazy val commonSettings = Seq(
   description := "Spark data source for the Cognite Data Platform.",
   licenses := List("Apache 2" -> new URL("http://www.apache.org/licenses/LICENSE-2.0.txt")),
   homepage := Some(url("https://github.com/cognitedata/cdp-spark-datasource")),
-  scalacOptions ++= Seq("-Xlint:unused", "-language:higherKinds", "-deprecation", "-feature"),
+  scalacOptions ++= Seq("-Xlint:unused", "-language:higherKinds", "-deprecation", "-feature") ++ (CrossVersion.partialVersion(scalaVersion.value) match {
+    // We use JavaConverters to remain backwards compatible with Scala 2.12,
+    // and to avoid a dependency on scala-collection-compat
+    case Some((2, 13)) => Seq("-Wconf:src=src/main/scala/cognite/spark/v1/MetricsSource.scala&cat=deprecation:i")
+    case _ => Seq.empty
+  }),
+  // scalacOptions ++= Seq("-Xlint:unused", "-language:higherKinds", "-deprecation", "-feature", "-Wconf:src=src/main/scala/cognite/spark/v1/MetricsSource.scala&cat=deprecation:i"),
   resolvers ++= Resolver.sonatypeOssRepos("releases"),
   developers := List(
     Developer(
--- a/build.sbt
+++ b/build.sbt
@@ -52,7 +52,6 @@ lazy val commonSettings = Seq(
     case Some((2, 13)) => Seq("-Wconf:src=src/main/scala/cognite/spark/v1/MetricsSource.scala&cat=deprecation:i")
     case _ => Seq.empty
   }),
-  // scalacOptions ++= Seq("-Xlint:unused", "-language:higherKinds", "-deprecation", "-feature", "-Wconf:src=src/main/scala/cognite/spark/v1/MetricsSource.scala&cat=deprecation:i"),
   resolvers ++= Resolver.sonatypeOssRepos("releases"),
   developers := List(
     Developer(
--- a/build.sbt
+++ b/build.sbt
@@ -49,7 +49,7 @@ lazy val commonSettings = Seq(
   scalacOptions ++= Seq("-Xlint:unused", "-language:higherKinds", "-deprecation", "-feature") ++ (CrossVersion.partialVersion(scalaVersion.value) match {
     // We use JavaConverters to remain backwards compatible with Scala 2.12,
     // and to avoid a dependency on scala-collection-compat
-    case Some((2, 13)) => Seq("-Wconf:src=src/main/scala/cognite/spark/v1/MetricsSource.scala&cat=deprecation:i")
+    case Some((2, 13)) => Seq("-Wconf:src=src/test/scala/cognite/spark/v1/SparkTest.scala&cat=deprecation:i")
     case _ => Seq.empty
   }),
   resolvers ++= Resolver.sonatypeOssRepos("releases"),
--- a/src/main/scala/cognite/spark/v1/CdfRelation.scala
+++ b/src/main/scala/cognite/spark/v1/CdfRelation.scala
@@ -12,20 +12,18 @@ abstract class CdfRelation(config: RelationConfig, shortNameStr: String)
     extends BaseRelation
     with Serializable {
   protected val shortName: String = shortNameStr
-  @transient lazy protected val itemsRead: Counter =
-    MetricsSource.getOrCreateCounter(config.metricsPrefix, s"$shortName.read")
-  @transient lazy protected val itemsCreated: Counter =
-    MetricsSource.getOrCreateCounter(config.metricsPrefix, s"$shortName.created")
-  @transient lazy protected val itemsUpdated: Counter =
-    MetricsSource.getOrCreateCounter(config.metricsPrefix, s"$shortName.updated")
-  @transient lazy protected val itemsDeleted: Counter =
-    MetricsSource.getOrCreateCounter(config.metricsPrefix, s"$shortName.deleted")
+  protected def itemsRead: Counter = getOrCreateCounter("read")
+  protected def itemsCreated: Counter = getOrCreateCounter("created")
+  protected def itemsUpdated: Counter = getOrCreateCounter("updated")
+  protected def itemsDeleted: Counter = getOrCreateCounter("deleted")
   // We are not aware if it is `created` or `updated in flexible data modelling case
-  @transient lazy protected val itemsUpserted: Counter =
-    MetricsSource.getOrCreateCounter(config.metricsPrefix, s"$shortName.upserted")
+  protected def itemsUpserted: Counter = getOrCreateCounter("upserted")
 
   @transient lazy val client: GenericClient[IO] =
     CdpConnector.clientFromConfig(config)
+
+  private def getOrCreateCounter(action: String): Counter =
+    MetricsSource.getOrCreateCounter(config.metricsPrefix, s"$shortName.$action")
 
   def incMetrics(counter: Counter, count: Int): IO[Unit] =
     IO(
--- a/src/main/scala/cognite/spark/v1/CdfRelation.scala
+++ b/src/main/scala/cognite/spark/v1/CdfRelation.scala
@@ -5,6 +5,7 @@ import com.codahale.metrics.Counter
 import com.cognite.sdk.scala.common.{NonNullableSetter, SetNull, SetValue, Setter}
 import com.cognite.sdk.scala.v1._
 import io.scalaland.chimney.Transformer
+import org.apache.spark.TaskContext
 import org.apache.spark.datasource.MetricsSource
 import org.apache.spark.sql.sources.BaseRelation
 
@@ -12,18 +13,22 @@ abstract class CdfRelation(config: RelationConfig, shortNameStr: String)
     extends BaseRelation
     with Serializable {
   protected val shortName: String = shortNameStr
-  protected def itemsRead: Counter = getOrCreateCounter("read")
-  protected def itemsCreated: Counter = getOrCreateCounter("created")
-  protected def itemsUpdated: Counter = getOrCreateCounter("updated")
-  protected def itemsDeleted: Counter = getOrCreateCounter("deleted")
+
+  private def getOrCreateCounter(action: String): Counter =
+    MetricsSource.getOrCreateAttemptTrackingCounter(
+      config.metricsPrefix,
+      s"$shortName.$action",
+      Option(TaskContext.get()))
+
+  @transient lazy protected val itemsRead: Counter = getOrCreateCounter("read")
+  @transient lazy protected val itemsCreated: Counter = getOrCreateCounter("created")
+  @transient lazy protected val itemsUpdated: Counter = getOrCreateCounter("updated")
+  @transient lazy protected val itemsDeleted: Counter = getOrCreateCounter("deleted")
   // We are not aware if it is `created` or `updated in flexible data modelling case
-  protected def itemsUpserted: Counter = getOrCreateCounter("upserted")
+  @transient lazy protected val itemsUpserted: Counter = getOrCreateCounter("upserted")
 
   @transient lazy val client: GenericClient[IO] =
     CdpConnector.clientFromConfig(config)
-
-  private def getOrCreateCounter(action: String): Counter =
-    MetricsSource.getOrCreateCounter(config.metricsPrefix, s"$shortName.$action")
 
   def incMetrics(counter: Counter, count: Int): IO[Unit] =
     IO(
--- a/src/main/scala/cognite/spark/v1/MetricsSource.scala
+++ b/src/main/scala/cognite/spark/v1/MetricsSource.scala
@@ -4,13 +4,11 @@ import cats.Eval
 import com.codahale.metrics._
 import java.util.concurrent.ConcurrentHashMap
 import org.apache.spark._
-import org.log4s._
-import scala.collection.JavaConverters._
 
 class MetricsSource {
   // Add metricNamespace to differentiate with spark system metrics.
   // Keeps track of all the Metric instances that are being published
-  val metricsMap = new ConcurrentHashMap[String, Eval[SourceCounter]]
+  val metricsMap = new ConcurrentHashMap[String, Eval[Counter]]
 
   def getOrCreateCounter(metricNamespace: String, name: String): Counter = {
     val ctx = Option(TaskContext.get())
@@ -29,14 +27,13 @@ class MetricsSource {
     val key = s"$metricNamespace.$metricName"
 
     val wrapped = Eval.later {
-      getLogger.info(s"Creating and registering counter for $key")
       val counter = new Counter
-      val source = registerMetricSource(metricNamespace, metricName, counter)
-      new SourceCounter(source, counter)
+      registerMetricSource(metricNamespace, metricName, counter)
+      counter
     }
 
     metricsMap.putIfAbsent(key, wrapped)
-    metricsMap.get(key).value.counter
+    metricsMap.get(key).value
   }
 
   /**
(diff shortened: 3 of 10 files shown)


src/main/scala/cognite/spark/v1/MetricsSource.scala
```diff
@@ -43,9 +62,53 @@ class MetricsSource {
...
+    metricsMap.remove(key)
+    SparkEnv.get.metricsSystem.removeSource(source)
   }
 }
 
 // Singleton to make sure each metric is only registered once.
 object MetricsSource extends MetricsSource
+
+class ExpiringCounter(key: String, source: Source) extends Counter {
+  private var (deathTimer, deathTimerTask): (Timer, TimerTask) = createTimer
```
mr-celo, line 81: Unsure if the timer will survive long enough to deregister the metric upon task completion.
This will also not work on a crash of the executor, and the metrics will linger.

Another solution would be to move this to the driver, but the instance of the MetricSource (so that it can be removed from the MetricsSystem) gets a little more complicated. Looking into this approach now, though

src/main/scala/cognite/spark/v1/MetricsSource.scala
```diff
@@ -2,22 +2,39 @@ package org.apache.spark.datasource
...
+    val metricName = s"$stageId.$stageAttempt.$partitionId.$taskAttempt.$name"
+    val key = s"$metricNamespace.$metricName"
 
-  def getOrCreateCounter(metricNamespace: String, metricName: String): Counter = {
     val wrapped = Eval.later {
       val counter = new Counter
-      registerMetricSource(metricNamespace, metricName, counter)
-      counter
+      val source = registerMetricSource(metricNamespace, metricName, counter)
+      new ExpiringCounter(key, source)
```
dmivankov, line 35: feels like it may start timer even if entry is already present, then unadded value will unregister prematurely. Could be that Eval.later introduces some laziness that prevents this, but sounds fragile even if it does

src/main/scala/cognite/spark/v1/MetricsSource.scala
```diff
@@ -2,22 +2,39 @@ package org.apache.spark.datasource
...
+      ctx.map(getter).map(_.toString()).getOrElse("")
+
+    // The number of fields in the name should be consistent, even if there is no value for
+    // the attempt numbers
+    val stageId = getNumberOrEmpty(_.stageId())
+    val stageAttempt = getNumberOrEmpty(_.stageAttemptNumber())
+    val partitionId = getNumberOrEmpty(_.partitionId())
+    val taskAttempt = getNumberOrEmpty(_.taskAttemptId())
+
+    val metricName = s"$stageId.$stageAttempt.$partitionId.$taskAttempt.$name"
```
dmivankov, line 29: it'd change metric name? Then we'll lose metrics under old names?

mr-celo: It will. This will be coupled with driver changes

src/main/scala/cognite/spark/v1/MetricsSource.scala
mr-celo: That is fair, I'd say. The registering of a new metric with the same name can be considered as the metric being "active". Therefore, it will reset the inactivity timer.
Probably fairer than receiving a metric that will die moments later

Or did I misunderstand your concern?

src/main/scala/cognite/spark/v1/MetricsSource.scala
```diff
@@ -43,9 +62,53 @@ class MetricsSource {
...
+    deathTimerTask.cancel()
+    deathTimer.cancel()
+
+    val (timer, task) = createTimer
+    deathTimer = timer
+    deathTimerTask = task
+  }
+
+  override def inc(n: Long): Unit = {
+    resetTimer()
```
dmivankov, line 105: some potential for perf degradation here, before it was more or less simple increment, now it is noticeable work per increment (a little less scary since it'd probably be batch-increment but still). Especially the recreation of Timer and its thread

src/main/scala/cognite/spark/v1/MetricsSource.scala
dmivankov: Splitting the PR in 2 (expiration and per-stage breakdown) + adding feature flags for those would make sense as both changes have their own risks (perf or breaking metrics somehow and doing something to TaskContext that isn't allowed in executors + if anything depended on old metric names)

mr-celo: I broke it down like that in terms of commits, for the same reason.
But I coupled them in a single PR because the number of metrics will increase significantly. So, by not having them together, a new bigger perf risk arises with the great increase of metrics.

But your point is valid. If something on the driver can address the stale metrics (what I'm looking into now), then both should be paired together within a deployment.

src/main/scala/cognite/spark/v1/MetricsSource.scala
dmivankov: If 2 getOrCreateCounter calls create and start 2 counters, only one ends up in the map and receive events, the other will decide nothing happens and will deregister both by metric name

src/main/scala/cognite/spark/v1/MetricsSource.scala
```diff
@@ -43,9 +62,53 @@ class MetricsSource {
...
+class ExpiringCounter(key: String, source: Source) extends Counter {
+  private var (deathTimer, deathTimerTask): (Timer, TimerTask) = createTimer
+
+  private def createTimer = {
+    val task = new TimerTask {
+      override def run =
+        MetricsSource.deregisterMetricSource(key, source)
+    }
+
+    val timer = new Timer(key)
```
dmivankov, line 89: here's a thread leak here. Even when we don't have references to TimerThread it will still be considered a garbage collection root by being an alive/running thread

dmivankov: Or at least it seems so. There's some code in Timer class that tries to stop the thread when Timer becomes garbage-collectible which might be enough to let the thread stop and be collected, but a bit hard to be really sure this chain would work and no other references path keeps it all alive

build.sbt
```diff
@@ -46,7 +46,13 @@ lazy val commonSettings = Seq(
...
   licenses := List("Apache 2" -> new URL("http://www.apache.org/licenses/LICENSE-2.0.txt")),
   homepage := Some(url("https://github.com/cognitedata/cdp-spark-datasource")),
-  scalacOptions ++= Seq("-Xlint:unused", "-language:higherKinds", "-deprecation", "-feature"),
+  scalacOptions ++= Seq("-Xlint:unused", "-language:higherKinds", "-deprecation", "-feature") ++ (CrossVersion.partialVersion(scalaVersion.value) match {
+    // We use JavaConverters to remain backwards compatible with Scala 2.12,
+    // and to avoid a dependency on scala-collection-compat
+    case Some((2, 13)) => Seq("-Wconf:src=src/main/scala/cognite/spark/v1/MetricsSource.scala&cat=deprecation:i")
+    case _ => Seq.empty
+  }),
+  // scalacOptions ++= Seq("-Xlint:unused", "-language:higherKinds", "-deprecation", "-feature", "-Wconf:src=src/main/scala/cognite/spark/v1/MetricsSource.scala&cat=deprecation:i"),
```
dmivankov, line 55:
```suggestion
```

src/main/scala/cognite/spark/v1/MetricsSource.scala
dmivankov: What was the conclusion of doing some of the logic in the driver?

src/main/scala/cognite/spark/v1/MetricsSource.scala
```diff
@@ -2,24 +2,41 @@ package org.apache.spark.datasource
...
+      ctx.map(getter).map(_.toString()).getOrElse("")
+
+    // The number of fields in the name should be consistent, even if there is no value for
+    // the attempt numbers
+    val stageId = getNumberOrEmpty(_.stageId())
+    val stageAttempt = getNumberOrEmpty(_.stageAttemptNumber())
+    val partitionId = getNumberOrEmpty(_.partitionId())
+    val taskAttempt = getNumberOrEmpty(_.taskAttemptId())
+
+    val metricName = s"$stageId.$stageAttempt.$partitionId.$taskAttempt.$name"
```
dmivankov, line 28: On the cardinality front, this can make x200 (raw partitions), x3-5 (task attempts), x3-5 (stage attempts) -> x?? (num stages). And when it gets propagated to prometheus (in driver it does) it still impacts prometheus even after it stops being exported (prometheus still remembers names+labels for the retention period).

Would be good to
- keep old metric for transition period
- for new metrics add a simple prefix so that driver could easily ignore them when exporting to prometheus if it decides to
- add feature flag that allows reducing cardinality back (disable extended metrics completely), increase could be causing too big change in load towards metrics store, potentially, would be good to have a quick switch off
- maybe add feature flags per dimension to disable the breakdown: partition (very deep debugging thing and big multiplier at the same time, first candidate for being disabled), maybe others

Also maybe it is still to difficult to correctly discard metrics for failed/lost and retried data reads/writes, maybe there's some better way to get number of items read/write. Otoh metrics cleanup for completed jobs is a pretty clear improvement which would be nice to get done sooner than retry-aware counting (and so independently of it as it is proving to be not that trivial to get those numbers..)

dmivankov: Following some discussion I'll need to read into what exactly happens with those metrics :eyes: plus when the driver side is ready would be easier to review them together :+1:

mr-celo: * The new driver is backwards compatible, so we can bring this change in an out
* The driver will only send an extra label to prometheus: `completed`

The driver PR is ready: https://github.com/cognitedata/jetfire-backend/pull/2285

src/main/scala/cognite/spark/v1/MetricsSource.scala
```diff
@@ -32,20 +49,64 @@ class MetricsSource {
...
+    source
+  }
+
+  /**
+    * Remove the metrics from a job.
+    *
+    * This method will deregister the metric from Spark's org.apache.spark.metrics.MetricsSystem
+    * and stops tracking that it was published
+    */
+  def removeJobMetrics(metricNamespace: String, jobId: String): Unit = {
```
dmivankov, line 72:
```suggestion
  def unregisterJobMetrics(metricNamespace: String, jobId: String): Unit = {
```
nit: maybe something like that to reflect that it's mostly about in-mem metrics

src/main/scala/cognite/spark/v1/CdfRelation.scala
```diff
@@ -12,21 +12,19 @@ abstract class CdfRelation(config: RelationConfig, shortNameStr: String)
...
-  @transient lazy protected val itemsCreated: Counter =
-    MetricsSource.getOrCreateCounter(config.metricsPrefix, s"$shortName.created")
-  @transient lazy protected val itemsUpdated: Counter =
-    MetricsSource.getOrCreateCounter(config.metricsPrefix, s"$shortName.updated")
-  @transient lazy protected val itemsDeleted: Counter =
-    MetricsSource.getOrCreateCounter(config.metricsPrefix, s"$shortName.deleted")
+  protected def itemsRead: Counter = getOrCreateCounter("read")
+  protected def itemsCreated: Counter = getOrCreateCounter("created")
+  protected def itemsUpdated: Counter = getOrCreateCounter("updated")
+  protected def itemsDeleted: Counter = getOrCreateCounter("deleted")
```
dmivankov, line 18:
```suggestion
  @transient lazy protected val itemsRead: Counter = getOrCreateCounter("read")
  @transient lazy protected val itemsCreated: Counter = getOrCreateCounter("created")
  @transient lazy protected val itemsUpdated: Counter = getOrCreateCounter("updated")
  @transient lazy protected val itemsDeleted: Counter = getOrCreateCounter("deleted")
```
we can still keep it being a field?

src/main/scala/cognite/spark/v1/MetricsSource.scala
```diff
@@ -2,22 +2,36 @@ package org.apache.spark.datasource
...
+import org.apache.spark._
 
 class MetricsSource {
   // Add metricNamespace to differentiate with spark system metrics.
   // Keeps track of all the Metric instances that are being published
   val metricsMap = new ConcurrentHashMap[String, Eval[Counter]]
 
-  def getOrCreateCounter(metricNamespace: String, metricName: String): Counter = {
+  def getOrCreateCounter(metricNamespace: String, name: String): Counter = {
+    val ctx = Option(TaskContext.get())
```
dmivankov, line 14: is that TaskContext coming from a thread-local horror land? :thinking:

src/main/scala/cognite/spark/v1/MetricsSource.scala
```diff
@@ -2,22 +2,36 @@ package org.apache.spark.datasource
...
 import java.util.concurrent.ConcurrentHashMap
+import org.apache.spark._
 
 class MetricsSource {
   // Add metricNamespace to differentiate with spark system metrics.
   // Keeps track of all the Metric instances that are being published
   val metricsMap = new ConcurrentHashMap[String, Eval[Counter]]
 
-  def getOrCreateCounter(metricNamespace: String, metricName: String): Counter = {
+  def getOrCreateCounter(metricNamespace: String, name: String): Counter = {
```
dmivankov, line 13:
```suggestion
  def getOrCreateDetailedCounter(metricNamespace: String, name: String): Counter = {
```
or something like that to indicate that meaning changes a lot while type signature remains the same `(String, String)->Counter`, this would enforce fixing references to it here or for external consumers

src/main/scala/cognite/spark/v1/MetricsSource.scala
```diff
@@ -2,22 +2,36 @@ package org.apache.spark.datasource
...
+    val ctx = Option(TaskContext.get())
+
+    def getNumberOrEmpty(getter: TaskContext => AnyVal): String =
+      ctx.map(getter).map(_.toString()).getOrElse("")
```
dmivankov, line 17:
```suggestion
    def getNumberOrEmpty(getter: TaskContext => String): String =
      ctx.map(getter).getOrElse("")
```
to be more type-safe, at call-sites just add a quick .toString()

src/main/scala/cognite/spark/v1/MetricsSource.scala
dmivankov: if it is then val->def change makes sense but would be better than to move that stuff around, let MetricSource.getCounter take TaskContext as parameter and pass it in CdfRelation and even there maybe also take it as a parameter and require filling it at metric update sites like
```scala
        if (config.collectMetrics) {
          itemsRead.inc()
        }
```
so we definitely have more control on which thread it gets executed on (datasource repo uses cats-effect in various places, to switches between threads are quite likely)

src/main/scala/cognite/spark/v1/MetricsSource.scala
dmivankov: Not sure if there's a common numeric type that we could use to make sure getter gets a (non-floating point) number (Long?)

src/main/scala/cognite/spark/v1/MetricsSource.scala
```diff
@@ -2,22 +2,36 @@ package org.apache.spark.datasource
...
+
+    // The number of fields in the name should be consistent, even if there is no value for
+    // the attempt numbers
+    val stageId = getNumberOrEmpty(_.stageId())
+    val stageAttempt = getNumberOrEmpty(_.stageAttemptNumber())
+    val partitionId = getNumberOrEmpty(_.partitionId())
+    val taskAttempt = getNumberOrEmpty(_.taskAttemptId())
+
+    val metricName = s"<$stageId>[$stageAttempt]#<$partitionId>[$taskAttempt]#$name"
```
dmivankov, line 26: for a big format like that it might be nice to put the reading part also in this lib + a test or two to illustrate the format (normal case, case with empty strings). Also maybe since fields are so many a small data type to represent taskContext+name

Only a little surprised that current tests didn't notice much of a change. But to hypothesize, if a consumer of old metric format sees it it will interpret the whole 5-part thing as "metricName"? Might be especially interesting how they interact in case metrics are persisted and either old or new reader might see both kinds in transition periods.

Also had a look at https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md - exposition format probably doesn't matter a lot as it should encode, but https://prometheus.io/docs/concepts/data_model/ metric name restrictions might be important as not sure what will happen when disallowed prometheus-metric-name characters are used (would they be replaced, converted, or will something fail?). Not seeing an explicit length restriction there. Would feel more safe if we stay within allowed prometheus names, so maybe some variation on `::`? Like `":stage_:$stageId:stagetry_:$stageAttempt:part_:$partitionId:tasktry_:$taskAttempt:$name" ?`

src/main/scala/cognite/spark/v1/MetricsSource.scala
mr-celo: Got it. Yeah, I suppose that can be done. Then the counters can become fields again (referring to the comments above 👆 )

src/main/scala/cognite/spark/v1/MetricsSource.scala
mr-celo: Good point. Will change to not risk having the name change from underneath our feet.
As for the tests, the brunt of it happened it the repo that uses this as a datasource

Will also bring the parser here.

src/main/scala/cognite/spark/v1/MetricsSource.scala
mr-celo: Refactoring time

src/main/scala/cognite/spark/v1/MetricsSource.scala
mr-celo: `Long` should cover them all, yeah 👍 Especially bc this is a local function

src/test/scala/cognite/spark/v1/MetricsSourceTest.scala
```diff
@@ -0,0 +1,76 @@
...
+      }
+    }
+  }
+
+  it should "work with old name format" in {
+    val name = "my.name:goes.here"
+
+    name match {
+      case MetricsSource.metricNameRegex(stageId, stageAttempt, partitionId, taskAttempt, name) => {
+        assert(stageId === null)
```
Jacob-Eliat-Eliat, line 48: Ah I get why it's like this but that's kind of sad :s Have you considered writing a class that would contain the regex and these value and would put them in options with value None? Might not be worth the effort though, but null in scala is meh

Jacob-Eliat-Eliat (approved):
lgtm obviously the lack of null handling by scala's regex is a bit sad but that's a bit of a nitpick since it's java seeping through

src/test/scala/cognite/spark/v1/MetricsSourceTest.scala
mr-celo: Wrapped and handled empty strings as well

PR #929, opened 2024-05-02
Add deregistering of metrics

mr-celo:
Upon job end, the driver may call a method to both clean the tracking hashmap and deregister the source from the MetricsSystem.

The new method will remove all metrics belonging to a given job UUID.

### DIFF
--- a/src/main/scala/cognite/spark/v1/MetricsSource.scala
+++ b/src/main/scala/cognite/spark/v1/MetricsSource.scala
@@ -2,24 +2,28 @@ package org.apache.spark.datasource
 
 import cats.Eval
 import com.codahale.metrics._
-import org.apache.spark._
-
 import java.util.concurrent.ConcurrentHashMap
+import org.apache.spark.SparkEnv
+import org.log4s.getLogger
+import scala.collection.JavaConverters._
 
 class MetricsSource {
   // Add metricNamespace to differentiate with spark system metrics.
   // Keeps track of all the Metric instances that are being published
-  val metricsMap = new ConcurrentHashMap[String, Eval[Counter]]
+  val metricsMap = new ConcurrentHashMap[String, Eval[SourceCounter]]
 
   def getOrCreateCounter(metricNamespace: String, metricName: String): Counter = {
-    val wrapped = Eval.later {
-      val counter = new Counter
-      registerMetricSource(metricNamespace, metricName, counter)
-      counter
-    }
     val key = s"$metricNamespace.$metricName"
+
+    val wrapped = Eval.later {
+      getLogger.info(s"Creating and registering counter for $key")
+      val counter = new Counter
+      val source = registerMetricSource(metricNamespace, metricName, counter)
+      SourceCounter(source, counter)
+    }
+
     metricsMap.putIfAbsent(key, wrapped)
-    metricsMap.get(key).value
+    metricsMap.get(key).value.counter
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/main/scala/cognite/spark/v1/MetricsSource.scala
```diff
@@ -32,20 +36,47 @@ class MetricsSource {
...
+  /**
+    * Remove the metrics from a job.
+    *
+    * This method will deregister the metric from Spark's org.apache.spark.metrics.MetricsSystem
+    * and stops tracking that it was published
+    */
+  def removeJobMetrics(metricNamespace: String, jobId: String): Unit = {
+    val key = s"$metricNamespace.$jobId"
+    val removed =
+      metricsMap.keys().asScala.filter(k => k.startsWith(key)).map(k => (k, metricsMap.remove(k)))
```
mr-celo, line 62: This tuple here is created just for logging purposes. I intend to remove the logging (and simplify this) after a bit of validation

src/main/scala/cognite/spark/v1/MetricsSource.scala
```diff
@@ -32,20 +36,47 @@ class MetricsSource {
...
+    source
+  }
+
+  /**
+    * Remove the metrics from a job.
+    *
+    * This method will deregister the metric from Spark's org.apache.spark.metrics.MetricsSystem
+    * and stops tracking that it was published
+    */
+  def removeJobMetrics(metricNamespace: String, jobId: String): Unit = {
```
dmivankov, line 59:
```suggestion
  def removeJobMetrics(metricNamespace: String): Unit = {
```
I think in datasource and this class metricNamespace is being actually metricsPrefix spark parameter and doesn't need to append jobId. Further hint at that is that jobId wasn't mentioned in this class before

src/main/scala/cognite/spark/v1/MetricsSource.scala
```diff
@@ -32,20 +36,47 @@ class MetricsSource {
...
+
+    if (removed.nonEmpty) {
+      val logger = getLogger
+      val metricsSystem = SparkEnv.get.metricsSystem
+      removed
+        .map({ case (k, v) => (k, v.value.source) })
+        .foreach({
+          case (key, source) => {
+            logger.info(s"Deleting and removing counter for $key")
+            metricsSystem.removeSource(source)
```
dmivankov, line 72: https://github.com/LucaCanali/Miscellaneous/blob/master/Spark_Dashboard/Spark_dropwizard_metrics_info.md
> Spark is instrumented with the Dropwizard/Codahale metrics library. Several components of Spark are instrumented, notable for this work several metrics generating from Spark driver and executors components can be instrumented.
>
> An important architectural details is that the metrics are sent directly from the sources to the sink. This is important when running on a cluster, as each executor will communicate directly to the sink, for better scalability.

So it seems like when we'll call this from driver it will not cleanup metrics on executors, but didn't try to verify what actually happens when there's driver and separate executors. Also would be good to double-check if all registered metrics continue to be exported. Just found https://github.com/cognitedata/spark-jdbc/blob/master/src/main/scala/org/apache/spark/metrics/sink/JdbcSink.scala#L89 that seems to try pruning them a little bit (but only the counter type)

src/main/scala/cognite/spark/v1/MetricsSource.scala
mr-celo: Yeah.. This in not super clear and it was a bit of a pain to stitch together (the jdbc sink adds a bunch of parameters, and it was hard to follow where each came from)
Essentially, it is true that from the point-of-view of the data source, there is no jobId. But the `metricNamespace` already contains the jobId when called. So, from the point of view of the driver, it does have it.

But this is a very good point. There is no reason to expose this to the datasource. I will remove it here, and let the backend do its thing with the knowledge of what the driver does. Will do 👍 

https://github.com/cognitedata/jetfire-backend/blob/93088e6bfc03784137e41aea20a63ff9be89405d/driver/src/main/scala/com/cognite/jetfire/driver/TransformService.scala#L112

src/main/scala/cognite/spark/v1/MetricsSource.scala
mr-celo: I don't understand why it wouldn't remove from the executors based on this.
Is it because the instance that is removing the sources and the counter will be different in the executor? Meaning that none will be found in the map, and therefore none will be remove as a source?

dmivankov: I suspect that even though metrics registry is placed in sparkenv, it might still be isolated per-jvm

PR #933, opened 2024-05-23
Update Chimney to 1.0.0, provide TransformerConfiguration to keep the old behavior…

MateuszKubuszok:
…, fix 2 incorrectly used `u.into[U].withFieldConst(...).transform` where `.withFieldConst` was silently ignored in the presence of an `implicit`

### DIFF
--- a/src/main/scala/cognite/spark/v1/SdkV1Relation.scala
+++ b/src/main/scala/cognite/spark/v1/SdkV1Relation.scala
@@ -61,7 +61,7 @@ abstract class SdkV1Relation[A <: Product, I](config: RelationConfig, shortName:
       updates: Seq[P],
       resource: T,
       isUpdateEmpty: U => Boolean
-  )(implicit transform: Transformer[P, U]): IO[Unit] = {
+  )(implicit transform: Transformer.AutoDerived[P, U]): IO[Unit] = {
     if (!updates.forall(u => u.id.isDefined || u.getExternalId.isDefined)) {
       throw new CdfSparkException("Update requires an id or externalId to be set for each row.")
     }
@@ -81,7 +81,7 @@ abstract class SdkV1Relation[A <: Product, I](config: RelationConfig, shortName:
     }
     val updateExternalIds = if (updatesByExternalId.isEmpty) { IO.unit } else {
       val updatesByExternalIdMap = updatesByExternalId
-        .map(u => u.getExternalId.get -> u.into[U].withFieldComputed(_.externalId, _ => None).transform)
+        .map(u => u.getExternalId.get -> u.transformInto[U])
         .toMap
 
       resource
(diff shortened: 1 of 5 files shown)


src/main/scala/cognite/spark/v1/SdkV1Relation.scala
```diff
@@ -61,7 +61,7 @@ abstract class SdkV1Relation[A <: Product, I](config: RelationConfig, shortName:
       updates: Seq[P],
       resource: T,
       isUpdateEmpty: U => Boolean
-  )(implicit transform: Transformer[P, U]): IO[Unit] = {
+  )(implicit transform: Transformer.AutoDerived[P, U]): IO[Unit] = {
```
MateuszKubuszok, line 64: `Transformer.AutoDerived` is a parent of `Transformer`, that should be used where auto derivation should be triggered in the absence of user-provided implicit.

src/main/scala/cognite/spark/v1/package.scala
```diff
@@ -3,9 +3,13 @@ package cognite.spark
 import com.cognite.sdk.scala.common.{NonNullableSetter, SdkException, SetNull, SetValue, Setter}
 import com.cognite.sdk.scala.v1.{SequenceColumn, SequenceColumnCreate}
 import io.scalaland.chimney.Transformer
+import io.scalaland.chimney.dsl.TransformerConfiguration
 
 // scalastyle:off
 package object v1 {
+  implicit val chimneyConfiguration =
+    TransformerConfiguration.default.enableMethodAccessors.enableDefaultValues
```
MateuszKubuszok, line 11: Both `MethodAccessors` and `DefaultValues` were made opt-in as users considered them to be unsafe defaults.

src/main/scala/cognite/spark/v1/SdkV1Relation.scala
```diff
@@ -81,7 +81,7 @@ abstract class SdkV1Relation[A <: Product, I](config: RelationConfig, shortName:
     }
     val updateExternalIds = if (updatesByExternalId.isEmpty) { IO.unit } else {
       val updatesByExternalIdMap = updatesByExternalId
-        .map(u => u.getExternalId.get -> u.into[U].withFieldComputed(_.externalId, _ => None).transform)
+        .map(u => u.getExternalId.get -> u.transformInto[U])
```
MateuszKubuszok, line 84: `withFieldComputed` was silently ignored - `implicit transform` was used, and code emitted no warning that generating a conversion is impossible for abstract types.

dmivankov:
Thank you!
I took the liberty to apply it in small steps
https://github.com/cognitedata/cdp-spark-datasource/pull/794 - 0.7.5 first
https://github.com/cognitedata/cdp-spark-datasource/pull/937 ignored withFieldComputedFix
more to follow

src/main/scala/cognite/spark/v1/package.scala
dmivankov: @MateuszKubuszok what was the need for `MethodAccessors`, was the default for it changed in chimney moving to v1?
Going to 0.7.5 required only `DefaultValues` to compile https://github.com/cognitedata/cdp-spark-datasource/pull/794/files#diff-0b0d81435732de336a95503c47d48d0fffe6a09474a06d0475d0626a7ce3d14bR11
Going to v1 also doesn't seem to need it https://github.com/cognitedata/cdp-spark-datasource/pull/941/files (copy of this PR updated for latest master branch)

MateuszKubuszok: I added it when I saw the error that there is no accessor with a particular name - however right after that I noticed that virtually all derivations in your codebase rely on default values rather than method accessors. It can be skipped.

dmivankov:
merged via https://github.com/cognitedata/cdp-spark-datasource/pull/942

PR #939, opened 2024-06-07
Update sdk

Jacob-Eliat-Eliat:
This looks suspicious since we are ignoring ReverseDirectRelation from the views, but it makes sense when you consider that ReverseDirectRelation always reference a DirectRelation, so there is still the opportunity to indirectly update reverseDirectRelations by changing the associated DirectRelations.

I don't think it is actually possible to ingest directly into RDR which is why we are simply bypassing it here.


dmivankov (approved):
Discussed a bit outside the PR, still only considering EdgeConnection is ok because that path is only for writes to connections and that isn't supported for ReverseDirectRelationConnection on the API level so we shouldn't allow it either

PR #946, opened 2024-06-28
Direct relations list fix

Jacob-Eliat-Eliat:
Changes:

* Cosmetic: creating a method "getListPropAsSeq" to handle parsing of a row into a list The method is generic and can parse into any type, but I decided to use it to replace the code that was there in place, so I parse into whatever it was first parsed into before. This was done because this piece of code was repeated for each possible list property.

* Cosmetic: dealing with number parsing in a generic way. There were hundreds of lines of code dedicated to parsing numbers and the way they all worked was to parse like this: => any => string => bigint => specific type. I decided to constrain my change to just code deduplication for now, so I created a second order function that takes the final transformation bigint => specific type and does everything.

* Cosmetic: I deleted the `if p.isList` in all cases of the list handling, and the opposite in the non-list handling because we already test this before entering the method.

 * Cosmetic/Functional: I moved the parsing of DirectRelations from happening before the flow separation into lists to happening inside it. So now non-list direct relations are handled in the "toInstancePropertyValueOfNonList" whereas before they were a separate case
 * Functional: I added handling of DirectRelations in `toInstancePropertyValueOfList` For that I reworked `extractDirectRelationReferenceFromStruct` and I added the method `tryAsDirectNodeRelationList`

In the end that's not a lot of functional changes, but it was a bit messy to figure out how it works and how the tests should actually work!


build.sbt
```diff
@@ -35,21 +36,26 @@ lazy val commonSettings = Seq(
   organization := "com.cognite.spark.datasource",
   organizationName := "Cognite",
   organizationHomepage := Some(url("https://cognite.com")),
-  version := "3.18." + patchVersion,
+  version := "3.19." + patchVersion,
```
dmivankov, line 39: is this the only real change in this file? Since we don't enforce the formatting would be easier to review without reformats

Jacob-Eliat-Eliat: ah yes it is... I must admit I got annoyed with permanentely being blocked by fmt from running the pipeline so I enabled autoformat at some point, I can disable and revert.

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -730,185 +750,139 @@ object FlexibleDataModelRelationUtils {
...
-      Left(new CdfSparkException(s"""Error parsing value for field '$propertyName'.
-                                    |Expecting an Int but found '$nAsStr'
-                                    |""".stripMargin))
+      throw new IllegalArgumentException(s"'${String.valueOf(n)}' is not a valid Float")
     }
-  }
 
-  private def tryAsIntSeq(ns: Seq[Any], propertyName: String): Either[CdfSparkException, Seq[Int]] =
+  private def tryConvertNumber[T](
+      n: Any,
```
dmivankov, line 826:
```suggestion
      nAsString: String,
```
`Any.toString` is bit too vague, would be better to move .toString to the caller, where hopefully a more specific type is available (or even better maybe there's a way to avoid String parsing..)

dmivankov:
Haven't read the bulk of dms change part, but so far changes are dms-only, old tests pass and one new is added so looks good so far :slightly_smiling_face: but will try to read through the main part too

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
Jacob-Eliat-Eliat: Yeah there is a bit of "any" abuse going on, but keep in mind this is just refactoring from my part because there was so much duplicated code. The real changes related to list of direct relations are pretty small actually!

Jacob-Eliat-Eliat: I think you could handle it in the row.get(i) doing instead row.getAs[type](i) but it might change how errors are handled, though arguably you could just wrap that in a Try

Jacob-Eliat-Eliat: I was supposing that there is a reason we are doing string => bigDecimal => float instead of directly going to the final number type... but it is possible that there isn't

dmivankov: If that was there before the PR then we can postpone looking deeper :eyes: , only need to check if we changed that part as with Any we get less protection by the compiler..

Jacob-Eliat-Eliat: Not changed. The any come from doing row.get(i)

Jacob-Eliat-Eliat:
> Haven't read the bulk of dms change part, but so far changes are dms-only, old tests pass and one new is added so looks good so far 🙂 but will try to read through the main part too

most of it is just cleanup/code deduplication, the actual changes needed for this were pretty small, but I had some trouble understanding the existing code correctly!

dmivankov:
> the actual changes needed for this were pretty small

could you point to a summary or place in diff? Looking at commits in the PR there is some separation of changes but a bit too many to easily split into cosmetic&refactor vs the fix

Jacob-Eliat-Eliat:
> > the actual changes needed for this were pretty small
> 
> could you point to a summary or place in diff? Looking at commits in the PR there is some separation of changes but a bit too many to easily split into cosmetic&refactor vs the fix

Will do in the PR header. Yeah sorry about the number of commits, lots of them are just scalafmt and stuff like that as well ^^

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -730,185 +750,139 @@ object FlexibleDataModelRelationUtils {
...
+    skipNulls(value).toVector
+      .traverse(extractDirectRelations(propertyName, "Direct Node Relation", instanceSpace, _))
 
-  private def tryAsDoubleSeq(
-      ns: Seq[Any],
-      propertyName: String): Either[CdfSparkException, Seq[Double]] =
+  private def extractDirectRelations(
+      propertyName: String,
+      descriptiveName: String,
+      defaultSpace: Option[String],
```
dmivankov, line 874:
```suggestion
      instanceSpace: Option[String],
```
at least the calling method uses that name

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -668,58 +665,81 @@ object FlexibleDataModelRelationUtils {
...
-          val floatSeq = Try(row.getSeq[Any](i)).getOrElse(row.getAs[Array[Any]](i).toSeq)
-          tryAsFloatSeq(floatSeq, propertyName)
+        case _: DirectNodeRelationProperty =>
+          val structSeq = getListPropAsSeq[Row](propertyName, row, i)
+          tryAsDirectNodeRelationList(structSeq, propertyName, instanceSpace)
+            .map(InstancePropertyValue.ViewDirectNodeRelationList)
+        case _: TextProperty =>
+          Try({
+            val strSeq = getListPropAsSeq[String](propertyName, row, i)
+            InstancePropertyValue.StringList(skipNulls(strSeq))
```
dmivankov, line 698: for TextProperty it was looking a bit differently before - used `[Any]` and `.map(String.valueOf)`. In case of String class it should behave the same as before but not sure if we could have TextProperty with some other type inside :thinking:

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -668,58 +665,81 @@ object FlexibleDataModelRelationUtils {
...
+          }).toEither
+        case _ @PrimitiveProperty(PrimitivePropType.Boolean, _) =>
```
dmivankov, line 700:
```suggestion
        case _ @ PrimitiveProperty(PrimitivePropType.Boolean, _) =>
```
for places like this, scalafmt would also want to insert such spaces probably

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -730,185 +750,139 @@ object FlexibleDataModelRelationUtils {
...
+      .leftMap(exception => {
         val seqAsStr = ns.map(String.valueOf).mkString(",")
-        Left(new CdfSparkException(s"""Error parsing value for field '$propertyName'.
-                                      |Expecting an Array[Float] but found '[$seqAsStr]'
-                                      |${e.getMessage}
-                                      |""".stripMargin))
-    }
+        new CdfSparkException(
+          s"""Error parsing value for field '$propertyName'.
+                                 |Expecting a $expectedType but found '$seqAsStr'
```
dmivankov, line 858:
```suggestion
                                 |Expecting a $expectedType but found '[$seqAsStr]'
```
to keep the old format

dmivankov, line 858:
```suggestion
                                 |Expecting a Array[$expectedType] but found '$seqAsStr'
```
to keep the old format

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -668,58 +665,81 @@ object FlexibleDataModelRelationUtils {
...
+            InstancePropertyValue.StringList(skipNulls(strSeq))
+          }).toEither
+        case _ @PrimitiveProperty(PrimitivePropType.Boolean, _) =>
+          Try({
+            val boolSeq = getListPropAsSeq[Boolean](propertyName, row, i)
+            InstancePropertyValue.BooleanList(boolSeq)
+          }).toEither
+        case _ @PrimitiveProperty(PrimitivePropType.Float32, _) =>
+          val floatSeq = getListPropAsSeq[Any](propertyName, row, i)
+          tryConvertNumberSeq(floatSeq, propertyName, Float.toString, safeConvertToFloat)
```
dmivankov, line 707:
```suggestion
          tryConvertNumberSeq(floatSeq, propertyName, "Float", safeConvertToFloat)
```
Multiple possibilities for error here otherwise
- `Float.toString` does look like a lambda `(x: Float) => x.toString` but it isn't the inention here afaics
- intellij says it'd rather be 
```scala
  /** The String representation of the scala.Float companion object. */
  override def toString = "object scala.Float"
```
which isn't the user-facing type to show (still readable but with more noise than before)

dmivankov (commented):
some small notes for the non-fix part

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
Jacob-Eliat-Eliat: I think it's ok because it's wrapped in the try. Maybe it would make more sense to modify the other ones to use a similar construct too?

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
dmivankov: Interesting, I thought it wanted to format with spaces, or maybe it depends on whether the capture is named or not :thinking:

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
dmivankov: In case there can be other type inside, those cases would now generate errors. Most likely it is always string but keeping Any to be on a safe side is an option

Jacob-Eliat-Eliat: Ah though not actually, the behavior will differ, since String.valueOf cannot fail

dmivankov (approved):

PR #956, opened 2024-07-22
node type

Jacob-Eliat-Eliat:
Context: Right now the "type" property is not seen when writing SQL queries for spark in jetfire. I think this is the reason.

In addition to the tests in cdp-spark-datasource, it makes this branch's new test pass: https://github.com/cognitedata/jetfire-backend/pull/2388/files


src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
```diff
@@ -149,8 +149,14 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
           .traverse(
             toInstanceFilter(instanceType, _, space = space, versionedExternalId = versionedExternalId))
           .map(FilterDefinition.Or.apply)
+      //Node types are optional so exists will be automatically checked
+      //Types are a property of the node/edge and should not use the view but directly check on the node
+      case IsNotNull(attribute) if attribute.equalsIgnoreCase("type") =>
+        Right(FilterDefinition.Exists(Seq("node", "type")))
```
dmivankov, line 155: is IsNotNull applicable to edge types too?

Jacob-Eliat-Eliat: Edge types are not nullable so it shouldn't be I think. This is called automatically when trying to filter by a nullable field, but to be honest I haven't checked if someone could intentionally call is not null on edge types. I'll do that

dmivankov: maybe we could check for instanceType=edge and fail as a start

Jacob-Eliat-Eliat: yeah makes sense, at a minimum I should check it's not called with an edge in the case

Jacob-Eliat-Eliat: Modified the test to try to call is not null by adding  `and type is not null` in the filter and it still works, I also added the check to not enter this case if it's an edge.

Jacob-Eliat-Eliat: note: there is some code duplication with toInstanceFilter and toNodeOrEdgeAttributeFilter. Would be sensible to remove it but I though it's better to make this work first becore doing some refactoring, the flow is quite complicated and both filters are used

dmivankov (approved):

PR #957, opened 2024-07-25
Split workflow publish

Jacob-Eliat-Eliat:
context: lately the pipeline for publish has been flaky (due to issue with sonatype) this PR will make it possible to rerun it without running the tests again (provided that the tests passed).

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -64,6 +64,7 @@ jobs:
           fail_ci_if_error: false
   publish:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     needs: test
     if: github.ref == 'refs/heads/master'
     environment: CD
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -77,6 +77,9 @@ jobs:
           if [ ${{github.ref }} == "refs/heads/master" ]; then
             LIB_PACKAGE="+fatJarShaded/assembly"
           fi
+      - name: Check dependencies
+        run: |
+          sbt -Dsbt.log.noformat=true +{,macroSub/,structType/}dependencyLockCheck
       - name: Install GPG & Sonatype credentials for SBT
         env:
           GPG_PUBLIC_KEY: ${{ secrets.SONATYPE_OSSRH_GPG_PUBLIC_KEY }}
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -58,7 +58,7 @@ jobs:
   publish:
     runs-on: ubuntu-latest
     timeout-minutes: 30
-    needs: test
+    needs: build-and-push
     if: github.ref == 'refs/heads/master'
     environment: CD
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yaml
```diff
@@ -3,7 +3,7 @@ name: CI
 on: ["push"]
 
 jobs:
-  build-and-push:
+  test:
```
dmivankov, line 6: https://github.com/cognitedata/terraform/blob/master/github/apps/main.tf#L1433 will need to be updated, once manually in repo settings is ok but to avoid terraform applies to revert it need to persist in terraform too

.github/workflows/main.yaml
```diff
@@ -62,14 +62,28 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: false
+  publish:
+    runs-on: ubuntu-latest
```
dmivankov, line 66:
```suggestion
    runs-on: ubuntu-latest
    timeout-minutes: 30
```
let's put some limit here too

.github/workflows/main.yaml
Jacob-Eliat-Eliat: ah that sucks, we could rename the pipeline back, but I feel like it makes more sense to change it in terraform too right?

.github/workflows/main.yaml
```diff
@@ -3,7 +3,7 @@ name: CI
 on: ["push"]
 
 jobs:
-  build-and-push:
+  test:
     runs-on: ubuntu-latest
     timeout-minutes: 80
     environment: ${{ github.ref == 'refs/heads/master' && 'CD' || 'CI' }}
```
dmivankov, line 9: `      - name: Set variables` section should be deleted for a cleaner definition, normally it looks like it sets stuff used in published artifacts, but in this particular repo all those BUILD_DATE,IMAGE_REVISION,DOCKER_IMAGE_ seem to be unused

.github/workflows/main.yaml
```diff
@@ -62,14 +62,28 @@ jobs:
...
+          fetch-depth: 0
+      - run: |
+          git rev-list --count --first-parent HEAD > patch_version.txt
+      - name: Set up JDK 8
+        uses: coursier/setup-action@v1
+        with:
+          jvm: adopt:8
       - name: Build JAR file
         run: |
           LIB_PACKAGE="+library/package"
```
dmivankov, line 76: I'd suggest keeping at least `sbt -Dsbt.log.noformat=true +{,macroSub/,structType/}dependencyLockCheck` somewhere in publish flow too. If build-test and publish are separated it means publish (retry) can happen much later and so external artifacts contents might have changed (even better would be if sbt enforced them, not just validated on demand..)

.github/workflows/main.yaml
dmivankov: yeah, it no longer pushes, but for a quick iteration we can keep old name and rename separately

.github/workflows/main.yaml
Jacob-Eliat-Eliat: Makes sense

dmivankov (approved):

.github/workflows/main.yaml
Jacob-Eliat-Eliat: let's do that

PR #959, opened 2024-07-26
CDF-22158: dual-publish to jfrog & sonatype

dmivankov:
[CDF-22158]

[CDF-22158]: https://cognitedata.atlassian.net/browse/CDF-22158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -55,7 +55,55 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: false
-  publish:
+  publish_jfrog:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    needs: build-and-push
+    if: github.ref == 'refs/heads/master'
+    environment: CD
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - run: |
+          git rev-list --count --first-parent HEAD > patch_version.txt
+      - name: Set up JDK 8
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yaml
```diff
@@ -55,7 +55,55 @@ jobs:
...
+          LIB_PACKAGE="+library/package"
+          if [ ${{github.ref }} == "refs/heads/master" ]; then
+            LIB_PACKAGE="+fatJarShaded/assembly"
+          fi
+      - name: Check dependencies
+        run: |
+          sbt -Dsbt.log.noformat=true +{,macroSub/,structType/}dependencyLockCheck
+      - name: Install GPG & Sonatype credentials for SBT
+        env:
+          GPG_PUBLIC_KEY: ${{ secrets.SONATYPE_OSSRH_GPG_PUBLIC_KEY }}
```
Jacob-Eliat-Eliat, line 85: secrets for sonatype shouldn't be in the publish jfrog one, we use different secrets right?

dmivankov: those are gpg signing keys, if we want to keep signing for jfrog too we need them, makes sense to sign both, also should be fine to sign the same way

Jacob-Eliat-Eliat: Ok makes sense

Jacob-Eliat-Eliat (approved):

PR #963, opened 2024-08-01
node types

Jacob-Eliat-Eliat:


dmivankov:
drafting for now as we except changes to how to handle node types

src/test/scala/cognite/spark/v1/FlexibleDataModelCorePropertyRelationTest.scala
```diff
@@ -493,20 +825,36 @@ class FlexibleDataModelCorePropertyRelationTest
       cursor = ""
     )
 
-    val readAllDf = readRows(
```
Jacob-Eliat-Eliat, line 496: I don't quite understand how this was working before. Reading the code it looks like we are clearly requesting instances of viewAll but of type edge and node, and then combining them. This should return only instances of viewAll which there are 2, not 8, yet it's expected later to contain 8 elements?

src/test/scala/cognite/spark/v1/FlexibleDataModelCorePropertyRelationTest.scala
```diff
@@ -493,20 +825,36 @@ class FlexibleDataModelCorePropertyRelationTest
...
+    val edgesMissingCols = nodesColumns -- edgesColumns
+    val nodesMissingCols = edgesColumns -- nodesColumns
+
+    val nodesWithAllCols = nodesMissingCols.foldLeft(readNodesDf) {
+      (df, col) => df.withColumn(col, lit(null))
+    }
+    val edgesWithAllCols = edgesMissingCols.foldLeft(readEdgesDf) {
+      (df, col) => df.withColumn(col, lit(null))
+    }
+    val readAllDf = nodesWithAllCols.unionAll(edgesWithAllCols)
```
Jacob-Eliat-Eliat, line 857: obviously this is not ideal... My understanding is that we used to return null columns in type, startNode and endNode when doing this, but obviously this is not possible if you want to also support node "type" property, because then you are unioning the edge.type and the node's view type as the same column.

This test passes if you artificially increase the number of column to match, as done here. Remains to be seen if it won't break something else in the jetfire backend or the sql queries customers might do. I wonder if-or how-joining edges and nodes instances of the same view would make or be possible in sql

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
```diff
@@ -149,13 +154,19 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
       //Node types are optional so exists will be automatically checked
       //Types are a property of the node/edge and should not use the view but directly check on the node
       case IsNotNull(attribute)
-          if attribute.equalsIgnoreCase("type") && instanceType == InstanceType.Node =>
+          if attribute.equalsIgnoreCase("_type") && instanceType == InstanceType.Node =>
```
dmivankov, line 157: no need for Edge case with IsNotNull(_type)?

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
```diff
@@ -230,7 +254,7 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
             case s if s.equalsIgnoreCase("spaceExternalId") => n.space
             case s if s.equalsIgnoreCase("externalId") => n.externalId
             case s if s.equalsIgnoreCase("metadata.cursor") => cursor.getOrElse("")
-            case s if s.equalsIgnoreCase("type") =>
+            case s if s.equalsIgnoreCase("_type") =>
```
dmivankov, line 257: should "type" also be kept?

dmivankov: ah, maybe not for NodeDefinition case

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -440,11 +438,20 @@ object FlexibleDataModelRelationUtils {
                                       |""".stripMargin))
     }
 
-  private def extractNodeOrEdgeTypeDirectRelation(
+  private def extractNodeTypeDirectRelation(
       schema: StructType,
       instanceSpace: Option[String],
       row: Row): Either[CdfSparkException, DirectRelationReference] =
-    extractDirectRelation("type", "Node or Edge type", schema, instanceSpace, row)
+    extractDirectRelation("_type", "Node or Edge type", schema, instanceSpace, row)
```
dmivankov, line 445:
```suggestion
    extractDirectRelation("_type", "Node type", schema, instanceSpace, row)
```

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -440,11 +438,20 @@ object FlexibleDataModelRelationUtils {
...
+
+  private def extractEdgeTypeDirectRelation(
+      schema: StructType,
+      instanceSpace: Option[String],
+      row: Row): Either[CdfSparkException, DirectRelationReference] =
+    extractDirectRelation("_type", "Node or Edge type", schema, instanceSpace, row) match {
```
dmivankov, line 451:
```suggestion
    extractDirectRelation("_type", "Edge type", schema, instanceSpace, row) match {
```

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -440,11 +438,20 @@ object FlexibleDataModelRelationUtils {
...
+      case right @ Right(_) => right
+      case _ => extractDirectRelation("type", "Node or Edge type", schema, instanceSpace, row)
```
dmivankov, line 453:
```suggestion
      case _ => extractDirectRelation("type", "Edge type", schema, instanceSpace, row)
```

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
dmivankov: also not sure if it should spell `_type` in 2nd argument :thinking:

dmivankov (approved):
Not sure if isNull/isNotNull may happen for required _type case, might be worth handling but if it pops up can be fixed later too I guess

Jacob-Eliat-Eliat:
> Not sure if isNull/isNotNull may happen for required _type case, might be worth handling but if it pops up can be fixed later too I guess

I'm less concerned about that for the reasons you said and more about changes to handling of instanceType "All", I'll try publishing the snapshot and making jetfire work with this

PR #965, opened 2024-08-15
Improve filters

Jacob-Eliat-Eliat:
So there were always two cases:
 * filter on reserved keywords
 * filter on normal view attributes
 
 They were handled somewhat separately, but the entire code for filtering on reserved keywords was repeated. Similarly, the code for combined filter (not, and, or etc...) was repeated too.

This makes it so there is a single filter function with optional view version and space, which will check if the attribute is a reserved keyword and if so will use the filter on reserved keyword, if not will filter on view attributes.

This enables some cleanup in other parts as well, where these two were chained.

Some more cleanup was done to avoid having to individually check each attributes, and instead handle them together (while keeping in mind some are string attributes and others are references)

It's not perfect because you still need to pattern match the filter just to see if it's a reserved attribute or not...

### DIFF
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
@@ -57,13 +57,12 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
     })
   }
 
-
   private def isReservedAttribute(instanceType: InstanceType, attribute: String) = {
     val alwaysReservedAttributes: Set[String] = Set("space", "externalId", "_type")
     val edgeReservedAttributes: Set[String] = Set("type", "startNode", "endNode")
     // type is supported as an alias to _type for edges for legacy reasons.
     alwaysReservedAttributes.contains(attribute) ||
-      (instanceType == InstanceType.Edge && edgeReservedAttributes.contains(attribute))
+    (instanceType == InstanceType.Edge && edgeReservedAttributes.contains(attribute))
   }
 
   private def extractInstancePropertyValue(key: String, value: InstancePropertyValue): Any =
@@ -74,7 +73,7 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
       sparkFilter: Filter,
       space: Option[String],
       versionedExternalId: Option[String]
-  ): Either[CdfSparkException, FilterDefinition] = {
+  ): Either[CdfSparkException, FilterDefinition] =
     sparkFilter match {
       case EqualTo(attribute, _) if isReservedAttribute(instanceType, attribute) =>
         toNodeOrEdgeAttributeFilter(instanceType, sparkFilter)
@@ -88,13 +87,11 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
         toNodeOrEdgeAttributeFilter(instanceType, sparkFilter)
       case And(f1, f2) =>
         Vector(f1, f2)
-          .traverse(
-            toFilter(instanceType, _, space = space, versionedExternalId))
+          .traverse(toFilter(instanceType, _, space = space, versionedExternalId))
           .map(FilterDefinition.And.apply)
       case Or(f1, f2) =>
         Vector(f1, f2)
-          .traverse(
-            toFilter(instanceType, _, space = space, versionedExternalId))
+          .traverse(toFilter(instanceType, _, space = space, versionedExternalId))
           .map(FilterDefinition.Or.apply)
       case Not(f) =>
         toFilter(instanceType, f, space = space, versionedExternalId)
@@ -109,17 +106,16 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
                | and versionedExternalId: ${versionedExternalId.getOrElse("not specified")}
                | and instanceType: $instanceType """.stripMargin))
     }
-  }
 
   private def toInstanceFilter(
       sparkFilter: Filter,
       space: String,
-      versionedExternalId: String): Either[CdfSparkException, FilterDefinition] = {
+      versionedExternalId: String): Either[CdfSparkException, FilterDefinition] =
     sparkFilter match {
       case EqualTo(attribute, value) =>
         toFilterValueDefinition(attribute, value).map(
           FilterDefinition.Equals(Seq(space, versionedExternalId, attribute), _))
-    case In(attribute, values) =>
+      case In(attribute, values) =>
         toSeqFilterValueDefinition(attribute, values).map(
           FilterDefinition.In(Seq(space, versionedExternalId, attribute), _))
       case GreaterThanOrEqual(attribute, value) =>
@@ -147,7 +143,6 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
           new CdfSparkIllegalArgumentException(
             s"Unsupported filter '${f.getClass.getSimpleName}', ${f.toString}"))
     }
-  }
 
   // Some reserved attributes are not attached to a view but directly to the node/edge
   // This handles filtering on these attributes.
@@ -162,7 +157,8 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
           FilterDefinition.Equals(
             createNodeOrEdgeCommonAttributeRef(instanceType, attribute),
             FilterValueDefinition.String(String.valueOf(value))))
-      case EqualTo(attribute, value: GenericRowWithSchema) if nodeOrEdgeReferenceAttributes.contains(attribute) =>
+      case EqualTo(attribute, value: GenericRowWithSchema)
+          if nodeOrEdgeReferenceAttributes.contains(attribute) =>
         createEqualsAttributeFilter(createNodeOrEdgeCommonAttributeRef(instanceType, "space"), value)
       case In(attribute, values) if nodeOrEdgeStringAttributes.contains(attribute) =>
         toSeqFilterValueDefinition(attribute, values)
@@ -183,10 +179,14 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
       case IsNotNull(attribute) =>
         Right(FilterDefinition.Exists(createNodeOrEdgeCommonAttributeRef(instanceType, attribute)))
       case IsNull(attribute) =>
-        Right(FilterDefinition.Not(FilterDefinition.Exists(createNodeOrEdgeCommonAttributeRef(instanceType, attribute))))
+        Right(
+          FilterDefinition.Not(
+            FilterDefinition.Exists(createNodeOrEdgeCommonAttributeRef(instanceType, attribute))))
       case f =>
-        Left(new CdfSparkIllegalArgumentException(
-          s"""Unsupported node or edge attribute filter '${f.getClass.getSimpleName}': ${String.valueOf(f)}
+        Left(
+          new CdfSparkIllegalArgumentException(
+            s"""Unsupported node or edge attribute filter '${f.getClass.getSimpleName}': ${String
+                 .valueOf(f)}
              | for instanceType: $instanceType
              |""".stripMargin))
     }
@@ -342,7 +342,7 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
   // Filter definition for node/edge `type`, `startNode` & `endNode`
   private def createEqualsAttributeFilter(
       attributeVector: Seq[String],
-      struct: GenericRowWithSchema): Either[CdfSparkException, FilterDefinition] = {
+      struct: GenericRowWithSchema): Either[CdfSparkException, FilterDefinition] =
     Try {
       val space = struct.getString(struct.fieldIndex("space"))
       val externalId = struct.getString(struct.fieldIndex("externalId"))
@@ -358,14 +358,13 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
            |""".stripMargin
       )
     }
-  }
 
   // Filter definitions for attributes for nodes & edges
   private def createNodeOrEdgeCommonAttributeRef(
       instanceType: InstanceType,
       attribute: String): Seq[String] = {
     //type is a special case, and is reserved for edges only and is an alias of _type
-    if(attribute.equalsIgnoreCase("type")) {
+    if (attribute.equalsIgnoreCase("type")) {
       Vector(instanceType.productPrefix.toLowerCase(Locale.US), "_type")
     }
     Vector(instanceType.productPrefix.toLowerCase(Locale.US), attribute)
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
@@ -364,8 +364,8 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
       instanceType: InstanceType,
       attribute: String): Seq[String] =
     //type is a special case, and is reserved for edges only and is an alias of _type
-    if (attribute.equalsIgnoreCase("type")) {
-      Vector(instanceType.productPrefix.toLowerCase(Locale.US), "_type")
+    if (attribute.equalsIgnoreCase("_type")) {
+      Vector(instanceType.productPrefix.toLowerCase(Locale.US), "type")
     } else {
       Vector(instanceType.productPrefix.toLowerCase(Locale.US), attribute)
     }
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
@@ -366,8 +366,9 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
     //type is a special case, and is reserved for edges only and is an alias of _type
     if (attribute.equalsIgnoreCase("type")) {
       Vector(instanceType.productPrefix.toLowerCase(Locale.US), "_type")
+    } else {
+      Vector(instanceType.productPrefix.toLowerCase(Locale.US), attribute)
     }
-    Vector(instanceType.productPrefix.toLowerCase(Locale.US), attribute)
   }
 
   private def relationReferenceInnerStruct(): StructType =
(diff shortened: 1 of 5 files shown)


src/test/scala/cognite/spark/v1/FlexibleDataModelCorePropertyRelationTest.scala
```diff
@@ -51,13 +51,13 @@ class FlexibleDataModelCorePropertyRelationTest
...
+  private val viewAllAmbiguousTypeExternalId = "sparkDsTestViewAllAmbiguousType6"
+  private val viewNodesAmbiguousTypeExternalId = "sparkDsTestViewNodesAmbiguousType6"
+  private val viewEdgesAmbiguousTypeExternalId = "sparkDsTestViewEdgesAmbiguousType6"
 
-  private val viewAllTypeExternalId = "sparkDsTestViewAllType5"
-  private val viewNodesTypeExternalId = "sparkDsTestViewNodesType5"
-  private val viewEdgesTypeExternalId = "sparkDsTestViewEdgesType5"
+  private val viewAllTypeExternalId = "sparkDsTestViewAllType6"
+  private val viewNodesTypeExternalId = "sparkDsTestViewNodesType6"
+  private val viewEdgesTypeExternalId = "sparkDsTestViewEdgesType6"
```
Jacob-Eliat-Eliat, line 61: Might be just my new tests that are badly setup bc they are the main culprit for flakyness?

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
```diff
@@ -127,113 +131,67 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
...
+    val nodeOrEdgeStringAttributes = Seq("space", "externalId")
+    val nodeOrEdgeReferenceAttributes = Seq("type", "_type", "startNode", "endNode")
     sparkFilter match {
-      case EqualTo(attribute, value) if attribute.equalsIgnoreCase("space") =>
+      case EqualTo(attribute, value) if nodeOrEdgeStringAttributes.contains(attribute) =>
         Right(
           FilterDefinition.Equals(
-            createNodeOrEdgeCommonAttributeRef(instanceType, "space"),
-            FilterValueDefinition.String(String.valueOf(value))))
-      case EqualTo(attribute, value) if attribute.equalsIgnoreCase("externalId") =>
```
dmivankov, line 191: there was ignoreCase part, iirc scala has CIString
not sure if easy to check for real usage with wrong casing and then drop ignorecase

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
```diff
@@ -382,37 +340,35 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
...
     }
-  }
 
-  // Filter definitions for "space" & "externalId" attributes for nodes & edges
+  // Filter definitions for attributes for nodes & edges
   private def createNodeOrEdgeCommonAttributeRef(
       instanceType: InstanceType,
       attribute: String): Seq[String] =
-    Vector(instanceType.productPrefix.toLowerCase(Locale.US), attribute)
+    if (attribute.equalsIgnoreCase("_type")) {
```
dmivankov, line 367: nit: might also be an option to wrap only attribute value replacement in if-else and keep vector&instanceType in place

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
```diff
@@ -127,113 +131,67 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
...
-        Right(FilterDefinition.Exists(Seq("edge", "type")))
-      case IsNull(attribute)
-          if attribute.equalsIgnoreCase("_type") && instanceType == InstanceType.Node =>
-        Right(FilterDefinition.Not(FilterDefinition.Exists(Seq("node", "type"))))
-      case IsNull(attribute)
-          if attribute.equalsIgnoreCase("type") && instanceType == InstanceType.Edge =>
-        Right(FilterDefinition.Not(FilterDefinition.Exists(Seq("edge", "type"))))
+          if nodeOrEdgeReferenceAttributes.contains(attribute) =>
+        createEqualsAttributeFilter(createNodeOrEdgeCommonAttributeRef(instanceType, attribute), value)
+      case In(attribute, values) if nodeOrEdgeStringAttributes.contains(attribute) =>
```
dmivankov, line 164: Looks like In&StringStartsWith handling wasn't here before, did it move in because of refactoring or it's adding/fixing support for those cases?

dmivankov:
From a quick look only have couple notes, will try to do a full compare before<>after a bit later to see if there could be any behavior changes

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
Jacob-Eliat-Eliat: It was in the other filter not this one, so yeah this is adding/fixing. Presumably in case of sync, only these filters were called so this adds support for these pushdown filters in the sync endpoints, I think in other places the two filters were chained.

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
Jacob-Eliat-Eliat: Ah yeah might be worth ignoring case everywhere

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
dmivankov: For sync actually there was effort to remove most of pushdown as it may be not backend by indexes :eyes:

Jacob-Eliat-Eliat: Though... I think sync filters had filtering on these "special" attributes already (externalId, space, type, start & endNode) so there is an index on them, and that shouldn't be different if you do indexedColumn = something or indexedColumn in/startswith something... unless the implementation on the side of fdm is weird.

Jacob-Eliat-Eliat: Should I ask them then?

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
Jacob-Eliat-Eliat: though as you say, it would be weird for a customer to write externalId or other keywords the "wrong" way...  since attributes aren't case insensitive I think this is overeager use of ignore case.

Jacob-Eliat-Eliat: It seems like nothing supports case insensitiveness (not querying directly, nor using gaphql) so we are alone in doing this, which raises the question of whether we should.

Jacob-Eliat-Eliat: I mean it's more of a hassle than just supporting it, but we aren't testing it or anything, which feels like an undocumented feature and could cause problems down the line

dmivankov: Yeah, most likely it's wrong to have it but safer to keep and remove in a separate PR in case of reverts or failing tests

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
dmivankov: Makes sense, if they were pushed down then equality filters should be fine, prefix a bit less certain. Won't hurt to double-check still :+1:

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
Jacob-Eliat-Eliat: Actually it doesn't matter, we have in jetfire set the property:     .config("spark.sql.caseSensitive", "true")
So even though the pushdown filters are case-insensitive, the second pass done in SQL will not be, so it won't work if you use the wrong case right now.

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
```diff
@@ -57,61 +57,65 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
...
+
   private def extractInstancePropertyValue(key: String, value: InstancePropertyValue): Any =
     FlexibleDataModelRelationUtils.extractInstancePropertyValue(schema.apply(key).dataType, value)
 
-  protected def toInstanceFilter(
+  protected def toFilter(
       instanceType: InstanceType,
+      sparkFilter: Filter,
+      space: Option[String],
+      versionedExternalId: Option[String]
```
dmivankov, line 76:
```suggestion
      viewReference: Option[ViewReference],
```
are those always view attributes and only set/unset together? Would be good to clarify, even better maybe find a more specific name than `viewReference` (or is it clear enough which view it is here?)

Jacob-Eliat-Eliat: good point, nicer like this. I think viewReference is fine here, that's what this is

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -61,7 +61,7 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
       HasData(List(viewRef))
     }
     val requestFilters: Seq[FilterDefinition] = (filters.map {
-      toNodeOrEdgeAttributeFilter(instanceType, _).toOption
+      toFilter(instanceType, _, None).toOption
```
dmivankov, line 64: interesting that here we may have viewReference!=None, not sure if it'd be applicable to use it though

Jacob-Eliat-Eliat:
> From a quick look only have couple notes, will try to do a full compare before<>after a bit later to see if there could be any behavior changes

Worth noting that for sync we would call `toNodeOrEdgeAttributeFilter` directly, which means you need to compare that with the flow that would happen now when you call `toFilter` if you send None as the viewReference

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
Jacob-Eliat-Eliat: I think that's what you were talking about, where we didn't want to pushdown filters on non-indexed columns

dmivankov: let's add a comment that we don't pass viewRef to only do pushdown on `filters`? Though probably `filters` would only include what we want and we're safe either way? :thinking:

Jacob-Eliat-Eliat: Why would filter only include what we want?

dmivankov: Ugh, tricky place indeed, before it was using `toNodeOrEdgeAttributeFilter` that only does pushdown on built-in attributes but incoming filters can be anything

dmivankov: Though old code should fail with "Unsupported node or edge attribute filter" if filters contain anything else :thinking:

Jacob-Eliat-Eliat: That's true hmm, I wonder how the incoming filters are restricted then hmm...

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
Jacob-Eliat-Eliat: no reply from fdm, what do I do there ^^?

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertySyncRelation.scala
Jacob-Eliat-Eliat: Should I leave a comment then or not? I guess somehow the sync calls is generated using only the correct filters, maybe its something in jetfire itself

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
```diff
@@ -57,183 +58,162 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
...
+      viewReference: Option[ViewReference]
+  ): Either[CdfSparkException, FilterDefinition] =
+    sparkFilter match {
+      case EqualTo(attribute, _) if isReservedAttribute(instanceType, attribute) =>
+        toNodeOrEdgeAttributeFilter(instanceType, sparkFilter)
+      case IsNull(attribute) if isReservedAttribute(instanceType, attribute) =>
+        toNodeOrEdgeAttributeFilter(instanceType, sparkFilter)
+      case IsNotNull(attribute) if isReservedAttribute(instanceType, attribute) =>
+        toNodeOrEdgeAttributeFilter(instanceType, sparkFilter)
+      case StringStartsWith(attribute, _) if isReservedAttribute(instanceType, attribute) =>
```
dmivankov, line 94: so for /sync case we will now allow pushdown of prefix filters which isn't necessarily backed by indexes, so either we take risk here or make sure we don't pushdown it on sync (on non-sync pushdown was present, so might be a sign index exists, or might be a bug or conscious decision, no pre-existing comments on it...)

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
```diff
@@ -57,183 +58,162 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
...
+  protected def toFilter(
       instanceType: InstanceType,
       sparkFilter: Filter,
-      space: String,
-      versionedExternalId: String): Either[CdfSparkException, FilterDefinition] = {
+      // viewReference is required for non-reserved attribute filters resolution
+      viewReference: Option[ViewReference]
+  ): Either[CdfSparkException, FilterDefinition] =
+    sparkFilter match {
+      case EqualTo(attribute, _) if isReservedAttribute(instanceType, attribute) =>
```
dmivankov, line 88: there's also `EqualNullSafe` and some others potentially applicable but let's stick to existing list of filters for now :+1:

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
Jacob-Eliat-Eliat: The most logical thing would be that the index is used, but yes it is a bit of a risk. I would go for it personally.

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
```diff
@@ -57,183 +58,164 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
...
-      case IsNull(attribute)
-          if attribute.equalsIgnoreCase("type") && instanceType == InstanceType.Edge =>
-        Right(FilterDefinition.Not(FilterDefinition.Exists(Seq("edge", "type"))))
+          if nodeOrEdgeReferenceAttributes.contains(attribute) =>
+        createEqualsAttributeFilter(createNodeOrEdgeCommonAttributeRef(instanceType, attribute), value)
+      case In(attribute, values) if nodeOrEdgeStringAttributes.contains(attribute) =>
+        toSeqFilterValueDefinition(attribute, values)
+          .filterOrElse(
+            {
+              case FilterValueDefinition.StringList(_) => true
```
dmivankov, line 192: we start accepting `In` for any reserved property now, should we drop `StringList` requirement?

dmivankov (approved):
Quite many small behavior changes coming along with this, none seem particularly dangerous though
- removal of case insensitive property name handling should be ok
- accepting a bit more pushdown cases

But overall it looks matching existing intent, specifically /sync case behavior looks ~same: almost no pushdown (and maybe still rejecting `where` clauses on non-reserved properties?), and ~same pushdown for other cases

src/main/scala/cognite/spark/v1/FlexibleDataModelBaseRelation.scala
Jacob-Eliat-Eliat: This only accepts in for string attributes (because of the if nodeOrEdgeStringAttributes.contains(attribute) guard) but I could add support for the other ones as well, think I should?

I probably should look at how this sync request is created because I would assume it's standardized somehow otherwize we would encounter "Unsupported node or edge attribute filters" everywhere... though we absorb the exceptions in some places via transforming to optional

Jacob-Eliat-Eliat: I guess that's what's happening! When we call the filters in sync we do a .toOption which basically cancel out the "Unsupported node or edge attribute filter" exception.

Jacob-Eliat-Eliat: So probably there is tons of filters that are on attributes of views, but they all result in exceptions, then .toOption removes the exceptions into a None then the ++ adds them together and leaves only the filters on instances attributes

dmivankov: though it looks broken then, it'd probably give up on having any supported filters if there's unsupported filter somewhere, instead it should probably do what other resources do via `pushdownToFilters` by collecting things looking at logical tree

Jacob-Eliat-Eliat: Hmm, not the case everywhere, I'm pretty sure it would indeed throw an error in some, but the one in sync should work: `(filters.map {
      toNodeOrEdgeAttributeFilter(instanceType, _).toOption
    } ++ Array(hasData)).flatten.toSeq` => doesn't that properly discards failing filters?

Jacob-Eliat-Eliat:
To avoid changing anything in term of what we actually push down I changed the pushdown filters to have a flag to see if we are in a sync request, and if so, not to push down filter on "startsWith" and "in"

PR #967, opened 2024-08-19
remove duplication in createNodeOrEdge

Jacob-Eliat-Eliat:
This PR proposes removing a duplicate flow for empty sources.

Currently we handle the situations where sources are empty (writing a node/edge as a standalone, not referencing any view) and the situations where we are writing the instance to a view completely separately, with method overload.

This proposes removing the separated flow, and handling it together, with the sources being optional.

Advantages:
 * removal of redundant code
 * easier to maintain (only have to maintain one flow)
 * relatively straightforward handling of None when sources are empty

The price is that it's very slightly more complex and if we want to make more code specific to one situation in the future we might have to duplicate some parts again.

Reasoning for doing this now:
 * Edge property support is coming soon, and I'd rather not have to deal with two separate functions depending if the source exists or not.

### DIFF
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertyRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertyRelation.scala
@@ -55,8 +55,8 @@ private[spark] class FlexibleDataModelCorePropertyRelation(
     val firstRow = rows.headOption
     (firstRow, viewReference) match {
       case (Some(fr), _) =>
-        upsertNodesOrEdges(rows, fr.schema, viewReference, allProperties, instanceSpace).flatMap(results =>
-          incMetrics(itemsUpserted, results.length))
+        upsertNodesOrEdges(rows, fr.schema, viewReference, allProperties, instanceSpace)
+          .flatMap(results => incMetrics(itemsUpserted, results.length))
       case (None, _) => incMetrics(itemsUpserted, 0)
     }
   }
--- a/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertyRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertyRelation.scala
@@ -53,7 +53,7 @@ private[spark] class FlexibleDataModelCorePropertyRelation(
 
   override def upsert(rows: Seq[Row]): IO[Unit] = {
     val firstRow = rows.headOption
-    (firstRow, viewReference) match {
+    firstRow match {
       case (Some(fr), _) =>
         upsertNodesOrEdges(rows, fr.schema, viewReference, allProperties, instanceSpace)
           .flatMap(results => incMetrics(itemsUpserted, results.length))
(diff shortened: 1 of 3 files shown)


src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -93,7 +84,9 @@ object FlexibleDataModelRelationUtils {
           externalId = externalId,
           instanceSpace = space,
           source,
-          typeDirectRelation =
+          nodeTypeDirectRelation =
+            extractNodeTypeDirectRelation(schema, instanceSpace.orElse(Some(space)), row).toOption,
+          edgeTypeDirectRelation =
```
Jacob-Eliat-Eliat, line 89: This is the only functional change. By doing this, I make sure that if writing a node we are only getting from "_type" and not from "type" even on a nodeOrEdge view.

Before if you had a nodeOrEdge view and were writing to a node instance implementing this view, and the view had a "type" attribute you would encounter an issue, this should fix that problem.

The reason this fix was not higher priority is because I don't think it actually happens for anyone -yet- NodeOrEdges are rare in fdm and attributes of views named "type" are too.

src/main/scala/cognite/spark/v1/FlexibleDataModelCorePropertyRelation.scala
```diff
@@ -240,31 +234,6 @@ private[spark] class FlexibleDataModelCorePropertyRelation(
...
   }
 
-  private def upsertNodesOrEdges(
-      instanceSpace: Option[String],
-      rows: Seq[Row],
-      schema: StructType): IO[Seq[SlimNodeOrEdge]] = {
-    val nodesOrEdges = intendedUsage match {
-      case Usage.Node => createNodes(rows, schema, instanceSpace)
-      case Usage.Edge => createEdges(rows, schema, instanceSpace)
-      case Usage.All => createNodesOrEdges(rows, schema, instanceSpace)
```
dmivankov, line 250: note: looks like now they will pass along respect config.ignoreNullFields, which sounds like a correct fix :+1:

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
```diff
@@ -605,14 +496,25 @@ object FlexibleDataModelRelationUtils {
         }
     }
 
+  private def validateSourceSchema(
+      source: Option[SourceReference],
+      schema: StructType,
+      propertyDefMap: Map[String, ViewPropertyDefinition]): Either[CdfSparkException, Boolean] =
```
dmivankov, line 502: note: looks like propertyDefMap is empty when source=None so maybe it should be Option too and maybe bundled into a single type together with reference. Atm it looks odd that we match on source but don't really use it

Can be left for future improvements

dmivankov (approved):

src/main/scala/cognite/spark/v1/FlexibleDataModelRelationUtils.scala
Jacob-Eliat-Eliat: hmm, that's a good idea

Jacob-Eliat-Eliat: I'll leave it for later as you say, but I'll keep that in mind

PR #971, opened 2024-09-03
starting some cleanup

Jacob-Eliat-Eliat:


dmivankov (approved):
tests sources only, CI green, diff is more red lines than green - nice :tada:

PR #976, opened 2024-10-04
File content source

Jacob-Eliat-Eliat:


src/test/scala/cognite/spark/v1/SparkTest.scala
```diff
@@ -246,11 +246,11 @@ trait SparkTest {
 
   val updateAndUpsert: TableFor1[String] = Table(heading = "mode", "upsert", "update")
 
-  def getDefaultConfig(auth: CdfSparkAuth, projectName: String): RelationConfig =
+  def getDefaultConfig(auth: CdfSparkAuth, projectName: String, cluster: String = "api", applicationName: Option[String] = Some("SparkDatasourceTestApp")): RelationConfig =
```
dmivankov, line 249: nit: maybe it shouldn't be called "default config" anymore with new customization options

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,108 @@
...
+
+        val validUrl = for {
+          mimeType <- client.files.retrieveByExternalId(fileExternalId).map(_.mimeType)
+          downloadLink <- client.files
+            .downloadLink(FileDownloadExternalId(fileExternalId))
+            .map(_.downloadUrl)
+          isFileWithinLimits <- isFileWithinLimits(downloadLink)
+        } yield {
+          if (!isFileWithinLimits)
+            throw new CdfSparkException("File size above 5Gb limit")
```
dmivankov, line 54: nit: would be nice to use size from the constant, or omit specific value

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,108 @@
...
+          if (mimeType.isEmpty || !mimeType.contains("application/jsonlines"))
+            throw new CdfSparkException("Wrong mimetype. Expects application/jsonlines")
+          else
+            downloadLink
```
dmivankov, line 58:
```suggestion
          downloadLink
```
would make more sense probably - number of if-throw checks then return (nesting if-else tree could be alternative but would get ugly here)

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,108 @@
...
+        val validUrl = for {
+          mimeType <- client.files.retrieveByExternalId(fileExternalId).map(_.mimeType)
+          downloadLink <- client.files
+            .downloadLink(FileDownloadExternalId(fileExternalId))
+            .map(_.downloadUrl)
+          isFileWithinLimits <- isFileWithinLimits(downloadLink)
+        } yield {
+          if (!isFileWithinLimits)
+            throw new CdfSparkException("File size above 5Gb limit")
+          if (mimeType.isEmpty || !mimeType.contains("application/jsonlines"))
```
dmivankov, line 55: nit: maybe first check mimetype then size, in some sense it's more wrong to have incorrect type than go over size limit

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,108 @@
...
+        Array(CdfPartition(0))
+    }
+    import sparkSession.implicits._
+    sparkSession.read.json(rowsRdd.toDS())
+  }
+
+  private def readUrlContent(url: String): Stream[IO, String] =
+    Stream
+      .bracket(IO(new URL(url).openStream())) { inStream =>
+        IO(inStream.close()) // Ensure the InputStream is closed properly
```
dmivankov, line 81: we could probably use Resource class for that, but need to check

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,108 @@
...
+      }
+      .flatMap { inStream =>
+        val source = Source.fromInputStream(inStream)
+        Stream.fromIterator[IO](source.getLines(), chunkSize = 4096)
+      }
+
+  private def isFileWithinLimits(downloadUrl: String): IO[Boolean] = {
+    val backend = Resource.make(IO(HttpURLConnectionBackend()))(b => IO(b.close()))
+    val request = basicRequest.head(uri"$downloadUrl")
```
dmivankov, line 90: may consider having such method in the sdk :thinking: 
Also I guess everywhere those links are multi-use?

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,108 @@
...
+      .bracket(IO(new URL(url).openStream())) { inStream =>
+        IO(inStream.close()) // Ensure the InputStream is closed properly
+      }
+      .flatMap { inStream =>
+        val source = Source.fromInputStream(inStream)
+        Stream.fromIterator[IO](source.getLines(), chunkSize = 4096)
+      }
+
+  private def isFileWithinLimits(downloadUrl: String): IO[Boolean] = {
+    val backend = Resource.make(IO(HttpURLConnectionBackend()))(b => IO(b.close()))
```
dmivankov, line 89: might be nice to configure client with some extra headers (if sdk does that, or if that's a common practice)

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,108 @@
...
+    val request = basicRequest.head(uri"$downloadUrl")
+
+    backend.use { backend =>
+      IO {
+        val response = request.send(backend)
+        response.header("Content-Length").exists { sizeStr =>
+          sizeStr.toLong < sizeLimit
```
dmivankov, line 96: nit: if header is missing for some reason "file too big" error might be inaccurate, maybe we should fail the transform with a different message

dmivankov:
Only did a quick surface read yet, will later look closer at which spark methods are used etc

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,108 @@
...
+    with Serializable
+    with WithSizeLimit {
+
+  override val sizeLimit: Long = 5 * FileUtils.ONE_GB
+
+  @transient lazy val client: GenericClient[IO] =
+    CdpConnector.clientFromConfig(config)
+
+  override def schema: StructType =
+    createDataFrame(sqlContext.sparkSession).schema
```
dmivankov, line 36: will it read file in full? There's some streaming in how we fetch and pass around lines so might be better for mem consumption to do it twice, also should not be a problem for first iteration to read twice. But for bigger files future we might need to look into it (explicitly limit schema inference, or explicitly put schema as just jsons?)

src/test/scala/cognite/spark/v1/FileContentRelationTest.scala
```diff
@@ -0,0 +1,160 @@
...
+
+  it should "infer the schema" in {
+    val sourceDf: DataFrame = dataFrameReaderUsingOidc
+      .useOIDCWrite
+      .option("type", "filecontent")
+      .option("inferSchema", value = true)
+      .option("externalId", fileExternalId)
+      .load()
+    sourceDf.createOrReplaceTempView("fileContent")
+    sourceDf.schema.fields should contain allElementsOf Seq(
```
dmivankov, line 119: so looking at this test it will try to deduce schemas inside the lines. If we want to keep that, we still should have at least opt-out of schema inference (like in raw)

src/test/scala/cognite/spark/v1/FileContentRelationTest.scala
```diff
@@ -0,0 +1,160 @@
...
+        StructField("name", StringType, nullable = true),
+        StructField("test", StringType, nullable = true),
+        StructField("age", LongType, nullable = true)
```
dmivankov, line 122: maybe also be good to add tests with
- nested object types
- conflicting column types between rows

src/main/scala/cognite/spark/v1/FileContentRelation.scala
dmivankov: actually looks like there's
```
  def download(item: FileDownload, out: java.io.OutputStream): F[Unit] = {
```
but would be good to check how good it is wrt streaming and early closing

dmivankov: the sdk method above is for download, no method for HEAD request to get size yet

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,108 @@
...
+          None
+        )
+
+      }
+
+      override protected def getPartitions: Array[Partition] =
+        Array(CdfPartition(0))
+    }
+    import sparkSession.implicits._
+    sparkSession.read.json(rowsRdd.toDS())
```
dmivankov, line 75: interestingly this will try to infer schema once more inside, so `sparkSession.read.schema(schema).json(..)` might make sense instead

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  private def readUrlContent(url: String): Stream[IO, String] =
+    Stream
+      .bracket(IO(new URL(url).openStream())) { inStream =>
```
dmivankov, line 80: for extra safety might be good to check that URL schema is https (not file:// or other funny things)

src/main/scala/cognite/spark/v1/FileContentRelation.scala
dmivankov, line 89:
```suggestion
    val backend = Resource.fromAutoCloseable(IO(HttpURLConnectionBackend()))
```
nit: there's also this convenience method

dmivankov, line 89: HttpURLConnectionBackend is odd and doesn't quite fit here..
- it's synchronous so should be wrapped in IO.blocking if put into IO
- we do have shared sttpBackend most other places, though it could be hidden within sdk client, but some if it's construction isn't hidden in CdpConnection, so might borrow some stuff (if not switch entirely to sdk client)
- sdk client would also have some features that it lacks: tracing (we'll see only request going out to blob storage I guess), maybe some metrics, retry policies and rate limiters etc

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: I can't know for sure if they are multi use on all clusters unfortunately. It should be but we might encounter a bad surprise later.... head request is the only way to make this work, otherwize we can't know the size of the file without downloading it first.

The reason I didn't add it to the SDK quite yet is because it doesn't require credentials or anything here, so it's just a very basic request.

Jacob-Eliat-Eliat: Also when it comes to the download link, I'm pretty sure it didn't fit my usecase unfortunately so I didn't use it. I think my code changed since so I'll check if that still holds true.

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: I've been told not too check mimetype too closely so now I'm accepting empty mimetypes

Jacob-Eliat-Eliat: because they could be uploading the files via some other tool, without being able to control mimetype in the file service

Jacob-Eliat-Eliat: Also there is no standardized jsonlines mimetype, I can add some options, but impossible to be exhaustive

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: Yes, it will read it in full, as far as I understand it

src/main/scala/cognite/spark/v1/FileContentRelation.scala
dmivankov: There's already full upload&download support there https://github.com/cognitedata/cognite-sdk-scala/blob/master/src/main/scala/com/cognite/sdk/scala/v1/resources/files.scala#L160 so most of the plumbing should be in place and it may make PR tiny bit slimmer on datasource side. Won't be a blocker, but potentially may actually speed up review by splitting into 2 PRs and reusing some of SDK :slightly_smiling_face:

Jacob-Eliat-Eliat: I'm pretty ignorant about how the different backend interact and what they do I'll read up on it

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: sdk does it, I'll try to see if I can use that, I think I'll need a pr on the sdk

src/test/scala/cognite/spark/v1/FileContentRelationTest.scala
Jacob-Eliat-Eliat: Is it opt out in raw? I though that was a hack ^^

src/test/scala/cognite/spark/v1/SparkTest.scala
Jacob-Eliat-Eliat: Vast majority is still with default, just with some more override, also they have default value.  I think the spirit is still there ^^

src/test/scala/cognite/spark/v1/FileContentRelationTest.scala
dmivankov: opt-out definitely is in place and documented https://docs.cognite.com/cdf/integration/guides/transformation/write_sql_queries/#avoid-schema-inference the choice&errors are not always obvious to users and we don't have a solid recommendation for one or another, so it's not perfect. It also requires SQL rewrite (it may potentially be automated, the column name<>get_json_object part, but still the casting part needs to be encoded somehow so still needs more info to be present in SQL compared to inference-on mode)

dmivankov: For some cases no-inference is required to make transformation work despite being less user-friendly (well, more user-friendly in one aspect when there are type conflicts that can be handled via json rather than throwing)

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: Hmm, I agree and for the head request it's fairly straightforwards to just add a head method which returns the headers to the SDK. For the get method it's a bit more involved as we are using .openStream() but maybe we can do that more explicitely, I'll investigate

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: Unfortunately it's not that simple because the schema inference does it based on the RDD too, so that creates an infinite loop. I'll try to rethink it but I can't just do that :s

Jacob-Eliat-Eliat: Ok, I can just create the dataframe as a lazy val, and then either the schema or the scan will need it first and the other will just reuse the same dataframe.

src/test/scala/cognite/spark/v1/FileContentRelationTest.scala
Jacob-Eliat-Eliat: I think I found how to make it work :)

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: I replaced this entire stuff, I'm not sure if it's any better memory wize. I'm guessing we don't want to stream the response but instead read the entire thing then create a stream based on it, but I could be wrong, maybe we could stream before it finished reading.

Jacob-Eliat-Eliat: I guess I am wrong... it would be tricky to do using the current client though

src/test/scala/cognite/spark/v1/FileContentRelationTest.scala
Jacob-Eliat-Eliat: So, it should now be optional to infer the schema. Should I in addition attempt to flatten the schema? It should be possible as well

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: does this still make sense with the new client? It sounds to me like we are still not testing it so maybe

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: there is a setting to only read part of it if we want to (it's set as a percent)

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: I added a check

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: So in the end it's impossible to use the same client as everywhere else, because we use AsyncHttpClientCatsBackend which doesn't have streaming capabilities (see: AsyncHttpClientCatsBackend)

src/test/scala/cognite/spark/v1/FileContentRelationTest.scala
```diff
@@ -0,0 +1,248 @@
...
+    result.map(_.toSeq.toList) should contain theSameElementsAs
+      Seq(
+        List("Alice", null),
+        List("Bob", null),
+        List("Charlie", null),
+        List("Charlie2", "test")
+      )
+  }
+
+  it should "not infer schema is not asked to" in {
```
dmivankov, line 211: assertion on resulting schema is missing? Should still be able to check that it is "json"

src/test/scala/cognite/spark/v1/FileContentRelationTest.scala
```diff
@@ -0,0 +1,248 @@
...
+
+  it should "infer the schema correctly with conflicting types" in {
+    val sourceDf: DataFrame = dataFrameReaderUsingOidc
+      .useOIDCWrite
+      .option("type", "filecontent")
+      .option("inferSchema", value = true)
+      .option("externalId", fileExternalIdWithConflicts)
+      .load()
+    sourceDf.createOrReplaceTempView("fileContent")
+    sourceDf.schema.fields should contain allElementsOf Seq(
```
dmivankov, line 185: what do conflicts looks like and what happens to them, could there be an assert on that? or at least extra comment

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -141,6 +141,11 @@ class DefaultSource
         new EventsRelation(config)(sqlContext)
       case "files" =>
         new FilesRelation(config)(sqlContext)
+      case "filecontent" =>
+        val inferSchema = toBoolean(parameters, "inferSchema")
```
dmivankov, line 145: I guess we can add inferSchemaLimit parameter support later if needed (looking at RawTableRelation having it)

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+import org.apache.spark.sql.{DataFrame, Row, SQLContext, SparkSession}
+import org.apache.spark.{Partition, TaskContext}
+import sttp.capabilities.WebSockets
+import sttp.capabilities.fs2.Fs2Streams
+import sttp.client3.asynchttpclient.SttpClientBackendFactory
+import sttp.client3.asynchttpclient.fs2.AsyncHttpClientFs2Backend
+import sttp.client3.{SttpBackend, UriContext, asStreamUnsafe, basicRequest}
+import sttp.model.Header
+
+trait WithSizeLimit {
```
dmivankov, line 21: is this trait used somewhere or name has special meaning to spark? :eyes:

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+    for {
+      dispatcher <- Dispatcher.parallel[IO]
+      backend <- Resource.make(
+        IO(
+          AsyncHttpClientFs2Backend
+            .usingClient[IO](SttpClientBackendFactory.create("file content download"), dispatcher))
+      )(backend => backend.close())
+    } yield backend
+
+  private lazy val dataFrame: DataFrame = createDataFrame(sparkSession = sqlContext.sparkSession)
```
dmivankov, line 49:
```suggestion
  private lazy val dataFrame: DataFrame = createDataFrame
```
since sqlContext is a field and createDataFrame is non-static method, it could just take it directly? Doesn't seem like we put any other value in

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+            byteStream
+              .through(fs2.text.utf8.decode)
+              .through(fs2.text.lines)
+          case Left(error) =>
+            Stream.raiseError[IO](new Exception(s"Error while requesting underlying file: $error"))
+        }
+      }
+    }
+
+  private def isFileWithinLimits(downloadUrl: String): IO[Boolean] = {
```
dmivankov, line 126: nit: here's opportunity to return size & detailed error, so roughly `Either[Long, FileTooBig | FileSizeUnknown]`

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+    val headers: IO[Seq[Header]] = client.requestSession.head(uri"$downloadUrl")
+
+    headers.map { headerSeq =>
+      val sizeHeader = headerSeq.find(_.name.equalsIgnoreCase("content-length"))
```
dmivankov, line 130: didn't quickly find how it plays with compression (does sdk support it, would there be a header for unpacked size or for packed size), investigation todo

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+            if (!downloadLink.startsWith("https")) {
+              throw new CdfSparkException("File storage is not using https protocol")
+            }
+          )
+          isFileWithinLimits <- isFileWithinLimits(downloadLink)
+        } yield {
+          if (mimeType.isDefined && !Seq(
+              "application/jsonlines",
+              "application/x-ndjson",
+              "application/jsonl").contains(mimeType.get))
```
dmivankov, line 77: nit: extracting the list to constant would look nicer

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+      val maxParallelism = 1
+      override def compute(split: Partition, context: TaskContext): Iterator[String] = {
+
+        val validUrl = for {
+          mimeType <- client.files.retrieveByExternalId(fileExternalId).map(_.mimeType)
+          downloadLink <- client.files
+            .downloadLink(FileDownloadExternalId(fileExternalId))
+            .map(_.downloadUrl)
+          _ <- IO.pure(
+            if (!downloadLink.startsWith("https")) {
```
dmivankov, line 68: string prefix may potentially be deceiving (schema-less urls?), would make sense to parse URL first

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+
+  def createDataFrame(sparkSession: SparkSession): DataFrame = {
+    val rdd: RDD[String] = new RDD[String](sparkSession.sparkContext, Nil) with Serializable {
+
+      import cognite.spark.v1.CdpConnector.ioRuntime
+
+      val maxParallelism = 1
+      override def compute(split: Partition, context: TaskContext): Iterator[String] = {
+
+        val validUrl = for {
```
dmivankov, line 62: name suggests it's also validated to be URL, so let's do that too, and then download&checkSize functions can accept URL type (or Uri, or other analogue)

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+          mimeType <- client.files.retrieveByExternalId(fileExternalId).map(_.mimeType)
```
dmivankov, line 63: wonder if mimetype in metadata can diverge in files-api vs actual download, we already do HEAD request so it show show up there, we could add a log line if they mismatch to investigate later at the very least

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+          downloadLink <- client.files
+            .downloadLink(FileDownloadExternalId(fileExternalId))
+            .map(_.downloadUrl)
+          _ <- IO.pure(
```
dmivankov, line 67:
```suggestion
          _ <- IO.raiseWhen(
```
looks like there's a dedicated helper for that

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+          isFileWithinLimits <- isFileWithinLimits(downloadLink)
+        } yield {
+          if (mimeType.isDefined && !Seq(
+              "application/jsonlines",
+              "application/x-ndjson",
+              "application/jsonl").contains(mimeType.get))
+            throw new CdfSparkException("Wrong mimetype. Expects application/jsonlines")
+          if (!isFileWithinLimits)
+            throw new CdfSparkException(
+              f"File size above size limit, or file size header absent from head request")
```
dmivankov, line 81: nit: those checks&errors may look better in for{} with same .raiseWhen wrapping

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+          downloadLink
+        }
+
+        StreamIterator(
+          Stream
+            .eval(validUrl)
+            .flatMap(readUrlContent),
+          maxParallelism,
```
dmivankov, line 89:
```suggestion
          maxParallelism * 2,
```
maybe 1->2, SdkV1Rdd does that

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+              "application/x-ndjson",
+              "application/jsonl").contains(mimeType.get))
+            throw new CdfSparkException("Wrong mimetype. Expects application/jsonlines")
+          if (!isFileWithinLimits)
+            throw new CdfSparkException(
+              f"File size above size limit, or file size header absent from head request")
+          downloadLink
+        }
+
+        StreamIterator(
```
dmivankov, line 85: found that there's a `deduplicateRows = true` default argument, and that is is true for Raw which is odd as I thought it already makes sure to read in order preventing duplicates :thinking: something to look at, fine to keep in initial files implementation, maybe consider documenting it later (if that is visible externally much, I guess count can be seen as lower in case of duplicates)

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+      override protected def getPartitions: Array[Partition] =
+        Array(CdfPartition(0))
+    }
+    import sparkSession.implicits._
+    val dataset = rdd.toDS()
+
+    if (inferSchema) {
+      sqlContext.sparkSession.read.json(dataset)
+    } else {
+      dataset.toDF()
```
dmivankov, line 104: so this will produce single String as far as I can see, based on using `Rdd[String]`, which is different from Raw no-inference case. Both Raw columns and ndjson lines are expected to be Json, so here we should also be producing Json I guess. It isn't expected that we provide "file lines" support that aren't json

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+      }
+
+      override protected def getPartitions: Array[Partition] =
+        Array(CdfPartition(0))
+    }
+    import sparkSession.implicits._
+    val dataset = rdd.toDS()
+
+    if (inferSchema) {
+      sqlContext.sparkSession.read.json(dataset)
```
dmivankov, line 102: just to double-check, this will to read whole <=5gb file into current executor memory?

dmivankov: or maybe not until .collect :thinking:

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+    } else {
+      dataset.toDF()
+    }
+  }
+
+  private def readUrlContent(link: String): Stream[IO, String] =
```
dmivankov, line 108:
```suggestion
  private def readUrlContentLines(link: String): Stream[IO, String] =
```
nit: maybe use even more explicit name

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+    with PrunedFilteredScan
+    with Serializable
+    with WithSizeLimit {
+
+  override val sizeLimit: Long = 5 * FileUtils.ONE_GB
+
+  @transient lazy val client: GenericClient[IO] =
+    CdpConnector.clientFromConfig(config)
+
+  @transient private lazy val sttpFileContentStreamingBackendResource
```
dmivankov, line 38: note maybe at some point we can look into adding retries here and making it shared via CdpConnector, may save some overhead with connection reuse potentially. Also gzip support maybe, but depends a bit on whether backends support it and whether we'll be able to know uncompressed side in advance

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+    Stream.resource(sttpFileContentStreamingBackendResource).flatMap { backend =>
+      val request = basicRequest
+        .get(uri"$link")
+        .response(asStreamUnsafe(Fs2Streams[IO]))
+
+      Stream.eval(backend.send(request)).flatMap { response =>
+        response.body match {
+          case Right(byteStream) =>
+            byteStream
+              .through(fs2.text.utf8.decode)
```
dmivankov, line 118: looks fair, maybe we'll need to mention it in docs

dmivankov:
Main notes would be
- see what content-length does under compression
- change no schema inference String->Json

otherwise looks good to me

src/main/scala/cognite/spark/v1/DefaultSource.scala
Jacob-Eliat-Eliat: Yes, I'm still not totally certain about how it works under the hood, but I *think* as it is it isn't reading it twice, so it might not lead to any improvement.

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: no, just for testing...

src/test/scala/cognite/spark/v1/FileContentRelationTest.scala
Jacob-Eliat-Eliat: what do you mean? There is a test with conflicting schema, and string takes over.

Jacob-Eliat-Eliat: "infer the schema correctly with conflicting types" is the test

src/test/scala/cognite/spark/v1/FileContentRelationTest.scala
Jacob-Eliat-Eliat: added

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: It plays badly, I need to rethink this...

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: I'm not sure what would be sending the mimetype there to be honest. Maybe whatever tool they used to upload the file is setting the header, maybe not, but since this is less directly in user's hand maybe we should avoid checking for it.

To be honest I am hesitating to remove mimetype check altogether: I've been told there is no real guarantee on it.

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: cool

src/test/scala/cognite/spark/v1/FileContentRelationTest.scala
dmivankov: ouch sorry, I was sure when I looked at it that it was checking rows contents :facepalm: probably lines from other tests somehow were in view

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: Yeah, I mean you have to do it at some point right? There might be something that helps around how since the dataframe is backed by a stream, it might be able to partition before reading all of it? But My best guess is that it does. Worth experimenting a bit once this is in bluefield

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,141 @@
...
+    if (inferSchema) {
+      sqlContext.sparkSession.read.json(dataset)
+    } else {
+      dataset.toDF()
+    }
+  }
+
+  private def readUrlContent(link: String): Stream[IO, String] =
+    Stream.resource(sttpFileContentStreamingBackendResource).flatMap { backend =>
+      val request = basicRequest
```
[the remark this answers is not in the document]
Jacob-Eliat-Eliat: maybe a better idea is to start reading and use a trick like suggested here: https://softwaremill.community/t/limit-the-download-size-of-an-sttp-client/270 to limit the size?

Since content length does not work with gzip (see here: https://stackoverflow.com/questions/3819280/content-length-when-using-http-compression) that might be a better approach, or we could use both (if content-length > 5gb, don(t bother downloading, if it's < download but still enforce limit using the hack in the first link?).

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: We might never be able to know the size in advance.... 

Retries tuning can be another PR, does that make sense?

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: Hmm, I'm going to read a bit more on what this does, I *assumed* it had to be 1 so we wouldn't do several requests (since only 1 executor can possibly hold the request at the same time?) but I am not very well informed here so I4ll come back with an answer

src/main/scala/cognite/spark/v1/FileContentRelation.scala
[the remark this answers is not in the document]
dmivankov: worth a try is also doing HEAD request without gzip support, if server would send uncompressed length that way we're good

but otherwise maybe when we don't have info on file size indeed we could do limit how much we read from it (maybe on line stream level as ideally we should not damage the last line, or always drop last line for limit exceeded case)

src/main/scala/cognite/spark/v1/FileContentRelation.scala
dmivankov: yep

src/main/scala/cognite/spark/v1/FileContentRelation.scala
dmivankov: it can be anything I think, just a buffer size, with some code comments that we're neve able to fill it up to big sizes (maybe because for other resources reads are slow compared to that stream adaptor processing)

src/main/scala/cognite/spark/v1/FileContentRelation.scala
dmivankov: :white_check_mark:  maybe also add a comment that it's for tests

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -0,0 +1,140 @@
...
+      import cognite.spark.v1.CdpConnector.ioRuntime
+
+      val maxParallelism = 1
+      override def compute(split: Partition, context: TaskContext): Iterator[String] = {
+
+        val validUrl = for {
+          mimeType <- client.files.retrieveByExternalId(fileExternalId).map(_.mimeType)
+          downloadLink <- client.files
+            .downloadLink(FileDownloadExternalId(fileExternalId))
+            .map(_.downloadUrl)
```
dmivankov, line 69: it may be null? Not marked as required in https://api-docs.cognite.com/20230101/tag/Files/operation/downloadLinks#!c=200&path=items/0/downloadUrl&t=response
but otoh SDK already has some failure handling https://github.com/cognitedata/cognite-sdk-scala/blob/master/src/main/scala/com/cognite/sdk/scala/v1/resources/files.scala#L155
So might be worth checking how downloadlink behaves for not-yet uploaded files (or some other cases without download link?)

src/main/scala/cognite/spark/v1/FileContentRelation.scala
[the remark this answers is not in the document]
Jacob-Eliat-Eliat: We can skip the headers, but it's unpractical to actually test if it's implemented correctly since there is a bunch of different services depending on the cluster...

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: Noted, for documentation

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: Is it actually different?       StructField("columns", DataTypes.StringType) (RawTableRelation: 310) seems to imply to me it's returned as string

From the user POV in sql it's a single column named "value" which seems pretty consistent, in raw you'd get key, columns and lastUpdatedTime

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: done :)

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: ... there is a "uploaded" on the file api which returns true if it's uploaded. I can just use that I guess.

dmivankov: most likely it won't be null and service-contracts just haven't marked it as required, but double-checking with files-api code might still be good

src/main/scala/cognite/spark/v1/FileContentRelation.scala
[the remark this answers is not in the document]
dmivankov: The vendors list isn't that long, it would be good info on which ones support compression and which ones report content-length (and maybe depending on compression on/off). If we have the "size unknown" fallback that tries and fails if too big might be a good path, but even then it might be good to know how often it would fail first doing the read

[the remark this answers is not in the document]
Jacob-Eliat-Eliat: For S3: Sounds like it *should* work
For Gcloud, I stumbled upon this: https://cloud.google.com/storage/docs/transcoding#decompressive_transcoding

Which means it's possible to use decompressive transcoding to download a file that is actually stored as a gzip (this can happen when the content-encoding is set to gzip when uploading)

But when decompressive transcoding is used, the head request will not return a content length`

`When an object meets these two criteria, it undergoes decompressive transcoding when served, and the response containing the object does not contain a Content-Encoding or Content-Length header.`

..Back to the drawing board I guess :/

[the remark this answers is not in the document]
dmivankov: we also can start with runtime limiting and see the best way to handle compression&contentLength before making it GA

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: No I mean I checked what happens when the file is not uploaded and it's a 400 error. Which seems fair to expose: we first check the "uploaded" status in file api, then if it's set to true, we get the downloadlink here, and if somehow the uploaded is true but the downloadlink is not present (I guess it could happen if the file got deleted in between) we get a 400

dmivankov (approved):

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: Noted I should mention in the doc to use "get_json_object" on the "value" field and it should behave the same way

PR #984, opened 2024-11-06
remove hardcoded relation names

Jacob-Eliat-Eliat:
This is my attempt to use a constant name for each relations.

There are a could of quirks.. 

One relation doesn't have a proper class instead is a Builder, but you can feasably just give a companion object and a name to the builder.

In two different places we were doing a switch over all relations. I changed one to make everything go through the pattern matching instead of having an if-else + matching for the rest

You cannot properly enforce that a class has a companion object with a certain attribute in scala using types unfortunately. I did create a trait for it, but it doesn't really provide anything.

"shortname" already existed but I guess you'd need an instance for that to work.


src/main/scala/cognite/spark/v1/ThreeDModelsRelation.scala
```diff
@@ -11,7 +11,7 @@ import org.apache.spark.sql.{Row, SQLContext}
 final case class ModelItem(id: Long, name: String, createdTime: Long)
 
 class ThreeDModelsRelation(config: RelationConfig)(val sqlContext: SQLContext)
-    extends SdkV1Relation[ThreeDModel, Long](config, "threeDModels.read") {
```
Jacob-Eliat-Eliat, line 14: this shortName was different for once. I think this is only used for metric so the change should have no impact.

src/main/scala/cognite/spark/v1/ThreeDModelRevisionsRelation.scala
```diff
@@ -36,3 +36,7 @@ class ThreeDModelRevisionsRelation(config: RelationConfig, modelId: Long)(val sq
       client: GenericClient[IO]): Seq[Stream[IO, ThreeDRevision]] =
     Seq(client.threeDRevisions(modelId).list(config.limitPerPartition))
 }
+
+object ThreeDModelRevisionsRelation extends NamedRelation {
+  override val name: String = "3dmodelrevisions"
```
dmivankov, line 41: value is changing here, but looks like nothing depends on it in transformations atm :thinking:

dmivankov: Maybe support wasn't added to transformations yet https://docs.cognite.com/cdf/integration/guides/transformation/write_sql_queries/ - not listed here

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -144,3 +144,7 @@ class FileContentRelation(config: RelationConfig, fileExternalId: String, inferS
     }
 
 }
+
+object FileContentRelation extends NamedRelation {
+  override val name: String = "filecontent"
```
dmivankov, line 149: looks like we have a todo for file content - add datasource metrics like others, make it extend CdfRelation

src/main/scala/cognite/spark/v1/ThreeDModelRevisionsRelation.scala
dmivankov: does match the datasource `case` string, so better to stick to that, only shortname for metrics changes here like in the other 3d file :+1:

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
```diff
@@ -379,6 +379,10 @@ object PushdownUtilities {
   }
 }
 
+trait NamedRelation {
```
dmivankov, line 382: nit: looks unrelated to pushdown. Also, maybe make CdfRelation extend NamedRelation, they can use the same shortname and it makes it more likely that named part is applied to new source types

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -188,104 +188,108 @@ class DefaultSource
...
+      case relation: AssetHierarchyBuilder =>
+        config.onConflict match {
           case OnConflictOption.Delete =>
-            relation.delete
+            relation.delete(data)
+          case _ =>
+            relation.buildFromDf(data)
         }
+        relation
+      case relation: CdfRelation with WritableRelation =>
```
dmivankov, line 245: It looks to be on same level as cases by relation types so it will not be applied if any of previous ones match
May need to open in IDE old&new to double-check though

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -188,104 +188,108 @@ class DefaultSource
...
 
-        val operation: Seq[Row] => IO[Unit] = config.onConflict match {
-          case OnConflictOption.Abort =>
-            relation.insert
-          case OnConflictOption.Upsert =>
-            relation.upsert
-          case OnConflictOption.Update =>
-            relation.update
+    relation match {
+      case relation: DataPointsRelationV1[_] =>
```
dmivankov, line 223: nit: while at prettifying maybe assets&datapoints could also be extracted into functions like createSequenceRows and createFlexibleDataModelRelation :thinking:

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
Jacob-Eliat-Eliat: This trait is for Objects so it doesn't really work, AFAIK you can't force classes to have companion objects with a variable

src/main/scala/cognite/spark/v1/DefaultSource.scala
Jacob-Eliat-Eliat: Think you are misreading this? there is two pattern matching statements now.

First ones does the case by case for each source type and create the relation

Second one does some handling depending on the type of relation, there is special handling for both Datapoints and ... I forgot which one and all the rest go through this case here

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
dmivankov: right, makes sense

src/main/scala/cognite/spark/v1/DefaultSource.scala
dmivankov: could be, in web UI it is a bit confusing to see the nesting, will try in a proper UI

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: Ah yes, this is the companion object but you are right here :s

src/main/scala/cognite/spark/v1/ThreeDModelRevisionsRelation.scala
Jacob-Eliat-Eliat: Ah yes, I mentionned the other one but forgot this one changes too. It does look right though!

src/main/scala/cognite/spark/v1/DefaultSource.scala
Jacob-Eliat-Eliat: There is definitely more to do here, but let's keep the scope limited

src/main/scala/cognite/spark/v1/DefaultSource.scala
Jacob-Eliat-Eliat: Definitely :)

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: Will do in another PR if that's ok

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
Jacob-Eliat-Eliat: Ok, I did the latter, now both the companion object and the relation itself extend namedRelation, I'm not sure if I like it better or not but it can't hurt

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: Will do in this PR, sounds like it's easy.

src/main/scala/cognite/spark/v1/DefaultSource.scala
Jacob-Eliat-Eliat: Ah you mean it won't be applied for Datapoints and AssetHierarchyBuilder? That is true, but it's the same as before, as it was before nested in the "else" statement

Jacob-Eliat-Eliat: Maybe it should be applied in these cases too, but that would be a behavior changes and should be looked at separately.

dmivankov: yep, for this PR best to keep existing behavior

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -188,104 +188,108 @@ class DefaultSource
...
+            .parEvalMapUnordered(maxParallelism) { chunk =>
+              operation(chunk.toVector)
+            }
+            .compile
+            .drain
+            .unsafeRunSync()
+        })
+        relation
+      case _ =>
+        sys.error(s"Resource type $resourceType does not support save")
```
dmivankov, line 292:
```suggestion
        sys.error(s"Resource type $resourceType does not support save()")
```
maybe keep existing message

src/main/scala/cognite/spark/v1/DefaultSource.scala
dmivankov, line 245: hm, a little confusing naming from old code DefaultSource.createRelation is actually createRelationForWriting or something like it. But indeed it was checking against a list of relations that are supposed to support writing, now we check it via a trait so overall looks ok. By a quick look there were no writable sources not covered before so logic should be same as before :+1:

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -188,104 +188,108 @@ class DefaultSource
...
+            import CdpConnector.ioRuntime
+            val batches = rows.grouped(batchSize).toVector
+            batches.parTraverse_(relation.delete).unsafeRunSync()
+          })
+        } else {
+          // datapoints need special handling of dataframes and batches
+          relation.insert(data, overwrite = true)
+        }
+        relation
+      case relation: AssetHierarchyBuilder =>
```
dmivankov, line 237:
```suggestion
      case relation: AssetHierarchyBuilder with WritableRelation =>
```
looks like in spirit it should also be checked (is it valid Scala syntax?)

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -188,104 +188,108 @@ class DefaultSource
...
+      case StringDataPointsRelation.name =>
+        new StringDataPointsRelationV1(config)(sqlContext)
+      case EventsRelation.name =>
+        new EventsRelation(config)(sqlContext)
+      case TimeSeriesRelation.name =>
+        new TimeSeriesRelation(config)(sqlContext)
+      case AssetsRelation.name =>
+        new AssetsRelation(config)(sqlContext)
+      case FilesRelation.name =>
+        new FilesRelation(config)(sqlContext)
```
dmivankov, line 206: I guess that we handle it here and don't bail with "doesn't support save()" might be ok in presence of another check later on, but looks a bit odd

dmivankov: nevermind, that's files metadata, not contents

src/main/scala/cognite/spark/v1/DefaultSource.scala
Jacob-Eliat-Eliat: This is also necessary to be able to call some stuff below (I believe the schema stuff notably)

Used to be this was in the "else" and then the typing of relation was enforced by the pattern matching, but in this case, we are pattern matching again on relation, which can be a AssetHierarchyBuilder which IIRC is not marked as writable, or something like that.

Basically some typing issue force this to be here, but it also makes perfect sense with the below code.

dmivankov (approved):
Looks like it should preserve existing behavior :+1:

src/main/scala/cognite/spark/v1/DefaultSource.scala
Jacob-Eliat-Eliat: It doesn't implement it, but maybe it should ^^

Jacob-Eliat-Eliat: Hmm, probably it should, I'll add it I guess

dmivankov: oh, that's annoying, maybe we need a more fine-grained trait then to make it super clean

Jacob-Eliat-Eliat: ah it's even worse.

It implements insert/delete/update but the signature is different (doesn't take row, takes vector of AssetUpdateSchema/AssetInsertSchema etc...) room for improvement here. I was looking at fixing this but looks too annoying to be in the same PR at elast

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -188,104 +188,108 @@ class DefaultSource
...
+        new LabelsRelation(config)(sqlContext)
+      case SequenceRowsRelation.name =>
+        createSequenceRows(parameters, config, sqlContext)
+      case RelationshipsRelation.name =>
+        new RelationshipsRelation(config)(sqlContext)
+      case DataSetsRelation.name =>
+        new DataSetsRelation(config)(sqlContext)
+      case FlexibleDataModelBaseRelation.name =>
+        createFlexibleDataModelRelation(parameters, config, sqlContext)
+      case _ => sys.error(s"Resource type $resourceType does not support save()")
```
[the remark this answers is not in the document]
Jacob-Eliat-Eliat: Agreed here, but it's due to the weird traits implementation of AssetHierarchyBuilder, or lack thereof, need separate work, and it's probably a pain, there must be a reason it's done this way.

PR #985, opened 2024-11-07
Unify schema management

Jacob-Eliat-Eliat:
Currently half of schemas are managed in datasource with unused traits, the other half in SchemaService.

This would allow doing something like

    (conflictMode, object) match {
      case ("upsert", relationWithSchema: UpsertSchema) =>
        Right(relationWithSchema.upsertSchema)
...

Because now all schemas live in the same place.

The minus is that it definitely feels like too many traits.

to note: insertSchema was unused, while abortSchema was always equals to insertSchema, so I renamed insertSchema => abortSchema. (edit dmitry pointed out insertSchema might be more explicit, so I changed it back to insertSchema everywhere)

The reasoning behind this PR:
  * we no longer rely on long list of which relations implement which schema 
  * we unify where the static schemas are located
  * use traits to simplify discovering which relationship implement which schema types.
  * this is actually removing code in total I think, because in jetfire we can basically delete everything in schemaService. Note that there is even a //TODO in schemaService indicating that whoever did it also though it needed work (his complain is about repeated code, which will be adressed by this along with other issues)

For file content, we need to modify how the schema endpoints work a little bit which is ugly if this isn't here.

NB: these are not the dynamic schema, they are all constant. To get the actual schema, in jetfire, it goes through more processing using this but also schema coming from actual schema inference.

You can see in this WIP PR: https://github.com/cognitedata/jetfire-backend/pull/2526/files that it enables deleting a fair amount of code in the schema part of jetfire


src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
```diff
@@ -418,9 +418,10 @@ class AssetHierarchyBuilder(config: RelationConfig)(val sqlContext: SQLContext)
     }
 }
 
-object AssetHierarchyBuilder extends NamedRelation {
+object AssetHierarchyBuilder extends NamedRelation with UpsertSchema with DeleteWithIdSchema {
```
dmivankov, line 421: currently there's no compile-time link between this exposed schema and Row->schematype conversion in datasource operations so they can potentially diverge. Additionally they do seem to diverge already :sadpanda: so DeleteItemByCogniteId used by EventsRelation.delete will accept either `id` or `externalId`

A step towards linking it together could be adding more type parameters to WritableRelation and others, but there might be all the edge cases that may break or quirks that need to be kept

But overall I think it makes sense that datasource lib should hold the knowledge of schemas, this incremental step should be fine to do

src/main/scala/cognite/spark/v1/AssetsRelation.scala
```diff
@@ -116,13 +116,19 @@ class AssetsRelation(config: RelationConfig, subtreeIds: Option[List[CogniteId]]
   override def uniqueId(a: AssetsReadSchema): Long = a.id
 }
 
-object AssetsRelation extends UpsertSchema with NamedRelation {
+object AssetsRelation
+    extends UpsertSchema
+    with ReadSchema
+    with NamedRelation
+    with AbortSchema
```
dmivankov, line 123: Abort naming is legacy, but a bit confusing, it seems it is actually InsertOrFail/Insert (as opposed to Upsert), not sure if it makes sense to rename in this PR or maybe later

Jacob-Eliat-Eliat: Think separate job. Too much stuff leftover for file-content, I don't wanna get sidetracked too much.

obviously it'd be trivial to rename here, but I think in jetfire the same naming exists so I'd need to do it there too.

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
dmivankov: Another thought is it seems schema isn't really checked or fully used anywhere :thinking: it is exposed, there's some validation that schema exists, but compliance to schema is up to each relation and Row->schematype parsing/conversion

src/main/scala/cognite/spark/v1/AssetsRelation.scala
Jacob-Eliat-Eliat: Sure Can rename to insert.

src/main/scala/cognite/spark/v1/fdm/FlexibleDataModelBaseRelation.scala
```diff
@@ -583,7 +585,13 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
     }
 }
 
-object FlexibleDataModelBaseRelation extends NamedRelation {
+object FlexibleDataModelBaseRelation
+    extends NamedRelation
+    with UpsertSchema
+    with DeleteWithExternalIdSchema {
```
dmivankov, line 591: might well be a bug but schema in https://github.com/cognitedata/jetfire-backend/blob/master/driver/src/main/scala/com/cognite/jetfire/driver/SchemaService.scala#L102 is full object, also per-fdm-type rather than shared via base. And FlexibleDataModelConnectionRelation is using something wider than just external id too

Jacob-Eliat-Eliat: Aren't they all using the exact same predefined type?

Jacob-Eliat-Eliat: by the way this predefined type is actually   val dataModelInstanceSchema = StructType(Seq(StructField("externalId", DataTypes.StringType))) not full object, or am I missing something here?.

src/main/scala/cognite/spark/v1/AssetHierarchyBuilder.scala
Jacob-Eliat-Eliat: Hmm, yes, these schemas could potentially be deleted eventually but they are used as-is in SchemaService. We could consider eventually reworking the jetfire side of things and maybe remove this flow completely but let's keep the scope limited

src/main/scala/cognite/spark/v1/fdm/FlexibleDataModelBaseRelation.scala
dmivankov: Hmm, got blinded by `.dataModelInstanceSchema` a bit, so that indeed remains, but FlexibleDataModelConnectionRelation.delete goes via FlexibleDataModelRelationUtils.createEdgeDeleteData and also takes spaceId if available in row. So I guess it's then a bug in exposed schema :thisisfine: that we can look at in a separate PR :+1:

Jacob-Eliat-Eliat:
Realized I was missing updateSchema, so now that's added too. Since it's always the same as upsertSchema but not defined for all relations that have an upsertSchema, I used created a trait UpdateSchemaFromUpsertSchema which does that by itself.

Jacob-Eliat-Eliat:
I'm aware that a larger rework around this part of the code could help things further, but I'm trying to do enough to improve things without blocking myself completely from advancing in my tasks (file content support).

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
```diff
@@ -395,6 +396,28 @@ trait UpdateSchema {
   val updateSchema: StructType
 }
 
-abstract class DeleteSchema {
-  val deleteSchema: StructType = StructType(Seq(StructField("id", DataTypes.LongType)))
+trait UpdateSchemaFromUpsertSchema extends UpsertSchema with UpdateSchema {
+  override val updateSchema: StructType = upsertSchema
 }
```
dmivankov, line 401: If I'm not missing anything, UpdateSchema trait isn't used anywhere, so this new one shouldn't be needed too and both can be dropped

Otherwise I'm curious about the diffrerences between update&upsert schemas. By the name it seems that maybe one is keeping unspecified (null?) fields and other resets all fields, but then schema is fixed and always lists all fields? :thinking:

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
```diff
@@ -395,6 +396,28 @@ trait UpdateSchema {
...
+  override val deleteSchema: StructType = StructType(Seq(StructField("id", DataTypes.LongType)))
+}
+
+trait DeleteWithExternalIdSchema extends DeleteSchema {
+  import cognite.spark.compiletime.macros.StructTypeEncoderMacro._
+
+  override val deleteSchema: StructType = structType[DeleteSchemaWithExternalId]()
+}
+
+final case class DeleteSchemaWithExternalId(
```
dmivankov, line 421:
```suggestion
final case class DeleteByExternalId(
```
or some other name, DeleteWithExternalIdSchema vs DeleteSchemaWithExternalId look too similar yet are slightly different things

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
```diff
@@ -395,6 +396,28 @@ trait UpdateSchema {
...
+trait ReadSchema {
+  val readSchema: StructType
+}
+
+trait DeleteSchema {
+  val deleteSchema: StructType
+}
+
+trait DeleteWithIdSchema extends DeleteSchema {
+  override val deleteSchema: StructType = StructType(Seq(StructField("id", DataTypes.LongType)))
```
dmivankov, line 412: Would make sense to extract a DeleteByInternalId/DeleteById class too to be more uniform, even if currently delete() code in some places uses DeleteItemByCogniteId union type
Maybe reuse DeleteItem class here, or in some other way make traits here use same pattern

Full move of id/externalId/eitherOfThem to a single deduplicated set of classes in datasource can be a separate PR though

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
dmivankov: oh, I think I can see how update<>upsert may be different, just that currently trait isn't used and all existing update schemas choose to be same as upsert schema

src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -298,12 +298,14 @@ class RawTableRelation(
   }
 }
 
-object RawTableRelation extends NamedRelation {
+object RawTableRelation extends NamedRelation with UpsertSchema {
```
dmivankov, line 301: also should have ReadSchema?

src/main/scala/cognite/spark/v1/fdm/FlexibleDataModelBaseRelation.scala
```diff
@@ -583,7 +585,13 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
...
+  override val upsertSchema: StructType = StructType(
+    Seq(StructField("externalId", DataTypes.StringType)))
```
dmivankov, line 593: I think it makes sense to put custom deleteSchema here with a comment that is potentially looking incorrect (missing space or even more fields)

src/main/scala/cognite/spark/v1/RawTableRelation.scala
Jacob-Eliat-Eliat: Readschema is not in use. I kept it where it is currently but didn't add it anywhere.

Since there is no val ReadSchema in RawTableRelation I didn't add it. I think by default we use abortSchema when reading but I haven't checked, might be upsertSchema as well.

When I add file content schema that might become relevant.

src/main/scala/cognite/spark/v1/SequenceRowsRelation.scala
```diff
@@ -272,8 +272,17 @@ class SequenceRowsRelation(config: RelationConfig, sequenceId: CogniteId)(val sq
   }
 }
 
-object SequenceRowsRelation extends NamedRelation {
+object SequenceRowsRelation extends NamedRelation with UpsertSchema with DeleteSchema {
```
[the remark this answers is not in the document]
Jacob-Eliat-Eliat: Same comment as for RAW. It's not used. Kept as-is in this PR, just added the trait where it was already.

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
Jacob-Eliat-Eliat: The difference is that not all Relation support update schema. There is different handling in SchemaService, where some don't support both upsert and update and some do, which I think can make sense if the relation doesn't support updating?

src/main/scala/cognite/spark/v1/PushdownUtilities.scala
Jacob-Eliat-Eliat: Kind of agree, but as you say it complexifies the PR a bit and should probably be separate. The other schema was more straightfoward they were all using externalId only

Jacob-Eliat-Eliat:
So If I gather the stuff that should be done in addition to this in potentially future PR:

 * Rework of delete by id/delete by both
 * Fix or delete ReadSchema (potentially this might have to be done before doing file content support)
 * Custom Schema for FDM is wrong?
 * Check if it makes sense to keep all three of insert/upsert/update

dmivankov:
One thing that might even be a good idea to put in code comments is how those traits are supposed to be used
- do they fully indicate which operations each resource supports
- trait <> spark operation correspondence
- note on meaning of schema structtype
- any other notes

It gets a little tricky with reading trait as it isn't used (or is it?) to determine if resource supports read or with which schema

Jacob-Eliat-Eliat:
> One thing that might even be a good idea to put in code comments is how those traits are supposed to be used
> 
>     * do they fully indicate which operations each resource supports
> 
>     * trait <> spark operation correspondence
> 
>     * note on meaning of schema structtype
> 
>     * any other notes
> 
> 
> It gets a little tricky with reading trait as it isn't used (or is it?) to determine if resource supports read or with which schema

I guess the real answer for the read trait is that they indicate which schema are supported in jetfire... except ReadSchema which jetfire doesn't support but I also don't want to remove it as it might be a good fundation to support read-only transformations in the future (ditch abortSchema in jetfire when checking the source and use read-schema instead).

I can leave //These traits indicate which predefined schema are available for each relation

dmivankov:
Another thought was to maybe have a set of traits that define operations only regardless of schemas, and then (extensions or just other traits) that expose schemas, buut it may get to convoluted..

dmivankov:
> I don't think right now anything in datasource is defining which operations are supported, this is only about schema. but maybe we can eventually get this "readableDatasource"/"writableDatasource" to be in cdp-spark-datasource instead of jetfire... or in both.

So initial motivation was to use those traits to check which resources are read-only but not it shifted to a smaller goal of making (write) schema visible on datasource level? Sounds fair, but even more reason to put a comment on traits about what are they for.

And then jetfire change for now would be scoped to schema? Also sounds safe enough, will need to reframe how to look at the PR and look again :+1:

Jacob-Eliat-Eliat:
> > I don't think right now anything in datasource is defining which operations are supported, this is only about schema. but maybe we can eventually get this "readableDatasource"/"writableDatasource" to be in cdp-spark-datasource instead of jetfire... or in both.
> 
> So initial motivation was to use those traits to check which resources are read-only but not it shifted to a smaller goal of making (write) schema visible on datasource level? Sounds fair, but even more reason to put a comment on traits about what are they for.
> 
> And then jetfire change for now would be scoped to schema? Also sounds safe enough, will need to reframe how to look at the PR and look again 👍

Ah no, this was always scoped to schema actually. The readable/writable datasource is a separate thing and has been jetfire-only from the start.

The "indirect" problem this is trying to address for file content is that the schema part of jetfire is poorly implemented which meant that trying to make changes is a mess. 

In particular it's not super clear what is actually necessary for schema to work (even more specific: the flatten/unflatten workflow is used there, but I think it's not necessary).

The other part of this is that I'm aware filecontent will need a schema, but it's not clear which schema to use since it's read only and jetfire doesn't use readSchema (yet). With this cleanup it becomes much easier to implement.

I admit this is a very loose relation to filecontent, which is why I ended up reworking my filecontent PR to not need this at all in accordance with your suggestion.

src/main/scala/cognite/spark/v1/fdm/FlexibleDataModelBaseRelation.scala
```diff
@@ -583,7 +586,20 @@ abstract class FlexibleDataModelBaseRelation(config: RelationConfig, sqlContext:
...
 
-object FlexibleDataModelBaseRelation extends NamedRelation {
+object FlexibleDataModelBaseRelation extends NamedRelation with UpsertSchema with DeleteSchema {
+  import cognite.spark.compiletime.macros.StructTypeEncoderMacro._
+  import cognite.spark.compiletime.macros.SparkSchemaHelper.structType
+
+  // TODO: this seems to be incomplete or fully wrong schema
+  final case class DataModelBaseRelationUpsertSchema(externalId: String) {}
+
+  override val upsertSchema: StructType = structType[DataModelBaseRelationUpsertSchema]()
```
dmivankov, line 596: ugh, compile error here, I'll investigate&fix

dmivankov (approved):

PR #986, opened 2024-11-13
Support optional filtering of null values from RAW

thorkildcognite:
This change exposes the Raw-API support for filtering nulls server side through the cdp data source. It does not impact schema interference, as that is still done without filtering.

### DIFF
--- a/src/main/scala/cognite/spark/v1/DefaultSource.scala
+++ b/src/main/scala/cognite/spark/v1/DefaultSource.scala
@@ -478,8 +478,8 @@ object DefaultSource {
       enableSinglePartitionDeleteAssetHierarchy = enableSinglePartitionDeleteAssetHierarchy,
       tracingParent = extractTracingHeadersKernel(parameters),
       useSharedThrottle = toBoolean(parameters, "useSharedThrottle", defaultValue = false),
-      filterNullsOnNonSchemaQueries =
-        toBoolean(parameters, "filterNullsOnNonSchemaQueries", defaultValue = false)
+      serverSideFilterNullValuesOnNonSchemaRawQueries =
+        toBoolean(parameters, "filterNullFieldsOnNonSchemaRawQueries", defaultValue = false)
     )
   }
 
--- a/src/main/scala/cognite/spark/v1/DefaultSource.scala
+++ b/src/main/scala/cognite/spark/v1/DefaultSource.scala
@@ -477,7 +477,9 @@ object DefaultSource {
       rawEnsureParent = toBoolean(parameters, "rawEnsureParent", defaultValue = true),
       enableSinglePartitionDeleteAssetHierarchy = enableSinglePartitionDeleteAssetHierarchy,
       tracingParent = extractTracingHeadersKernel(parameters),
-      useSharedThrottle = toBoolean(parameters, "useSharedThrottle", defaultValue = false)
+      useSharedThrottle = toBoolean(parameters, "useSharedThrottle", defaultValue = false),
+      filterNullsOnNonSchemaQueries =
+        toBoolean(parameters, "filterNullsOnNonSchemaQueries", defaultValue = false)
     )
   }
 
--- a/src/main/scala/cognite/spark/v1/RawTableRelation.scala
+++ b/src/main/scala/cognite/spark/v1/RawTableRelation.scala
@@ -66,6 +66,7 @@ class RawTableRelation(
           filter = RawRowFilter(),
           requestedKeys = None,
           schema = None,
+          filterNulls = false,
           collectMetrics = collectSchemaInferenceMetrics,
           collectTestMetrics = false
         )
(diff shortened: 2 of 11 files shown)


src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -477,7 +477,9 @@ object DefaultSource {
       rawEnsureParent = toBoolean(parameters, "rawEnsureParent", defaultValue = true),
       enableSinglePartitionDeleteAssetHierarchy = enableSinglePartitionDeleteAssetHierarchy,
       tracingParent = extractTracingHeadersKernel(parameters),
-      useSharedThrottle = toBoolean(parameters, "useSharedThrottle", defaultValue = false)
+      useSharedThrottle = toBoolean(parameters, "useSharedThrottle", defaultValue = false),
+      filterNullsOnNonSchemaQueries =
+        toBoolean(parameters, "filterNullsOnNonSchemaQueries", defaultValue = false)
```
dmivankov, line 482: I guess it should be raw-specific like `rawOptimizeNullTransfer` or something? May depend on whether is is purely an optimization or can be visible in the results. If can be visible then mostly only applies to dynamic schemas like raw&filecontent (less so for datamodels?) - if we want to expose pruned json outputs

dmivankov: another option is name it `experimentalFilterRawNulls` so that we don't apply it by default yet and can revisit naming&semantic later after experiments

thorkildcognite: It is very raw specific, so naming it something in that nature makes sense! I'll find a new name for it.

It will not be visible for the users (the dataframe is the same). So at some point we might just make it default.

src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -83,15 +84,17 @@ class RawTableRelation(
...
       numPartitions: Int)(client: GenericClient[IO]): Seq[Stream[IO, RawRow]] = {
     assert(numPartitions == cursors.length)
-    val rawClient = client.rawRows(database, table)
+    val rawClient = client.rawRows(database, table, filterNullFields)
     cursors.map(rawClient.filterOnePartition(filter, _, limit))
   }
 
   private def getStreamByKeys(client: GenericClient[IO], keys: Set[String]): Stream[IO, RawRow] = {
+    // Note that retrieveByKey does not currently support filtering out null fields. When/If that is
+    // added, we should also pass in the flag to filter out those here.
```
dmivankov, line 97: column filtering param is also missing for it, at least in the docs https://api-docs.cognite.com/20230101/tag/Raw/operation/getRow

src/main/scala/cognite/spark/v1/DefaultSource.scala
thorkildcognite: I've added the new name to README.md too, and specified that it will probably become default in the future.

src/test/scala/cognite/spark/v1/RawTableRelationTest.scala
```diff
@@ -894,4 +909,40 @@ class RawTableRelationTest
     err.getMessage shouldBe "Error while loading RAW row [key='k'] in column 'value': java.lang.NumberFormatException: For input string: \"test\""
 
   }
+
+  it should "filter out fields with null value but not impact schema inference" in {
```
dmivankov, line 913: let's also add test for no-null-filtering case

thorkildcognite: As if it isn't enabled, it should behave as before? Sure, can do that.

dmivankov: Ah, I think what I actually wanted is no-inference case with null filtering and without, both outcomes of there being a diff and no diff will be interesting.

For inference case not a lot to see looking at resulting dataframe, but otoh extra small test never hurts

dmivankov (approved):
As long as it is only under parameter also open to merging in current shape, looking into no-inference case closer at a later point when we consider enabling it by default

src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -127,6 +130,7 @@ class RawTableRelation(
       filter: RawRowFilter,
       requestedKeys: Option[Set[String]],
       schema: Option[StructType],
+      filterNulls: Boolean,
       collectMetrics: Boolean = config.collectMetrics,
       collectTestMetrics: Boolean = config.collectTestMetrics): RDD[Row] = {
     val configWithLimit =
```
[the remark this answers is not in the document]
thorkildcognite: I've added a test that verifies that the old behavior is still there if we don't enable it. If filtering is enabled, then it will be filtered, though, but that is kind of the point.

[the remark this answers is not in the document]
thorkildcognite: Also added one to verify that setting it to false and not set it is the same. That should cover most bases.

PR #991, opened 2024-11-15
change timeout

Jacob-Eliat-Eliat:
Default timeout is too short. Since we use single request with streaming, it takes forever... TBH should have done part reading to be able to partition from the start but for now this is a sensible change.


src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -42,7 +42,9 @@ class FileContentRelation(config: RelationConfig, fileExternalId: String, inferS
       backend <- Resource.make(
         IO(
           AsyncHttpClientFs2Backend
-            .usingClient[IO](SttpClientBackendFactory.create("file content download"), dispatcher))
+            .usingClient[IO](
+              SttpClientBackendFactory.create("file content download", Some(3600000)),
```
dmivankov, line 46: maybe also accept customization via relation options

src/main/scala/cognite/spark/v1/SttpClientBackendFactory.scala
```diff
@@ -6,7 +6,7 @@ import org.asynchttpclient.AsyncHttpClient
 import sttp.client3.SttpBackendOptions
 
 object SttpClientBackendFactory {
-  def create(prefix: String = "Cdf-Spark"): AsyncHttpClient = {
+  def create(prefix: String = "Cdf-Spark", requestTimeout: Option[Int] = None): AsyncHttpClient = {
```
dmivankov, line 9:
```suggestion
  def create(prefix: String = "Cdf-Spark", requestTimeoutMs: Option[Int] = None): AsyncHttpClient = {
```
or even better maybe some typed duration

dmivankov (approved):

src/main/scala/cognite/spark/v1/FileContentRelation.scala
Jacob-Eliat-Eliat: Not sure what you mean, this backend is only used for file download.

Or do you mean in the SttpClientBackendFactory? I'm not sure that should be aware of the relation

PR #996, opened 2024-11-25
Character line size limit

Jacob-Eliat-Eliat:
There is a debate between using byte size and character size.

This is my preferred approach because:
 * while it's easy for a customer to check the byte size of a file, it's less straightforwards with a line, because it depends on encoding, and the file won't report it to them directly.
 * it avoids unnecessary processing and uses ready-made fs2 methods

here is a draft of the alternative to give you an idea of what the flow would be like:
https://github.com/cognitedata/cdp-spark-datasource/pull/995

(not if the alternative is preferred, the processing code needs a heavy coat of cleanup but the main idea is there)

### DIFF
--- a/src/main/scala/cognite/spark/v1/FileContentRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FileContentRelation.scala
@@ -122,7 +122,7 @@ class FileContentRelation(config: RelationConfig, fileExternalId: String, inferS
               .handleErrorWith {
                 case e: fs2.text.LineTooLongException =>
                   throw new CdfSparkException(
-                    s"""Line too long in file with external id: "$fileExternalId" SizeLimit in characters: $lineSizeLimitCharacters""",
+                    s"""Line too long in file with external id: "$fileExternalId" SizeLimit in characters: $e.max, but $e.length characters accumulated""",
                     e)
                 case other =>
                   throw other
--- a/src/main/scala/cognite/spark/v1/FileContentRelation.scala
+++ b/src/main/scala/cognite/spark/v1/FileContentRelation.scala
@@ -36,7 +36,7 @@ class FileContentRelation(config: RelationConfig, fileExternalId: String, inferS
     with WithSizeLimit {
 
   override val fileSizeLimitBytes: Long = 5 * FileUtils.ONE_GB
-  //We enforce a 2.5MB limit per line. In jvm each character takes 2bytes so we divide by two to get the limit in characters.
+  //We enforce an arbitrarily chosen 2.5MB JVM mem limit per line. In jvm each character takes 2bytes so we divide by two to get the limit in characters.
   private val lineSizeLimitBytes: Int = (2.5 * FileUtils.ONE_MB).toInt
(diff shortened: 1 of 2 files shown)


src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -34,7 +35,10 @@ class FileContentRelation(config: RelationConfig, fileExternalId: String, inferS
     with Serializable
     with WithSizeLimit {
 
-  override val sizeLimit: Long = 5 * FileUtils.ONE_GB
+  override val fileSizeLimitBytes: Long = 5 * FileUtils.ONE_GB
+  //We enforce a 2.5MB limit per line. In jvm each character takes 2bytes so we divide by two to get the limit in characters.
+  private val lineSizeLimitBytes: Int = (2.5 * FileUtils.ONE_MB).toInt
```
dmivankov, line 40: would make sense to put that into trait & parameter too? If anything it can help writing a unit test for this part

dmivankov: ah, neither is a parameter

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -114,7 +118,15 @@ class FileContentRelation(config: RelationConfig, fileExternalId: String, inferS
             byteStream
               .through(enforceSizeLimit)
               .through(fs2.text.utf8.decode)
-              .through(fs2.text.lines)
+              .through(fs2.text.linesLimited(lineSizeLimitCharacters))
+              .handleErrorWith {
+                case e: fs2.text.LineTooLongException =>
+                  throw new CdfSparkException(
+                    s"""Line too long in file with external id: "$fileExternalId" SizeLimit in characters: $lineSizeLimitCharacters""",
```
dmivankov, line 125:
```suggestion
                    s"""Line too long in file with external id: "$fileExternalId" SizeLimit in characters: $e.max, but $e.length characters accumulated""",
```
nit: for extra details

src/main/scala/cognite/spark/v1/FileContentRelation.scala
```diff
@@ -114,7 +118,15 @@ class FileContentRelation(config: RelationConfig, fileExternalId: String, inferS
             byteStream
               .through(enforceSizeLimit)
               .through(fs2.text.utf8.decode)
-              .through(fs2.text.lines)
+              .through(fs2.text.linesLimited(lineSizeLimitCharacters))
+              .handleErrorWith {
```
dmivankov, line 122: there's `.zipWithIndex` method that can expose indexes, not sure if we can attach it to error message though :thinking: 
But that's only a nice to have extra error detail

Jacob-Eliat-Eliat: It is possible, but then we cannot use linesLimited, we'd need to do a second pass with indexes and limit there. Not very hard, how much do you think this is important?

PR #997, opened 2024-12-08
Allow control of how many outstanding raw requests there are (avoid OOM)

thorkildcognite:
Before this fix, all output raw rows for a partition were converted into queued HTTP requests. This meant that not only did we use memory for the input rows as part of the spark input dataframes, but also for the serialized HTTP body versions that were waiting to be sent.. If then there is a large amount of rows in a partition, or if we are slow to write to raw (such as for 429s etc), then we end up consuming a lot of memory intermittently, leading to an OOM (confirmed by analyzing heap dumps).

After this fix, a new setting, maxOutstandingRawInsertRequests, is exposed through dataframe options, which controls how many requests we prepare awaiting execution. This allows us to test this on staging only, and then roll out for the rest of the clusters. The associated test is a bit rubbish, but it just verifies that both work (as it cannot verify that we actually throttled the requests as it is hidden within the dataframe interface -- nor how many requests are done).

Note that this is potentially made slightly worse that, as far as I can see, we don't't have a max concurrent writes to Raw at the same time (as the throttled backend isn't used on write), which means we will probably hit 429s more frequently (and backoffs will make it use more time before it gets data out of the heap).

The implementation is not optimal, as noted in the code comment, but I had issues with cats.effect not wanting to cooperate.


src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -285,10 +285,30 @@ class RawTableRelation(
...
+      config.maxOutstandingRawInsertRequests match {
+        case Some(maxOutstandingRawInsertRequests) =>
+          // We first group by batch size of a write, and then group that by the number of allowed parallel
+          // outstanding requests to avoid queueing up too many requests towards the RAW API (and this potentially
+          // leading to an OutOfMemory)
+          // Note: This is a suboptimal fix, as if one of the requests in a batch is slow, we will not
+          // start on the next batch (this limitation used to be per partition). Instead, we should
+          // have a cats.effect.std.Semaphore permit with X number of outstanding requests
+          // or cats.effect.concurrent.Backpressure.
+          (rows.grouped(batchSize): Iterator[Seq[Row]])
```
dmivankov, line 297: Removed .toSeq here

And for some reason without type annotation IntelliJ didn't like the types :thinking:

dmivankov: ah, nvm, toSeq is lazy, will undo

src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -285,10 +285,32 @@ class RawTableRelation(
...
+          // Note: This is a suboptimal fix, as if one of the requests in a batch is slow, we will not
+          // start on the next batch (this limitation used to be per partition). Instead, we should
+          // have a cats.effect.std.Semaphore permit with X number of outstanding requests
+          // or cats.effect.concurrent.Backpressure.
+          rows
+            .grouped(batchSize)
+            .toSeq
+            .grouped(maxOutstandingRawInsertRequests)
+            .foreach { batch =>
+              batch.toVector
```
dmivankov, line 302:
```suggestion
            .foreach { outstandingBatchesGroup =>
              outstandingBatchesGroup.toVector
```
nit: maybe something like that to tie it to outstanding requests naming and also make it different from the other branch batches name

dmivankov (approved):
Looks simple & safe enough, a more optimal streaming may be added later

PR #1000, opened 2024-12-11
Use Cats backpressure to throttle writes (and conversions to HTTP requests)

thorkildcognite:
As mentioned in TODO we've had a way to limit RAW writes concurrency to `maxOutstandingRawInsertRequests` but chunking into groups of size `maxOutstandingRawInsertRequests`, doing groups sequentially and each batched item in parallel. Which can under-utilize allowed concurrency.

Now we're using cats backpressure (wrapper around cats semaphore)


src/main/scala/cognite/spark/v1/RawTableRelation.scala
```diff
@@ -287,22 +288,13 @@ class RawTableRelation(
...
-            .grouped(batchSize)
-            .toSeq
-            .grouped(maxOutstandingRawInsertRequests)
-            .foreach { batch =>
-              batch.toVector
-                .parTraverse_(postRows(columnNames, _))
-                .unsafeRunSync()
+          Backpressure[IO](Backpressure.Strategy.Lossless, maxOutstandingRawInsertRequests)
+            .flatMap { backpressure =>
+              rows.grouped(batchSize).toVector.parTraverse_ { batch: Seq[Row] =>
```
dmivankov, line 293: compared to `_.grouped.toSeq.grouped.foreach` it will now create a vector of all batches, they will reference all Row items from iterator which potentially could be not a lot of mem, less than request bodies, more than constant mem

but compared to `.grouped.toVector` it is about the same apart from having semaphore

dmivankov (approved):

Jacob-Eliat-Eliat:
PR description plz

src/main/scala/cognite/spark/v1/RawTableRelation.scala
dmivankov: Hm, we should probably have a look into not eagerly converting iterator to vector

dmivankov: Probably we have to utilize fs2 streams

PR #1002, opened 2024-12-17
filecontent: report itemsRead metric

dmivankov:
Since filecontent doesn't go through `SdkV1Relation.{buildScan, getStream}` to implement reads
read metrics weren't being automatically reported.

Now it will do `itemsRead.inc()` once per line


Jacob-Eliat-Eliat:
Do you remember why we decided not to implement sdkV1relation? I vaguely remember discussing it but I don't remember the reasons for our decisions back then

dmivankov:
> Do you remember why we decided not to implement sdkV1relation? I vaguely remember discussing it but I don't remember the reasons for our decisions back then

Nope, don't remember any specifics, maybe it doesn't provide suitable helpers or assumes more operations that filecontent supports? But could be something to look into

PR #1003, opened 2024-12-17
Pin coursier/setup-action by full version, sbtn->sbt to try fix the build

dmivankov:
Looks like something has changed and https://github.com/coursier/setup-action/ installing `sbtn` no longer installed `sbt` binary.
Let's pin action version for more explicit upgrades, and install `sbt` instead of `sbtn` to hopefully fix the build.

Without it build fail with
```
Run sbt -Dsbt.log.noformat=true +{,macroSub/,structType/}dependencyLockCheck
/home/runner/work/_temp/a737f15b-e7ab-4208-ba7f-f620b1b062d2.sh: line 1: sbt: command not found
Error: Process completed with exit code 127.
```

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -68,7 +68,7 @@ jobs:
       - run: |
           git rev-list --count --first-parent HEAD > patch_version.txt
       - name: Set up JDK 8
-        uses: coursier/setup-action@v1
+        uses: coursier/setup-action@v1.3.8
         with:
           jvm: adopt:8
       - name: Build JAR file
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yaml
```diff
@@ -68,7 +68,7 @@ jobs:
       - run: |
           git rev-list --count --first-parent HEAD > patch_version.txt
       - name: Set up JDK 8
-        uses: coursier/setup-action@v1
+        uses: coursier/setup-action@v1.3.8
```
dmivankov, line 71:
```suggestion
        uses: coursier/setup-action@v1.3.7
```

Jacob-Eliat-Eliat (approved):
interesting i cannot find the breaking change on the coursier repo, their doc always said "sbtn by default". But nice that you fixed this.

PR #1004, opened 2024-12-18
add apps: sbt for sonatype too

Jacob-Eliat-Eliat:
I approved and merged your PR but we both missed that only 2/3 places had this change

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -121,6 +121,7 @@ jobs:
         uses: coursier/setup-action@v1
         with:
           jvm: adopt:8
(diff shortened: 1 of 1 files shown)


dmivankov (approved):

.github/workflows/main.yaml
```diff
@@ -121,6 +121,7 @@ jobs:
         uses: coursier/setup-action@v1
```
dmivankov, line 121:
```suggestion
        uses: coursier/setup-action@v1.3.9
```

dmivankov (approved):

PR #1008, opened 2025-02-06
Handle connection error on cursor update

mr-celo:
We used to only handle non-OK responses from `jetfire-backend`, meaning that connection errors such as timeouts or connection refused would throw and be unhandled

Examples:
* https://jetfire-admin.iap.az-eastus-1.cognite.dev/jobs/filter?uuid=6f5ca314-e0c4-4ee3-95d7-8e8064f95a18
* https://jetfire-admin.iap.az-eastus-1.cognite.dev/jobs/filter?uuid=bc4d6305-142a-4703-ad9e-721e38a9d840
* https://jetfire-admin.iap.az-eastus-1.cognite.dev/jobs/filter?uuid=0fc73d32-4466-4005-b78e-e0cfe30a8744

### DIFF
--- a/src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
+++ b/src/main/scala/cognite/spark/v1/SyncCursorCallback.scala
@@ -5,13 +5,13 @@ import com.cognite.sdk.scala.sttp.{GzipBackend, RetryingBackend}
 import io.circe.Codec
 import io.circe.generic.semiauto.deriveCodec
 import org.log4s.getLogger
-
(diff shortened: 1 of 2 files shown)


dmivankov (approved):
Interesting that retryingBackend should be retrying connection errors too, maybe it need to retry harder?

mr-celo:
I repro'd by exhausting the 5 retry attempts. So it does retry on connection errors.
But ultimately, we need a limit on how many times to retry, and catch the case where we didn't succeed even after all retries

PR #1010, opened 2025-02-14
Remove support for scala 2.12

dmivankov:
Main reason to keep was to be compatible with Databricks
Databricks Runtime 16.4 (LTS) is finally here and supports 2.13, starting 16.7 it won't support 2.12 at all, at some later point it should add scala3 support as spark 4+ will only support it
https://docs.databricks.com/aws/en/release-notes/runtime/16.4lts

Note: not yet cleaning up code that could move to newer constructs not supported by 2.12


mr-celo (approved):
🧹

larshagencognite (approved):
🦄

PR #1011, opened 2025-03-12
Static initialization: unsafeRunSync -> evalOn(blocking).unsafeRunSync

dmivankov:
This should reduce chances of hitting a deadlock where all compute
threads are hitting class initialization lock that in turn wants
to run `.unsafeRunSync()` and needs a free compute thread to do that.

We may still get class initialization lock hits inside compute pool
which isn't great, but should only happen once per class and with this
fix hopefully won't lead to a deadlock as we now use unbounded blocking pool.

There doesn't seem to be other static class initialization using `.unsafeRunSync()`,
only calls to it from datasource implementation methods on non-effects <> cats-effect
boundary which we don't control because spark doesn't provide cats-effect interface.

### DIFF
--- a/src/main/scala/cognite/spark/v1/CdpConnector.scala
+++ b/src/main/scala/cognite/spark/v1/CdpConnector.scala
@@ -54,6 +54,17 @@ object CdpConnector {
     () => (),
     IORuntimeConfig()
   )
+
+  implicit class ExtensionMethods[A](f: IO[A]) {
+    // Run on unbound blocking thread pool
+    // Useful to reduce chances of deadlocks in class initializers that are initiated
+    // from non-cats-effect code or not being wrapped into IO().
+    // Compute (bounded) pool may still be running .unsafeRunBlocking() indirectly and have compute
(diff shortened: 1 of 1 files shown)


silvavelosa (approved):
LGTM ✔️ 

My only open question is whether we have these methods as pure convenience, or we actually need the static instance to comply with a third party interface or something like that.... eg, Could `retryingSttpBackend()` return an `IO[SttpBackend[IO, _]` rather than an `SttpBackend[IO,_]`?

at least for jetfire's case, I don't see us "needing" to force the static initialization

dmivankov:
If we want to keep those classes shared per executor then we don't have a good non-cats-effect entry point. Using cacheable effect could be an option, but also is complicated and may still need shifts to .blocking (we can still look at this option later on :+1: )
There is also an option to use spark plugin that "initializes" executor, and maybe we'll do that too, with some extra preparation. But this PR should still make sense for more clean current setup and for external setups.

larshagencognite (approved):
risk review ok

PR #1012, opened 2025-03-14
Add some logging for partition settings

dmivankov:
Notes:
- doesn't cover `NumericDataPointsRdd`, `StringDataPointsRdd` and `FileContentRelation`
- for `SdkV1Rdd` doesn't cover cases where `getStreams` returns fewer
  than `config.sparkPartitions`, for such cases logger value is `config.sparkPartitions`
  rather than `getStreams` length
- we could also add metrics, but then it probably needs to be a histogram
  and so may need a little more thought for implementation

### DIFF
--- a/src/main/scala/cognite/spark/v1/DefaultSource.scala
+++ b/src/main/scala/cognite/spark/v1/DefaultSource.scala
@@ -410,7 +410,7 @@ object DefaultSource {
   def saveTracingHeaders[F[_]: Functor: Trace](): F[Seq[(String, String)]] =
     Trace[F].kernel.map(saveTracingHeaders(_))
 
-  def logParamterOrDefault[A](name: String, value: Option[A], default: => A): A = value match {
+  def logParameterOrDefault[A](name: String, value: Option[A], default: => A): A = value match {
     case None => logger.info(s"using default value for ${name}: ${default}"); default
     case Some(value) => logger.info(s"using supplied value for ${name}: $value"); value
   }
@@ -439,7 +439,7 @@ object DefaultSource {
         throw new CdfSparkIllegalArgumentException(s"`project` must be specified"))
     val batchSize = toPositiveInt(parameters, "batchSize")
     val limitPerPartition = toPositiveInt(parameters, "limitPerPartition")
-    val partitions = logParamterOrDefault(
+    val partitions = logParameterOrDefault(
       "partitions",
       toPositiveInt(parameters, "partitions"),
       Constants.DefaultPartitions)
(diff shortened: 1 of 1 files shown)


src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -390,6 +410,11 @@ object DefaultSource {
   def saveTracingHeaders[F[_]: Functor: Trace](): F[Seq[(String, String)]] =
     Trace[F].kernel.map(saveTracingHeaders(_))
 
+  def logParamterOrDefault[A](name: String, value: Option[A], default: => A): A = value match {
```
mr-celo, line 413: Typo in the name
Also, this function also returns the value.. Below, when calling it, it looks weird that we are assigning a logging function 🤔

mr-celo (dismissed):
Will be a a bit of pain to get these logs just because, well, azure logs 😅 
But good to see the final value 👍

src/main/scala/cognite/spark/v1/DefaultSource.scala
dmivankov, line 413:
```suggestion
  def logParameterOrDefault[A](name: String, value: Option[A], default: => A): A = value match {
```

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -414,8 +439,10 @@ object DefaultSource {
         throw new CdfSparkIllegalArgumentException(s"`project` must be specified"))
     val batchSize = toPositiveInt(parameters, "batchSize")
     val limitPerPartition = toPositiveInt(parameters, "limitPerPartition")
-    val partitions = toPositiveInt(parameters, "partitions")
-      .getOrElse(Constants.DefaultPartitions)
+    val partitions = logParamterOrDefault(
```
dmivankov, line 442:
```suggestion
    val partitions = logParameterOrDefault(
```

mr-celo (approved):

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -390,6 +410,11 @@ object DefaultSource {
   def saveTracingHeaders[F[_]: Functor: Trace](): F[Seq[(String, String)]] =
     Trace[F].kernel.map(saveTracingHeaders(_))
 
+  def logParameterOrDefault[A](name: String, value: Option[A], default: => A): A = value match {
+    case None => logger.info(s"using default value for ${name}: ${default}"); default
+    case Some(value) => logger.info(s"using supplied value for ${name}: $value"); value
```
bdemirel, line 415: Here you're re-using the name from the function argument. Different languages have different standards for this stuff, is this acceptable, or even suggested in Scala?

bdemirel (approved):
LGTM, I will request another risk review since I'm onboarding.

src/main/scala/cognite/spark/v1/DefaultSource.scala
dmivankov: generally hiding names can be error-prone, but for very small code scope sounds ok, plus it's only Option unwrapping - meaning doesn't change much and type system would catch most misuse

Don't know if some of scala linters/formatters have configurations for it, at least not the ones currently enabled in this repo

bdemirel: Sounds good, thanks for the explanation!

simenl (approved):
risk review ok

PR #1021, opened 2025-04-15
drop reviewer-lottery

dmivankov:
We don't use auto-assignments now, instead we poll team-review-requested PRs


psalaberria002 (approved):
:unicorn:

PR #1023, opened 2025-04-29
count dms noop writes

dmivankov:
- bump sdk to get access to .wasModified response field
- add metric for dms noop write counts


mr-celo (approved):
LGTM
Are there other relations that might expose/benefit from counting noops?

dmivankov:
> Are there other relations that might expose/benefit from counting noops?

Did a 2nd look, fdm sources should be covered, for others it's also nice info to have but we probably don't have enough info easily available on services side (on case-by-case we can look into adding noop metric)

psalaberria002 (approved):
:unicorn:

PR #1035, opened 2025-08-05
upgrade sdk

Jacob-Eliat-Eliat:


dmivankov (approved):
:unicorn:

Jacob-Eliat-Eliat:
> though
> 
> > [info] Test still running after 39 minutes, 38 seconds: suite name: StringDataPointsRelationTest, test name: StringDataPointsRelation should be an error to specify an invalid (time series) id.
> 
> could be a change on API side too

Good point, let's rerun master then

Jacob-Eliat-Eliat:
Right looks like java been upgraded in the sdk, causing some incompatibilities... wonder what's gonna happen when I bring this to jetfire.

PR #1040, opened 2025-10-27
drop src/main/scripts

dmivankov:
There's some CVEs in Pipfile.lock, and scripts didn't have any real work
for a long time, plus aren't referenced anywhere in repo, let's just
cleanup & drop

[CDF-26081]

[CDF-26081]: https://cognitedata.atlassian.net/browse/CDF-26081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dmivankov:
:unicorn:

PR #1054, opened 2025-12-05
renovate.json5: explicitly add "https://repo.maven.apache.org/maven2"…

dmivankov:
… registryUrl

It might be not being used - no updates suggested for https://repo.maven.apache.org/maven2/com/cognite/cognite-sdk-scala_2.13/ Could be that sonatype url needs to be updated too, but not sure about better url, while maven2 should already be an improvement

It seems that maven2 shows up in renovate logs https://developer.mend.io/github/cognitedata/cdp-spark-datasource/-/job/806c707e-4794-4122-bce7-88c8bd617254 but won't hurt to try anyway


Jacob-Eliat-Eliat (approved):
looks good with fixed url

PR #1057, opened 2025-12-05
.github/workflows: add explicit on:pull_request trigger

dmivankov:
Similar to sdk repo https://github.com/cognitedata/cognite-sdk-scala/blob/master/.github/workflows/scala.yml

Seeing that sometimes build doesn't trigger in PRs, maybe this can help


silvavelosa (approved):
this will prevent the pipeline from running when making a PR to other branches.... not sure if that's desirable, given scenarios like bulking changes in a single PR before merging to master. But should be safe, and those scenarios are kind of discouraged anyway

dmivankov:
:unicorn:

dmivankov:
or maybe this time was just test timeout, retrying

PR #1059, opened 2025-12-08
CDF-26311: lift retrying sttp backend higher in wrappers stack

dmivankov:
In case requests and retries take long time auth header may expire, in
current stack order it wouldn't be refreshed.

If we move retries up then each try will go through auth wrapper and
have a chance to refresh expiring auth

[CDF-26311]

[CDF-26311]: https://cognitedata.atlassian.net/browse/CDF-26311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,6 +1,10 @@
 name: CI
 
-on: ["push"]
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
 
 jobs:
   build-and-push:


src/main/scala/cognite/spark/v1/CdpConnector.scala
dmivankov: We may need to also deal with this throttle-sleeping past token expiration 🤔 https://github.com/cognitedata/cognite-sdk-scala/blob/master/src/main/scala/com/cognite/sdk/scala/sttp/BackpressureThrottleBackend.scala
it needs to run on each try of course, but current impl uses sleep, not sure how to best integrate it with retrying backend, so maybe an improvement for different task&PR

PR #1079, opened 2026-01-28
CDF-26522: unsafeRunSync -> unsafeRunBlocking to reduce deadlock chances

dmivankov:
tl;dr replace all non-testcode calls to unsafeRunSync with unsafeRunBlocking which is
```scala
def unsafeRunBlocking(): A = f.evalOn(blocking).unsafeRunSync()(ioRuntime)
```

There's sometimes a deadlock on cats-effect compute pool where threads
are doing
- relation.upsert -> need to initialize lazy client
- lazy client init runs unsafeRunBlocking() which still needs compute
  pool after blocking pool work is done

Relation.upsert itself is run via unsafeRunSync and blocks compute
thread. So the deadlock happens like so
- lazy client isn't initialized yet
- 8+ threads (>= compute pool size) are trying to call relation.upsert in parallel
- all do call unsafeRunSync and take the entire compute pool
- all hit the need to initialize lazy client
- lazy client init can't complete (even though it uses unsafeRunBlocking) until
  some compute thread becomes free

Most of datasource is just networking rather than computation, we should
be safely able to switch to using the blocking pool

There's still some potential for deadlocks
- lazy init generally may have deadlocks in certain cases
- unsafeRunBlocking still needs compute pool for the completion step,
  so some chance of it not being available and maybe leading to deadlocks

But we're moving away the bulk of relation.upsert off compute pool with
this PR so there shouldn't be anything that can't proceed on it
(unsafeRunBlocking only need it to complete and shouldn't occupy, hopefully)

Some other tweaks are possible to implement via other PRs (like trying
to force client lazy init sooner before doing parallel operations)

[CDF-26522]

[CDF-26522]: https://cognitedata.atlassian.net/browse/CDF-26522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,7 +10,7 @@ val circeVersion = "0.14.9"
 val sttpVersion = "3.5.2"
 val natchezVersion = "0.3.1"
 val Specs2Version = "4.20.3"
-val cogniteSdkVersion = "2.35.958"
+val cogniteSdkVersion = "2.35.961"
 
 val prometheusVersion = "0.16.0"
 val log4sVersion = "1.10.0"
--- a/build.scala-2.13.sbt.lock
+++ b/build.scala-2.13.sbt.lock
@@ -1,6 +1,6 @@
 {
   "lockVersion" : 1,
-  "timestamp" : "2026-01-15T11:23:16.652439Z",
+  "timestamp" : "2026-01-29T13:44:02.142098Z",
   "configurations" : [
     "compile",
     "optional",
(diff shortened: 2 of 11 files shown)


silvavelosa (approved):
IIUC, this calls only happen once per read/write operation on the executro, right?.... I mean, we're not preventing the executor from doing the bulk of its work in parallel by doing this

PR #1082, opened 2026-02-02
change order of type of dm relation guesses

Jacob-Eliat-Eliat:
We try first to parse into corePropertyRelation which works if instanceType is sent as a df option with either "node" or "edge" value. 

On the other hand, connectionRelation will return if "edgeType" is defined.

This leads to a weird situation where if you define "instanceType" to "edge" and want to use connectionRelation, you end up using corePropertyRelation instead.

Which is confusing because when you send edgeType you do know it's an edge, yet you get the parsing if you send instanceType = edge. See PR:  https://github.com/cognitedata/jetfire-backend/pull/3175 (which won't work without this one).

There is an argument that these shouldn't be separate but that's a bigger change

### DIFF
--- a/src/test/scala/cognite/spark/v1/fdm/FlexibleDataModelEdgeTest.scala
+++ b/src/test/scala/cognite/spark/v1/fdm/FlexibleDataModelEdgeTest.scala
@@ -208,7 +208,7 @@ class FlexibleDataModelEdgeTest
     val instExtIds = toExternalIds(selectedConnectionInstances)
     created.size shouldBe 2
     instExtIds.size shouldBe 1
-    instExtIds should be(Seq("edge1"))
+    instExtIds should be(Seq("edgeForEdgeTypeFilter1"))
   }
 
   it should "fetch edges from a data model" in {


src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -74,11 +74,11 @@ class DefaultSource
     val connectionRelation = extractConnectionRelation(parameters, config, sqlContext)
 
     corePropertySyncRelation
+      .orElse(connectionRelation)
```
dmivankov: is it possible to unit test the fix? 🙃

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -74,11 +74,11 @@ class DefaultSource
...
       .orElse(corePropertyRelation)
```
dmivankov: would instanceType=edge be always present (or maybe set to something else) for connectionRelation?
and is instanetType=edge without edgeTypeSpace&edgeTypeExternalId a valid corePropertyRelation?

dmivankov: looking at linked PR 1st question is that currently
- for edges relation without edge type we get instanceType=edge, which will be corePropertyRelation both before&after
- for edges relation with edge type set edgeType* are set but not instanceType, so old code won't treat it as corePropertyRelation and it'd be connectionRelation before&after

But I guess new test fails without the reordering?

PR #1087, opened 2026-02-05
Fdm core property query

Jacob-Eliat-Eliat:


src/main/scala/cognite/spark/v1/fdm/FlexibleDataModelCorePropertyRelation.scala
```diff
@@ -109,24 +112,46 @@ private[spark] class FlexibleDataModelCorePropertyRelation(
...
+      val filterRequests = compatibleInstanceTypes(intendedUsage).map { instanceType =>
+        InstanceFilterRequest(
+          instanceType = Some(instanceType),
+          filter = instanceFilter,
+          sort = None,
+          limit = config.limitPerPartition,
+          cursor = None,
+          sources = viewReference.map(r => Vector(InstanceSource(r))),
+          includeTyping = Some(true)
+        )
```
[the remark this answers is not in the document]
Jacob-Eliat-Eliat: Good spot

src/test/scala/cognite/spark/v1/fdm/utils/FDMSparkDataframeTestOperations.scala
```diff
@@ -219,4 +243,11 @@ object FDMSparkDataframeTestOperations extends SparkTest {
   def toPropVal(rows: Array[Row], prop: String): Seq[String] =
     rows.toIndexedSeq.map(row => row.getString(row.schema.fieldIndex(prop)))
 
+  private def validateDataFrameEquivalence(dataFrame1: DataFrame, dataFrame2: DataFrame): Unit = {
+    dataFrame1.schema should equal(dataFrame2.schema)
+    dataFrame1.count() should equal(dataFrame2.count())
+    dataFrame1.except(dataFrame2).count() should equal(0L)
+    dataFrame2.except(dataFrame1).count() should equal(0L)
```
Jacob-Eliat-Eliat: Actually thinking about it, count here forces evaluation of the dm. Without it, the plan should push down the filters properly but if you add count here it has to evaluate the df before the rest of the query is applied...

src/main/scala/cognite/spark/v1/RelationConfig.scala
```diff
@@ -39,7 +39,9 @@ final case class RelationConfig(
     useSharedThrottle: Boolean,
     serverSideFilterNullValuesOnNonSchemaRawQueries: Boolean,
     maxOutstandingRawInsertRequests: Option[Int],
-    sendDebugFlag: Boolean
+    sendDebugFlag: Boolean,
+    useQuery: Boolean,
+    enabledAdditionalFlag: Option[String]
```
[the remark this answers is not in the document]
Jacob-Eliat-Eliat: Not sure how well it'll parse a Seq in there, should be fine I guess?

src/main/scala/cognite/spark/v1/fdm/FlexibleDataModelCorePropertyRelation.scala
```diff
@@ -109,24 +111,59 @@ private[spark] class FlexibleDataModelCorePropertyRelation(
...
+            config.limitPerPartition)
+        val selectExpression = SelectExpression(
+          sources = sourceReference(instanceType, viewReference, selectedInstanceProps),
+        )
+        client.instances
+          .queryStream(
+            inputTableExpression = tableExpression,
+            inputSelectExpression = selectExpression,
+            limit = config.limitPerPartition,
+            additionalFlags = config.enabledAdditionalFlag.map(_ -> true).toMap,
```
dmivankov: maybe allow flags to be true/false?

src/main/scala/cognite/spark/v1/fdm/FlexibleDataModelCorePropertySyncRelation.scala
```diff
@@ -88,29 +89,12 @@ private[spark] class FlexibleDataModelCorePropertySyncRelation(
...
-      instanceType match {
-        case InstanceType.Node => result
-        case InstanceType.Edge => result ++ Seq("startNode", "endNode", "type")
-      }
-    }
 
     val cursors = if (cursor.nonEmpty) Some(Map("sync" -> cursor)) else None
     def select(instanceType: InstanceType) =
-      Map("sync" -> SelectExpression(sources = sourceReference(instanceType)))
+      Map(
```
dmivankov: this too looks like something scalafmt would do, is the check working? Or did it get into "check passes, but re-format changes code" state?

Jacob-Eliat-Eliat: Hmm, could both are valid according to fmt rules

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -532,7 +540,9 @@ object DefaultSource {
       serverSideFilterNullValuesOnNonSchemaRawQueries =
         toBoolean(parameters, "filterNullFieldsOnNonSchemaRawQueries", defaultValue = false),
       maxOutstandingRawInsertRequests = toPositiveInt(parameters, "maxOutstandingRawInsertRequests"),
-      sendDebugFlag = toBoolean(parameters, "sendDebugFlag", defaultValue = false)
+      sendDebugFlag = toBoolean(parameters, "sendDebugFlag", defaultValue = false),
+      useQuery = toBoolean(parameters, "useQuery", defaultValue = true),
```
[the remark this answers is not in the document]
Jacob-Eliat-Eliat: Doesn't matter much I guess, we'll always set it in jetfire, but false seems like a better default for now

Jacob-Eliat-Eliat:
Filter pushdown is not well tested before or after. Imo should be a task to add filter pushdown tests that actually test what is pushed down. Since Spark always double-filters everything, just testing that the filters are applied on the final result set is not sufficient. I think scope is a bit bigger than this PR though

src/main/scala/cognite/spark/v1/fdm/FlexibleDataModelQueryUtils.scala
```diff
@@ -0,0 +1,48 @@
...
+  SourceSelector,
+  TableExpression
+}
+import com.cognite.sdk.scala.v1.fdm.views.ViewReference
+
+object FlexibleDataModelQueryUtils {
+  def generateTableExpression(
+      instanceType: InstanceType,
+      filters: Option[FilterDefinition],
+      limit: Option[Int] = None): TableExpression =
```
dmivankov: default is changing, it was 1000 before on sdk side, now it'd be set to force None (service default 10k iirc), should probably be under flag or keeping old default if this is called without limie=

src/main/scala/cognite/spark/v1/fdm/FlexibleDataModelCorePropertyRelation.scala
```diff
@@ -109,24 +111,58 @@ private[spark] class FlexibleDataModelCorePropertyRelation(
...
-        limit = config.limitPerPartition,
-        cursor = None,
-        sources = viewReference.map(r => Vector(InstanceSource(r))),
-        includeTyping = Some(true),
-        debug = optionalDebug(config.sendDebugFlag)
-      )
-    }
+    def queryFilterWithHasData(
+        instanceFilter: Option[FilterDefinition],
+        viewReference: Option[ViewReference]): Option[FilterDefinition] =
```
dmivankov: nit: probably can do something like
```scala
instanceFilter.toSeq ++ viewReference.map(ref => HasData(Seq(ref))).toSeq match {
  case s if s.length > 1 => Some(And(s))
  case s => s.headOption
}
```
and maybe introducing a helper andFilters: Seq[FilterDefinition] => Option[FilterDefinition] so be even cleaner

src/main/scala/cognite/spark/v1/fdm/FlexibleDataModelCorePropertyRelation.scala
```diff
@@ -109,24 +111,58 @@ private[spark] class FlexibleDataModelCorePropertyRelation(
...
+            inputSelectExpression = selectExpression,
+            limit = config.limitPerPartition,
+            additionalFlags = config.additionalFlags.getOrElse(Map.empty),
+            batchSize = config.batchSize
+          )
+          .map(toProjectedInstance(_, None, selectedInstanceProps))
+      }
+    } else {
+      val filterRequests = compatibleInstanceTypes(intendedUsage).map { instanceType =>
+        InstanceFilterRequest(
```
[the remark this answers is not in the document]
dmivankov: interesting, can we test it in datasource tests (query without hasdata/with hasdata vs filter)? Not necessarily within this PR scope

src/main/scala/cognite/spark/v1/DefaultSource.scala
```diff
@@ -348,6 +356,14 @@ object DefaultSource {
       case None => defaultValue
     }
 
+  private def getAdditionalFlags(
+      parameters: Map[String, String]
+  ): Map[String, Boolean] =
+    parameters.collect {
+      case (key, _) if key.startsWith("additionalFlag_") =>
+        key.stripPrefix("additionalFlag_") -> toBoolean(parameters, key)
```
dmivankov: nit: similar to tracing flags would be good to expose one of or both
- prefix value as constant
- reverse encoder: Map String->Bool -> Map String -> String or similar

src/main/scala/cognite/spark/v1/fdm/FlexibleDataModelRelationUtils.scala
```diff
@@ -869,4 +870,13 @@ object FlexibleDataModelRelationUtils {
 
   private def rowToString(row: Row): String =
     Try(row.json).getOrElse(row.mkString(", "))
+
+  private[spark] def toAndFilter(filters: Vector[FilterDefinition]): Option[FilterDefinition] =
+    if (filters.isEmpty) {
```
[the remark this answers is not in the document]
Jacob-Eliat-Eliat: Think existing is fine, difference is not large imo ^^

PR #1088, opened 2026-02-05
split relation utils

Jacob-Eliat-Eliat:
Wondering if something like this (minimum changes here) is viable to help clean things up a bit. It's a bit difficult to understand what exactly is included in "FlexibleDataModelsRelationUtils" and this split along Convertors, RowDataExtractors and Validators should help a bit... But it might be an absolute pain to review so not sure if worth it.

This is not a major improvement, just a (hopefully) relatively logical split of a large class into 4 more targeted ones, should be no functional changes. My biggest concern is how annoying it's going to be to review for a reviewer :/


Jacob-Eliat-Eliat:
Current state somewhat sucks but difficult to improve, won't work on this for now.

PR #1091, opened 2026-02-17
enable query for connection relations

Jacob-Eliat-Eliat:
existing test cover this (as in, they both send and don't send the flag, so they go through both flow and will detect if one of them is broken)

Ideally we probably want to fix the large amount of code duplication between this and coreProperty. This is not an issue stemming from this PR (it was already pretty much same between the two) but I'd understand if required to fix it. (oh, just noticed the bot sees it to!)

The difference between these two in the queryStream method seem to mostly stem from differences in filters (automatically added edgeType filter) and the fact that connectionRelation is always edge only while CorePRoperty can be one, the other or both. All of this can probably be generified, but I'd argue this would best be done separately and handle the rest of the similarities at the same time


dmivankov (approved):
🦄

